PR #1, opened 2016-09-09
rgw sls files for deployment

theanalyst:


jan--f:
The keyring seems rather permissive. According to tthe ceph docs `mon 'allow rwx' osd 'allow rwx'` should suffice for the daemon itself. Maybe there is some other reason for the permissions?

theanalyst:
> The keyring seems rather permissive. According to tthe ceph docs mon 'allow rwx' osd 'allow rwx' should suffice for the daemon itself. Maybe there is some other reason for the permissions?

Yeah my mistake, will get this

theanalyst:
Fixed the keyring, ready for merge, only command which we may need to temporarily replace maybe pkg.install -> zypper

PR #2, opened 2016-09-14
stage 2: bug fixes

rjfd:
There were 2 bugs when running stage 2.

The first was the lack of curly brackets surrounding the target, and the second was due to the rename of the ceph.event.begin to  ceph.events.begin_prep


rjfd:
This PR is no longer needed since the problem was corrected in commit f4069d4

PR #3, opened 2016-09-15
minion_id: Use fqnd instead of id

rjfd:
When calling salt-runner on master, that is also a minion,
grains['id'] will append the suffix "_master" to the minion_id.

This caused me problems, since the admin node is also the salt-master, and instead of reading the admin.ceph.sls file, was trying to read the admin.ceph_master.sls file that didn't exist.

### DIFF
--- a/srv/pillar/ceph/init.sls
+++ b/srv/pillar/ceph/init.sls
@@ -1,7 +1,7 @@
 
 
 
-{% include 'ceph/cluster/' + grains['id'] + '.sls' ignore missing %}
+{% include 'ceph/cluster/' + grains['fqdn'] + '.sls' ignore missing %}
 
 {% include 'ceph/master_minion.sls' ignore missing %}
 


swiftgist:
Ricardo,
  Would you check your /etc/salt/minion_id on your master?  The only time I see the _master appended is in the salt option '__pub_id'.  Does your "hostname -f" not match the contents of the minion_id?  What does 'salt-key -L' show?

  There's a runner and reactor files that would likely need to change as well; however, I would think not having the fqdn and minion id match would be the more confusing route.  I expect that the salt-key command above will show your master as 'admin.ceph_master'.  Salt, itself, as well as any commands you try to run will use the minion id.

rjfd:
@swiftgist the 'salt-key -L' shows "admin.ceph", and the command `salt 'admin.ceph' grains.item id` also returns admin.ceph, although when the runner executes it gets admin.ceph_master

rjfd:
@swiftgist I think this is solved in master, right? Can I close this?

PR #5, opened 2016-09-15
push runner: correct the comment string in customization yaml files

jan--f:
PR cause I'd like someone to look at the indentation of the long line split here https://github.com/SUSE/DeepSea/compare/master...jan--f:master#diff-bc35b5fa836936a3136544887240f188R141

I also took the chance and cleand up some trailing whitespaces.

### DIFF
--- a/srv/modules/runners/push.py
+++ b/srv/modules/runners/push.py
@@ -142,7 +142,8 @@ class PillarData(object):
                             custom_split[0],
                             "stack/default",
                             custom_split[1])
-                    yml.write("# Customizations for {}\n".format(custom_for))
+                    yml.write("# {}\n".format(custom))
+                    yml.write("# Overwrites configuration in {}\n".format(custom_for))
 
 
     def _merge(self, pathname, common):
--- a/srv/modules/runners/push.py
+++ b/srv/modules/runners/push.py
@@ -9,33 +9,33 @@ import imp
 import re
 
 """
-This runner is the complement to the populate.proposals. 
+This runner is the complement to the populate.proposals.
 
 The policy.cfg only includes matching files.  The admin may use globbing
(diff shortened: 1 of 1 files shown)


srv/modules/runners/push.py
```diff
@@ -137,9 +137,14 @@ def _custom(self, custom):
...
                 with open(custom, "w") as yml:
-                    yml.write("# Customizations for {}\n".format(custom))
-    
-    
+                    custom_split = custom.split("stack")
+                    custom_for = "{}{}{}".format(
+                            custom_split[0],
+                            "stack/default",
+                            custom_split[1])
+                    yml.write("# Customizations for {}\n".format(custom_for))
```
theanalyst: We could probably do it in one shot as something like 
```py
yml.write("customizations for {0}{1}{2}\n".format(split[0]...)
```
Though that doesn't describe it any better :)

jan--f: yeah, I had that initially but then the line gets even longer.

jan--f: Where is pathname from?
In any case imho right now it sounds like this file overwrites the content in the default tree. Maybe we should add two comments? Something like:
`# This file will overwrite configuration made in /srv/pillar/ceph/stack/default/global.yml when put in /srv/pillar/ceph/stack/global.yml`

swiftgist: line 96.  It's the keys of the relative filenames that made it this far.

maybe
# /srv/pillar/ceph/stack/global.yml
# Overwrites configuration in /srv/pillar/ceph/stack/default/global.yml

jan--f: sure lgtm...will push a fix

PR #7, opened 2016-09-23
validate.py: handle missing keys gracefully

tserong:
As mentioned in bsc#1000620, if the pillar data is largely
nonexistent (due to a misconfiguration in that bug),
validate.py will bail out with KeyError exceptions.  Because
it throws an exception rather than returning False in this
case, the "validate failed" state in ceph/stage/deploy.sls
doesn't trigger, and salt attempts to run all the other
deploy states which fail rather spectacularly at that point.

This commit makes validate.py cope properly with missing
keys, although some of the error text when the keys aren't
there could still be improved (empty strings get interpolated
in a couple of places).

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -104,9 +104,9 @@ class Validate(object):
         """
         Validate fsid from first entry
         """
-        fsid = self.data[self.data.keys()[0]]['fsid']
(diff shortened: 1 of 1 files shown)


tserong:
@swiftgist, mind reviewing?

PR #8, opened 2016-09-23
Add support for JSON output to validate.py

sebastian-philipp:
`salt-run --out=json validate.pillars` will now print
the report as JSON object.

Signed-off-by: Sebastian Wagner <sebastian.wagner@it-novum.com>

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -5,9 +5,11 @@ import salt.utils.error
 import logging
 import ipaddress
 import pprint
+import json
 import yaml
 import os
 import re
+import sys
 from subprocess import call, Popen, PIPE
 from os.path import dirname
 
@@ -41,6 +43,33 @@ class bcolors:
     ENDC = '\033[0m'
     BOLD = '\033[1m'
     UNDERLINE = '\033[4m'
+    
+class PrettyPrinter:
+    
+    def add(self, name, passed, errors):
+        # Need to make colors optional, but looks better currently
+        for attr in passed.keys():
+            print "{:25}: {}{}{}{}".format(attr, bcolors.BOLD, bcolors.OKGREEN, passed[attr], bcolors.ENDC)
+        for attr in errors.keys():
+            print "{:25}: {}{}{}{}".format(attr, bcolors.BOLD, bcolors.FAIL, errors[attr], bcolors.ENDC)
+            
+    def print_result(self):
+        pass
+            
+class JsonPrinter:
+    
+    def __init__(self):
+        self.result = {}
+    
+    def add(self, name, passed, errors):
+        self.result[name] = {'passed': passed, 'errors': errors}
+            
+    def print_result(self):
+        json.dump(self.result, sys.stdout)
+        
+def get_printer(__pub_output=None, **kwargs):
+    return JsonPrinter() if __pub_output in ['json', 'quiet'] else PrettyPrinter()
+    
 
 
 class SaltOptions(object):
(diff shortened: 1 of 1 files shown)


srv/modules/runners/validate.py
```diff
@@ -593,21 +616,25 @@ def pillars(**kwargs):
...
+def pillar(name = None, printer=None, **kwargs):
     """
     Check that the pillar for each cluster meets the requirements to install
     a Ceph cluster.
     """
-
-    if name == None:
-        if 'name' in kwargs:
-            name = kwargs['name']
-
```
swiftgist: I still need that check for name in kwargs.  The salt state files look a bit better using kwargs.

srv/modules/runners/validate.py
```diff
@@ -642,21 +669,27 @@ def pillar(name = None, **kwargs):
...
+    printer = printer = get_printer(**kwargs)
+
     v = Validate("setup", pillar_data, [])
     v.master_minion()
     v.ceph_version()
-    v.report()
-
+    v.report(printer)
+    
+    printer.print_result()
```
swiftgist: What do you think of moving the setting of the printer to the init of the Validate class?  And calling printer.print_result() from v.report()?

sebastian-philipp: sure!

srv/modules/runners/validate.py
swiftgist: You wouldn't have both... More an exclusive OR usage.  At the command line, most salt commands seem to allow/encourage plain args such as

salt-run validate.pillar ceph

but when I call that in an sls file, the kwargs syntax is a bit more informative and the only method I could get working for runners under the reactor.  

add minion:
  runner.filequeue.enqueue:
    - kwargs:
        queue: prep 
        item: {{ data['id'] }}

with the two different usages, that seemed a reasonable option.

sebastian-philipp: Jup, but this is also not possible, as function parameter names must be unique:
```
In [17]: my_kwargs = {'name': 'name2'}

In [18]: print_args('name1')
name1 {}

In [20]: print_args('name1', **my_kwargs)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-20-2d4c8d8ad7ba> in <module>()
----> 1 print_args('name1', **my_kwargs)

TypeError: print_args() got multiple values for keyword argument 'name'

In [21]: print_args(**my_kwargs)
name2 {}
```

swiftgist: Are you worried that somebody will call validate with

salt-run validate.pillar ceph name=ceph

I get what you expect

TypeError: pillar() got multiple values for keyword argument 'name'

which is what I would expect.  I'm not sure what expectations of running the command I'm hoping to avoid.

sebastian-philipp: I really didn't thought about the way `validate.pillar` is called by anyone. 

 I was just thinking about the Python function on it's own. If you're declaring a Python function (e.g. `print_args`) with some regular arguments (e.g. `name`) plus a star-star-argument for the rest of the function arguments (e.g. `kwargs`), a regular argument value will never end up in the star-star-argument. 

This is completely unrelated to the way how salt is calling your function, I just saw an `if`-expression that was always evaluated to `False`. Thus, I removed it.

swiftgist: I didn't understand what you were explaining.  I just tried validate without the three lines and it worked.  I have seen so many python examples of args and kwargs treated separately that I did not expect a named parameter to look like a kwargs and yet not be one.

PR #9, opened 2016-09-23
Change permissions to match salt for shared directories.

oms4suse:
In addition cleaned up globing of directories.

Signed-off-by: Owen Synge <osynge@suse.com>

### DIFF
--- a/deepsea.spec
+++ b/deepsea.spec
@@ -28,7 +28,6 @@ Group:          System/Libraries
 Url:            http://bugs.opensuse.org
 Source0:        deepsea-%{version}.tar.gz
 
-#BuildRequires:  
 Requires:       salt
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildArch:      noarch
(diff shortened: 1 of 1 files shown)


deepsea.spec
```diff
@@ -273,66 +272,90 @@ sed -i '/^master_minion:/s!_REPLACE_ME_!'`hostname -f`'!' /srv/pillar/ceph/maste
 %postun 
 
 %files
-%defattr(-,salt,salt,-)
-/srv/modules/pillar/*
-%dir /etc/salt
-%dir /etc/salt/master.d
+%defattr(-,root,root,-)
+/srv/modules/pillar/stack.py
+%dir %attr(0750, root, salt) %_sysconfdir/salt
```
l-mb: Aren't these directories owned by the salt package already?

deepsea.spec
```diff
@@ -273,66 +272,90 @@ sed -i '/^master_minion:/s!_REPLACE_ME_!'`hostname -f`'!' /srv/pillar/ceph/maste
...
+%dir %attr(0755, root, salt) %{_sysconfdir}/salt/master.d/
 %dir /%{_runners}
 %dir /%{_pillar}
 %dir /%{_pillar}/ceph
 %dir /%{_pillar}/ceph/stack
 %dir /srv/modules
 %dir /srv/modules/pillar
-%dir /srv/salt
 %dir /srv/salt/_modules
 %dir /%{_saltceph}
```
l-mb: Shouldn't we specify the attr here too?

deepsea.spec
```diff
@@ -273,66 +272,90 @@ sed -i '/^master_minion:/s!_REPLACE_ME_!'`hostname -f`'!' /srv/pillar/ceph/maste
...
-%config /%{_saltceph}/reactor/*
-%config /%{_saltceph}/refresh/*
-%config /%{_saltceph}/repo/*
-%config /%{_saltceph}/rgw/*
-%config /%{_saltceph}/stage/*
-%config /%{_saltceph}/sync/*
-%config /%{_saltceph}/time/*
-%config /%{_saltceph}/updates/*
+%config /srv/salt/_modules/*.py
+%config /%{_saltceph}/admin/*.sls
```
l-mb: If we included /%saltceph/ w/o the %dir directive, wouldn't that automatically pull in all subdirectories too?

deepsea.spec
```diff
@@ -273,66 +272,90 @@ sed -i '/^master_minion:/s!_REPLACE_ME_!'`hostname -f`'!' /srv/pillar/ceph/maste
...
+%config /%{_saltceph}/repo/*.sls
+%config /%{_saltceph}/rgw/*.sls
+%config /%{_saltceph}/stage/*.sls
+%config /%{_saltceph}/stage/configure/*.sls
+%config /%{_saltceph}/stage/discovery/*.sls
+%config /%{_saltceph}/stage/prep/*.sls
+%config /%{_saltceph}/sync/*.sls
+%config /%{_saltceph}/time/*.sls
+%config /%{_saltceph}/updates/*.sls
 %doc
```
l-mb: I think you may want %docdir here?

deepsea.spec
oms4suse: %defattr(-,root,root,-)

matches the salt packaging.

deepsea.spec
oms4suse: This is not a good plan as then the only rpm that can ever install content in %{_saltceph}.

It is also recommended as good practice to use the %dir in many rpm packaging guides.

deepsea.spec
oms4suse: I tend to agree here, but this is an enhancement that could lead to delay in getting the patch accepted.

deepsea.spec
oms4suse: I believe they are, and hope we can resolve this better with a future bug report and patch.

PR #12, opened 2016-09-26
spec: don't take ownership of salt-master dirs (bsc#1000670)

tserong:
Previously, deepsea took ownership of a few directories that
are also owned by salt-master.  Given that deepsea actually
requires salt-master in order to operate, I've udated the
Requires: and BuildRequires: appropriately, and removed
the relevant %dir lines to avoid conflicts in future.


smithfarm:
While this is possible to achieve by declaring the directories with the exact same ownership and permission bits, it seems fragile and I don't understand what we gain by doing this.

So, :+1: from my side.

PR #16, opened 2016-09-30
stage.3: fix bug in sync time formula

rjfd:


swiftgist:
@rjfd what error did you get?  Both formats seem to work for me

rjfd:
@swiftgist I can't remember exactly what was the error (yes, I should have copied it to the PR description, my bad), but it was something about a bash error when executing the whole command under double quotes.
If both ways work for you, and for me only one works, then we should go for the conservative solution :smile:

PR #17, opened 2016-10-06
cephfs benchmark

jan--f:
This PR adds basic benchmarking functionality.
So far only fio on CephFS is supported. Deploy an MDS, give at least one node the role `mds-client` and run `salt-run state.orch ceph.stage.benchmark_cephfs`.
This will run a simple fio benchmark and leave results on the master in `/run/cephfs_bench_logs`.

Looking forward to reviews, comments and bugs. I tried to modify the spec file accordingly and probably made mistakes.


tserong:
I had a quick skim of the spec file - you've got %dir entries for the new directories, but that's just the directories. You'll also need to add lines to have the package include the files in those directories. This can be wildcarded (see the various lines below the %dir lines for examples). One thing to try if you didn't already, is check out DeepSea from IBS, then make a tarball from your branch (not sure how @swiftgist is doing this ATM), drop it into the build service working directory, apply your spec file changes then run `osc build` and see what breaks at the end ;-)

(I still think we should be able to collapse/simplify a bunch of stuff in the specfile, but there's no need to do that as part of this PR. I've been meaning to spend some time on that myself but haven't gotten to it yet :-/)

tserong:
I'd probably suggest dumping the logs to either the current working directory or /var/log rather than /run (/run tends to be pidfiles and lockfiles and whatnot)

deepsea.spec
```diff
@@ -112,6 +130,20 @@ install -m 644 %{_saltceph}/admin/key/init.sls %{buildroot}/%{_saltceph}/admin/k
...
+install -m 644 %{_saltceph}/benchmark/cleanup.sls %{buildroot}/%{_saltceph}/benchmark
+install -m 644 %{_saltceph}/benchmark/cleanup_working_subdir.sls %{buildroot}/%{_saltceph}/benchmark
+install -m 644 %{_saltceph}/benchmark/default.sls %{buildroot}/%{_saltceph}/benchmark
+install -m 644 %{_saltceph}/benchmark/fio.sls %{buildroot}/%{_saltceph}/benchmark
+install -m 644 %{_saltceph}/benchmark/fio_service.sls %{buildroot}/%{_saltceph}/benchmark
+install -m 644 %{_saltceph}/benchmark/init.sls %{buildroot}/%{_saltceph}/benchmark
+install -m 644 %{_saltceph}/benchmark/mount.sls %{buildroot}/%{_saltceph}/benchmark
+install -m 644 %{_saltceph}/benchmark/prepare_master.sls %{buildroot}/%{_saltceph}/benchmark
+install -m 644 %{_saltceph}/benchmark/working_subdir.sls %{buildroot}/%{_saltceph}/benchmark
+
```
jan--f: There are even more errors in the following lines. I' want to do some file re-organisation anyway...will fix this then.

jan--f:
good point...will move them.

jan--f:
I will close this PR for now. Sorry if that causes unnecessary confusion.
After a longer chat with Eric, I'm not quite sure if we should merge this branch into master yet. The basic question is, if the benchmarking should be in SES4 or not. We identified quite a few open questions that would need exploration and DeepSea as a product for SES4 should take shape soon.
Until a decision is made the code will live in a branch in the SUSE repo under jan-benchmark-wip.

PR #20, opened 2016-10-13
Wip nettest

oms4suse:
Add ping and iperf3 network tests.

### DIFF
--- a/srv/modules/runners/nettest.py
+++ b/srv/modules/runners/nettest.py
@@ -1,0 +1,32 @@
+#!/usr/bin/python
+
+import salt.client
+import select
+
+def minion_link_ipv4(host = False, **kwargs):
+    """
+    Return list of (minion, ipv4) touples
+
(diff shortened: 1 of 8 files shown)


swiftgist:
Should I be adding iperf/ping to any of the stages?  I haven't thought this out, but should the orchestrate files be the beginning of a diagnostics directory?  I'm thinking 

salt-run state.orch ceph.diagnose.iperf

is better than 

salt-run state.orch ceph.stage.iperf

Also, we would be able to have

salt-run state.orch ceph.diagnose

that would run all of them (i.e. iperf, ping and whatever else falls into that category).  

I do want to run a test with this.

oms4suse:
I will wait until you have done your tests, 

It would be good to get these tests delivered in the next Milestone.

When you decide if you do or don’t want a diagnose directory I will do what ever you want. I have no strong opinion here.

I also have no proposals on the stages should have tests in them. Since they take some time to run and that time grows with the size of the cluster, and number of networks I would keep them separated.

PR #21, opened 2016-10-18
remove bootstrap runner; no longer used

jan--f:


jan--f:
A quick review of the spec file would be good.

PR #22, opened 2016-10-19
implementation for a sequential restarting mechanism in the right order

jschmid1:
what works:
* mon
* mds
* rgw
* igw
* osd


smithfarm:
The descriptions on this PR don't tell me much about what it's trying to do.

Why would someone want to start (or restart) OSDs in a particular order? (And anyway they are started by ceph-disk at boot, right?)

jschmid1:
It is part of:
FATE #321090 (I): Automated roll-out of maintenance updates

> Why would someone want to start (or restart) OSDs in a particular order? (And anyway they are started by ceph-disk at boot, right?)

trying to solve the problem of automated restarting after a update. We do not want all OSDs to be restarted at the same time.

swiftgist:
@smithfarm It's not so much about starting the OSDs in a specific order as it is restarting the OSDs sequentially.  In that vein, running through all the OSDs on a specific storage node seems as good as any.

swiftgist:
@jschmid1 I think we will need a salt-runner after all with respect to OS updates that require a reboot.  Down in stage/prep/minion/default.sls (i.e. DeepSea 0.6+), I am targeting all minions.  This is fine for the initial installation (and likely preferred), but needs to be sequential when a cluster already exists.  

The experiment is "can a salt runner return '*' and be correctly assigned to a Jinja variable to be used as a target in the orchestrate state?"  If so, then returning '*' when no roles are assigned would give the same behavior now.  If roles are assigned, then returning the minion names in the "preferred" order should do the trick.

I am looking at wrapping both the ceph.updates and ceph.updates.restart in this loop.

smithfarm:
OK, but when a node reboots all the OSDs on the node do start at the same time (I think).

jschmid1:
But we do not aiming for a reboot here. It's about restarting single services. 

If there is a update that requires a reboot, such as a kernel update, things are different and we can't intervene.

smithfarm:
Thanks for the clarifications. On principle, I suppose it makes sense for DeepSea to restart OSDs in a staggered fashion, provided this doesn't cause the restart operation to take a painfully long time (imagine a node with 30 OSDs).

Upgrades are typically done in a rolling fashion anyway, right? You upgrade a node, and then restart the OSDs, and wait for HEALTH_OK. Then you go on to the next node. So I'm not sure what we are gaining by "sequentializing" the OSD restart, other than adding wall-clock time.

We don't want to make upgrades more painful than they already are, and we do have a "ceph-osd.target" that automagically operates on all OSDs at once. So you could just do "systemctl stop ceph-osd.target ; systemctl start ceph-osd.target" and it might complete faster (maybe even much faster, depending on how you're doing the sequential restart).

jschmid1:
The benefits that you gain here are:

* No downtime - you'll only have 1 OSD down at a time.
* In case of a failure you won't end up with 30 broken OSDs

The downsides:

* Takes longer to complete ( do we actually care? It's automated and won't be monitored )
* I saw data reshuffling while during a restart of one OSD. ( might be connected to latency issue in my cloud setup )
* Might cause a problem when you have 1/3 of OSDs running on a different version - you will have the same problem when using ceph-osd.target but for a shorter duration though.


I think the decision that has to be made here depends on whether productionlevel clusters like bulk or sequential restarts better.

swiftgist:
I agree we do not want to make an upgrade more painful that it already is.  Also, rebooting a storage node does take all the OSDs offline.

Still, the logic in Salt for doing some of these things is sometimes easy and sometimes not.  We're also working ourselves towards a solution as opposed to having well lit path that we should follow.  So, the slowest solution likely needs to exist when we do want it.  Maybe the ultimate solution will be to do the first two OSDs on any given storage node sequentially and then the remainder in parallel.  (The idea that if it worked twice, then the rest of should work.)  That may be a compromise of keeping nearly the same upgrade time while also having only one broken OSD instead of 30 when bad things happen.

@jschmid1 Considering the restarting of OSDs shouldn't generally cause data shuffling, some servers have ridiculous reboot times (think 15+ minutes... yes, many things should've been disabled, but we still get the support call).  If we are controlling the upgrade process, then OSDs should likely be marked as noout and then unset at the end of each reboot/restart cycle.

smithfarm:
> Takes longer to complete ( do we actually care? It's automated and won't be monitored )

I wouldn't be so sure about the "won't be monitored" part. Any admin worth her salt will be watching the upgrade and waiting anxiously for it to finish. If upgrades start taking hours to complete, that might cause admins to worry more (I know I would).

By default the MONs will mark unresponsive OSDs down after 5 minutes. That probably explains the data reshuffling you were seeing. In an upgrade scenario, I agree with @swiftgist in principle, but shouldn't it be "nodown" instead of "noout"? Or, perhaps even:

```
ceph osd set noup      # prevent OSDs from getting marked up
ceph osd set nodown    # prevent OSDs from getting marked down
```

before the upgrade, then wait for upgrade to finish, restart the OSDs, and then:

```
ceph osd unset noup
ceph osd unset nodown
```

swiftgist:
@smithfarm I was thinking more about preventing data shuffling with noout.  I thought flapping is more related to latency issues or network connectivity and when an admin would use noup/nodown.  If a server is rebooting, the OSDs are down.  I just want Ceph to be patient and not move data.

smithfarm:
Ah, right. When the OSD is stopped, the MONs mark the OSD down immediately. Then, they wait for it to come back. If it doesn't come back, 300 seconds later it will be marked out and data movement commences. So "noout" makes sense. Thanks for bearing with me.

jschmid1:
> Maybe the ultimate solution will be to do the first two OSDs on any given storage node sequentially and then the remainder in parallel. (The idea that if it worked twice, then the rest of should work.) That may be a compromise of keeping nearly the same upgrade time while also having only one broken OSD instead of 30 when bad things happen.

True, we might want to have the best of two worlds here.

> OSDs should likely be marked as noout and then unset at the end of each reboot/restart cycle.

Will add that to the statefiles.

jschmid1:
> I wouldn't be so sure about the "won't be monitored" part. Any admin worth her salt will be watching the upgrade and waiting anxiously for it to finish. If upgrades start taking hours to complete, that might cause admins to worry more (I know I would).

We might disagree on that. Discussions with Eric led us to the conclusion that an automated deployment/maintenance tool should also be handled like one. Which means that it will install updates and perform restarts on a regular basis. We should not be fearful about applying updates and restarting services.

smithfarm:
That's a misunderstanding, sorry. No fear of restarting services, but if an upgrade that used to take 5 minutes starts to take 1 hour, that is reason for concern, no?

jschmid1:
Mh, I'm not certain on that one. If you don't even notice that your cluster is upgrading; Do you actually care?
But I agree in a general sense. Fast updates is something we want. We just need to measure out the pros & cons of the different approaches.

smithfarm:
> If you don't even notice that your cluster is upgrading; Do you actually care?

As an administrator, I want to have control over when upgrades take place, and I want to monitor progress. Especially for such a critical component as a storage system.

jschmid1:
So it seems we all have different opinions on how to proceed with updates/maintenance. 

imo scheduling updates sounds right to me. Eric has a different opinion here which sounds pretty convincing. @swiftgist could you please elaborate on that.

swiftgist:
My end goal is to provide the working automation for when the administrator trusts the process.  I understand your point about wanting to control the upgrade.  I distrust new software until proven otherwise.

To accommodate the need for manually controlling the process, the orchestration file will be calling the steps to accomplish the restart, upgrade, etc..  The admin can circumvent it and call the state files themselves on a single minion if desired.  So, if you have no desire for the entire cluster to upgrade and you want to oversee and control each step, you can.

For a different perspective of wanting to use the orchestration but view progress, Salt has events.  However, I think it's a bit weak at the moment.  All of our steps fire an event for this purpose, but running 'salt-run state.event pretty=True' is a bit like watching an osd log with debug set to 20.  It's too much noise and not enough signal.  One of the things we could use (or Salt could use) is a smart event viewer.  I experimented with making a primitive one but was not happy with it.  I am imagining a top-like command that does not display normal Salt events by default.  If it could aggregate these events and allow quick sorting by time and group, then I think "watching the progress" of anything in Salt would be solved.

jschmid1:
are we ready to merge here?

PR #23, opened 2016-10-21
Add nettests to spec file

oms4suse:
The nettests need to be added to the rpm.

Signed-off-by: Owen Synge <osynge@suse.com>


oms4suse:
Rebased to master

PR #27, opened 2016-10-26
Bf network testing tools no ipv4

oms4suse:
Check for node in return result.

In some situations salt does not return the ipv4 addresses for a node.
This fix logs this error and and issues a warning that this node will be skipped
in testing the network connectivity.

### DIFF
--- a/srv/modules/runners/nettest.py
+++ b/srv/modules/runners/nettest.py
@@ -2,6 +2,9 @@
 
 import salt.client
 import select
+import logging
+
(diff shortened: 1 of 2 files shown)


oms4suse:
This branch does not work yet due to not understanding the meaning of a defaulted variable expect an update.

This casues an issue with minions with long hostnames.

oms4suse:
Fix for long hostnames applied.

tserong:
Included in #44

PR #29, opened 2016-10-26
Wip nettests improved cleanup

oms4suse:
The cleanup of iperf3 tests was not working correctly.

If you ran the iperf3 tests in parallel systemd could refuse to start iperf3 as a service.


tserong:
Included in #44

PR #33, opened 2016-10-27
Makefile: create install target

tserong:
This moves all the `install` invocations from the specfile to
the Makefile, and collapses as many as possible with wildcards.
Where wildcards are not used, it's usually because there's
still a custom.sls file in the source tree that we don't want
to be installed.

Having `make install` means it's easy to try DeepSea on
arbitrary non-RPM-based distros (`git clone ; make install`)
and should also hopefully help with packaging for other distros.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,267 @@
+# Override this to install docs somewhere else
+DOCDIR = /usr/share/doc/packages
 
+usage:
+	@echo "Usage:"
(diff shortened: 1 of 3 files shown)


smithfarm:
+1 thanks, Tim!

oms4suse:
I like this also but have not tested it.

PR #37, opened 2016-10-28
Wip remove local includes

oms4suse:
Use absolute includes

Older salt versions need absolute include paths


tserong:
Included in #44

PR #39, opened 2016-10-28
Wip nettest support long minion

oms4suse:
Minion id's are needed not short id's

To support minion id's correctly rather than use short names, we must use the
'host=False' argument.


tserong:
Included in #44

PR #43, opened 2016-11-02
Rolling update & upgrade

jschmid1:
This serves the purpose of solving the problem of updating & rebooting. [0] -> detailed explanation

With this decoupled approach we update the kernel & reboot _per-role_ sequentially and in correct order ( mon -> osd .. etc ) and hardfail if something goes wrong.

Now you safely run ( for now called maintenance - naming needs to be straightened out though ) which installs everything BUT the kernel ( not implemented ) and does a graceful restart which was  previous implemented with `ceph.restart`.

[0]: 
If you currently run stage0 it will first install all updates and subsequently check if a kernel update was applied. If so DeepSea initializes a reboot. The issue with this is that you don't know whether you have a new ceph-* binary added to your system. In a larger scale cluster you want to run on different versions of ceph as short as possible. Reboots might take up to 15 minutes per node which means you end up with different versions of ceph-* for a rather long time.

* don't merge this yet. - WIP


jschmid1:
installing everything BUT the kernel might be parsed inside jinja with something like:

`zypper lu -a | awk '{print $5}' | xargs sudo zypper up -D`

but this might actually be a good candidate for a module as I can see it failing already.

jschmid1:
So we will most likely end up with something like this:

**Phase1)**
Take a host in the pool of roles:monitor
-  In case the cluster is health:
-  apply a kernel update
-  reboot the node, if there was a kernel update
-  move to the next node

Repeat for all other roles -> osd -> rgw -> mds -> igw

you now have applied the kernel updates _without downtime._

**Phase2)** 
Take a host in the pool of roles:monitor
- In case the cluster is health:
- apply all maintenance update
- restart the service, if its binary ( ceph-mon ) was updated
- move to the next node

Repeat for all other roles -> osd -> rgw -> mds -> igw

you are now running the latest version of ceph & kernel _without downtime._


In the grand scheme of things admins want to run this periodically and automated.

jschmid1:
the restarting condition for each service might look like this:

```
{% set installed_version = salt['cmd.run']('rpm -q --last ceph-mon | head -1 | cut -f1 -d+| grep -Po '(\d+\.)?(\d+\.)?(\*|\d+)'') %}
{% set used_version = salt['cmd.run']('ceph-mon --version | head -1 | cut -f1 -d- | grep -Po '(\d+\.)?(\d+\.)?(\*|\d+)'') %}
```
with the according unless block:
```
 - unless: "systemctl is-failed ceph-mon@{{ grains['host'] }}.service ||  if [[ {{ installed_version }} == {{ used_version }} ]]"
```

jschmid1:
any comments on that?

l-mb:
I'm not sure I agree with this particular approach. Precisely because you want to reduce the amount of reboots. Also, the updates might depend on each other - e.g., a new ceph-* package depends on a glibc update, iSCSI requires a new kernel version, etc.

I believe iterating over the nodes in the right sequence and upgrading them entirely in one step (stop, dup, restart, reboot if necessary) is the saner choice.

15 minutes really is not much. Two reboots are worse. (We'll have customers upgrading the nodes one by one over days, I guarantee it.)

jschmid1:
Thanks for the feedback!

I think Eric's initial plan was to run Stage 0-5 in a regular fashion ( 0..10 times a day). If you follow that path my proposed solution makes sense because minor updates won't effect ceph's ability to function properly hence no action is needed. That path acutually reduces the amount of reboots.

And due to that assumption we decided to go for a differentiated approach of only applying a restart or a reboot if it's actually needed. Your point on dependencies is correct, though. We have to agree on packages that will cause a reboot - Currently it's only a kernel update.

If you for example start the update process at the monitors the following scenario might eventuate:

If you followed the solution of just `zypper dup` you will end up rebooting the nodes anyways. So nothing is gained here. The crucial point here is also that `dup` might have installed a newer version of ceph-*. The reboot 'activates' this version - as the reboot might take up to 15 minutes, the version will be out of sync and perhaps cause some problems.

If you only installed - if there is - the latest kernel & reboot the machine, nothing serious happens and the ceph versions are still in sync. Now you can safely apply the remaining updates, which won't cause any reboots, and sequentially restart the ceph services.

l-mb:
Services might fail at any time anyway and then be restarted on the newer version. That's something that must work, otherwise we've got a potential problem. Or they could reboot for whatever reason.

Restarting individual services on a node is pretty much a question of luck (unless containers were to be used). I don't think we need to over-optimize for this case.

As for the reboot detection, zypper actually knows after the update has been applied:
       102 - ZYPPER_EXIT_INF_REBOOT_NEEDED
           Returned after a successful installation of a patch which requires reboot of computer.

On other distros, this causes /var/run/reboot-required, not sure if this also happens on SUSE.

jschmid1:
> Services might fail at any time anyway and then be restarted on the newer version. That's something that must work, otherwise we've got a potential problem. Or they could reboot for whatever reason.
> Restarting individual services on a node is pretty much a question of luck (unless containers were to be used). I don't think we need to over-optimize for this case.

If that's the case I'd support you approach aswell.

> As for the reboot detection, zypper actually knows after the update has been applied:
102 - ZYPPER_EXIT_INF_REBOOT_NEEDED
Returned after a successful installation of a patch which requires reboot of computer.

>On other distros, this causes /var/run/reboot-required, not sure if this also happens on SUSE.

That is indeed pretty helpful, thanks for that valuable information.

So what you propose is splitting the 'restart' and the 'update' part, right?

If you as a customer decide to update we iterate over the roles/nodes and let zypper decide if a reboot is necessary.

The `ceph.restart` procedure will mostly be used for manual intervention or for a intentional restart after a PTF or similar.

l-mb:
> So what you propose is splitting the 'restart' and the 'update' part, right?

Perhaps. What I'd do would be to:
- Build a list of all nodes, ordered { MON; MDS; OSD; iSCSI gateways; radosgw }
  - If a node is already in the list, because it co-hosted a higher priority service, don't add it again
  - Co-hosted services will restart some services earlier than optimal, but without dedicated nodes or containers, that is effectively impossible to prevent in all cases; and if it can happen, we need to test it and make sure that it works. The best way to do that is to make it the regular case ;-)
  - I just realize that based on the previous comment we might as well randomize the order in which we upgrade the nodes. Must work, right?

- Don't start unless HEALTH_OK

- Determine if there are any updates at all that'd be applied.
  - Skip node if not (already uptodate)
- Stop all services on a node
- Apply all updates
- Determine whether this requires a reboot (based on zypper feedback)
  - If zypper feedback is not sufficient, indeed we can look for newer kernels, modules, glibc etc. I'd suggest to have a whitelist of package patterns to trigger reboots then.
  - Reboot if so
- Restart Ceph services
  - which the reboot ought to do automatically?
- Wait until HEALTH_OK
- Rinse & repeat

jschmid1:
One thing I simply don't know is what could possibly happen if ceph runs on different versions for a longer period of time.Going with the all-in-one update and reboot strategy could theoretically cause a cluster to be out of sync(version-wise) for some hours.
My best guess is that it _depends_ if that is an actual problem or not(major vs minor version update). But making that decision is hard again.

Secondly I'd like know what the advantage 

> * Stop all services on a node 
> 
> * Apply all updates
> [..]
>  * Restart Ceph services

over:

* Apply all updates

* Check if a restart/_boot_ is actually needed.
  * Restart/Reboot

That would potentially save one round of restarts + a bit of cluster shakiness..

But If you know any potential problems that might occur with updating while the process is running, this might be a tradeoff we have to accept.


> I just realize that based on the previous comment we might as well randomize the order in which we upgrade the nodes. Must work, right?

Well, true. Going with the update-reboot solution we loose the ability to control the order in which we restart services to some extend.. We might need some practical feedback/testing/eval here.

smithfarm:
> One thing I simply don't know is what could possibly happen if ceph runs on different versions for a longer period of time.Going with the all-in-one update and reboot strategy could theoretically cause a cluster to be out of sync(version-wise) for some hours.

That should not be any problem at all. Ceph undergoes extensive upgrade testing upstream.

Another question is whether the customers will be prepared (psychologically speaking) for the update to take several hours.

jschmid1:
> That should not be any problem at all. Ceph undergoes extensive upgrade testing upstream. 

Thats good to hear.

> Another question is whether the customers will be prepared (psychologically speaking) for the update to take several hours.

I guess we need to document the process very precisely to make sure they understand why it might take so long.

@smithfarm Does upstream also test a random order? That is not strating as recommended with the MON->OSD .. ?

smithfarm:
No, it's not random. But they do test different orders. See e.g. https://github.com/ceph/ceph/blob/master/qa/suites/upgrade/jewel-x/parallel/3-upgrade-sequence/upgrade-mon-osd-mds.yaml

jschmid1:
great, thanks for the hint. 

So the last question is whether to:

* stop services -> update -> start 
or
* update -> maybe restart

swiftgist:
@l-mb @smithfarm  Just for clarification, when either of you say "random" order, both of you are okay with a standard approach of monitors, then storage, then remaining in general?  For the Ceph cluster with dedicated nodes, there's no real issue even with reboots.  

My concern is when things go bad, how does that leave the cluster and what is the admin left holding?  
I think I'd rather leave an admin with "Yeah, the monitors upgraded fine as did most of the storage nodes, but we're in a catch-22 with some of the gateways on shared storage nodes" is better than "Well, some monitors upgraded, some storage nodes upgraded, one igw is working, another isn't, some other services partially upgraded too, but  the upgrade stopped".  The number of conversations asking which nodes were upgraded and that the admin needs to create a scorecard for management and support isn't really a selling point.  

I guess six partial upgrades leaves me edgy, where as two complete upgrades followed by a failure is easy to describe.

l-mb:
@swiftgist Trying to order the nodes is fine, but ordering them is probably not a hard mandatory aspect.

If the update fails for any node, flag that node, abort the update sequence? And once the admin has manually resolved whatever problem occurred, they can just restart the update process? We don't even need to start in the middle, since upgrading an already upgraded node is effectively a no-op and would just be skipped anyway.

jschmid1:
That pullrequest tries to archive:

* abstract the packagemanager calls to a module to make deepsea portable
* handle the reaction to packagemanagers returncodes in that module
  * srv/salt/_modules/packagemanager.py
* granularize updates (packages vs kernel)
  * srv/salt/ceph/updates/regular
  * srv/salt/ceph/updates/kernel
  * call them combined with ceph.updates
* add states for maintenance mode
   * updates
     * get nodes _ordered_ and _unique_ 
     * ask the pm module to do reboots
     * abort if the cluster cant recover from HEALTH_ERR in $timeout
   * upgrades
     * prevalidate checks (not yet)
     * get nodes _ordered_ and _unique_ 
     * zypper dup

* refactor stage0
  * when prep detects a running cluster execute 'maintenance mode update'
  * if not call ceph.updates in parallell

tests/unit/_modules/test_packagemanager.py
```diff
@@ -0,0 +1,345 @@
...
+    @mock.patch('srv.salt._modules.packagemanager.linux_distribution')
+    @mock.patch('srv.salt._modules.packagemanager.Popen')
+    def test_reboot(self, po, dist_return):
+        """
+        Reboot method
+        """
+        dist_return.return_value = ('opensuse', '42.2', 'x86_64')
+        ret = PackageManager()
+        ret.reboot_in()
+        assert po.called is True
```
BlaineEXE: If I'm reading this correctly, this is only asserting that Popen is called, but it doesn't specify what Popen should be doing. For a reboot, we want to assert that Popen is called and that the command issued is a reboot, otherwise this test isn't very explicit.

I see this elsewhere in this test module too.

srv/salt/_modules/packagemanager.py
```diff
@@ -0,0 +1,274 @@
...
+            self.pm = Apt(**kwargs)
+        else:
+            raise ValueError("Failed to detect PackageManager for OS."
+                             "Open an issue on github.com/SUSE/DeepSea")
+
+    def reboot_in(self):
+        """
+        Assuming `shutdown -r` works on all platforms
+        """
+        log.info("The PackageManager asked for a systemreboot. Rebooting in 1 Minute")
```
BlaineEXE: "system reboot"

Also, I don't see in the code anywhere we wait for 1 minute before rebooting.

srv/salt/_modules/packagemanager.py
```diff
@@ -0,0 +1,274 @@
...
+            log.info("Found {}. Using {}".format(self.platform, Zypper.__name__))
+            self.pm = Zypper(**kwargs)
+        elif "fedora" in self.platform or "centos" in self.platform:
+            log.info("Found {}. Using {}".format(self.platform, Apt.__name__))
+            self.pm = Apt(**kwargs)
+        else:
+            raise ValueError("Failed to detect PackageManager for OS."
+                             "Open an issue on github.com/SUSE/DeepSea")
+
+    def reboot_in(self):
```
BlaineEXE: By the convention used for Apt and Zypper, I think this should be `_reboot_in(self)`. These classes don't seem to have a public API, or rather, the public API seems to be `_handle`.

Also, why is it `reboot_in` and not just `reboot`?

tests/unit/_modules/test_packagemanager.py
```diff
@@ -0,0 +1,345 @@
...
+        """
+        self.linux_dist = patch('srv.salt._modules.packagemanager.linux_distribution')
+        self.lnx_dist_object  = self.linux_dist.start()
+        self.lnx_dist_object.return_value = ('opensuse', '42.2', 'x86_64')
+        args = {'debug': False, 'kernel': False, 'reboot': False}
+        yield PackageManager(**args).pm
+        self.linux_dist.stop()
+
+    @mock.patch('srv.salt._modules.packagemanager.Popen')
+    def test__refresh(self, po, zypp):
```
BlaineEXE: For testing Zypper, I would only test the public API, and I would assert that all the correct Popen commands are called with the correct strings without testing private methods individually. What we are wanting to test is that the package manager will execute the correct commands in the correct order, not merely that it is executing any command and where order isn't explicitly tested. It seems the class has no public API, but the `up` and `dup` functions use `_handle`, so I think we can treat `_handle` as though it's the public API. And also `reboot_in` can be tested on its own, just make sure that `_handle` calls `reboot_in` at the right time.

BlaineEXE:
As an aside, and definitely for another pull request, I think the user should be able to specify somewhere explicitly that they do or don't want to update their kernel. Some customers may standardize on a particular kernel or even have a custom kernel they want to run on their servers. If we update the kernel out from under these customers without telling them, I can see someone being very upset. I conjecture that a worst-possible case might even be that this could put them at risk of legal non-compliance if the kernel they have standardized on is for compliance reasons.

srv/modules/runners/orderednodes.py
```diff
@@ -0,0 +1,38 @@
...
+    helper.
+    """
+    seen = set()
+    return [x for x in seq if x not in seen and not seen.add(x)]
+
+
+def unique(cluster=None, exclude=[]):
+    """ 
+    Assembling a list of nodes.
+    Ordered(MON < OSD < MDS < RGW < IGW)
```
jschmid1: That's the suggested order of upgrading nodes/services

Arrows would've been less confusing and more descriptive

srv/modules/runners/orderednodes.py
```diff
@@ -0,0 +1,38 @@
...
+    `list(set(list_in_question))`
+    This method will mess with the sorting though.
+    As we rely on the sorting in this scenario, we have to use this
+    helper.
+    """
+    seen = set()
+    return [x for x in seq if x not in seen and not seen.add(x)]
+
+
+def unique(cluster=None, exclude=[]):
```
jschmid1: cluster='ceph' will most likely be better as, you pointed out, None won't yield anything.

I'd still like to keep the varialbe with a default as it makes all the invocations easier to read -  especially where they appear most of the time - in jinja files.

`% for host in salt.saltutil.runner('orderednodes.unique', cluster='ceph') %}`

as opposed to

`% for host in salt.saltutil.runner('orderednodes.unique', 'ceph') %}`

srv/modules/runners/orderednodes.py
```diff
@@ -0,0 +1,38 @@
...
+
+    cluster_assignment = "I@cluster:{}".format(cluster)
+    roles = ['mon', 'storage', 'mds', 'rgw', 'igw', 'ganesha']
+    # Adding an exclude param here to allow skipping of individual
+    # roles. 
+    # Usecase: If an admin wants to have manual control over the upgrade
+    # process in missioncritical connections like iscsi where one needs
+    # to be sure that the MPIO successfully failed over before re booting/starting
+    # the service.
+    roles = [role for role in roles if role not in exclude]
```
jschmid1: `return _preserve_order_sorted(all_clients)`

-> `return [x for x in seq if x not in seen and not seen.add(x)]`

will get rid of duplicates with the sideeffect that nodes that host more than one service will not be in the right order.

srv/salt/_modules/advise.py
```diff
@@ -29,6 +29,14 @@ def reboot(running, installed):
 
     return True
 
+def generic(message):
+    """
+    Used to print arbitrary text to the screen.
+    """
+    message = message
```
jschmid1: c&p mistake. but good point - will update that

srv/salt/_modules/advise.py
```diff
@@ -29,6 +29,14 @@ def reboot(running, installed):
...
+    log.info(message)
 
+    proc = Popen([ "/usr/bin/wall" ], stdin=PIPE, stdout=PIPE, stderr=PIPE)
+    output = proc.communicate(input=message)
```
jschmid1: `proc.communicate(input=message)`

with message being one of [None, False, ""]

`proc.returncode` is always 0

I'll get rid of the assignment

srv/modules/runners/orderednodes.py
```diff
@@ -0,0 +1,38 @@
...
+    for role in roles:
+         dct = client.cmd("I@roles:{} and {}".format(role, cluster_assignment), 'pillar.get', ['roles'], expr_form="compound")
```
jschmid1: changed

srv/salt/_modules/packagemanager.py
```diff
@@ -0,0 +1,274 @@
...
+import logging
+import os
+
+log = logging.getLogger(__name__)
+
+
+class PackageManager(object):
+
+    """
+    That is not the nativ salt module and is not ment to
```
jschmid1: fixed

srv/salt/_modules/packagemanager.py
```diff
@@ -0,0 +1,274 @@
...
+    replace it.
+    This module was created to react on packagemanagers
```
jschmid1: true, sounds better.

srv/salt/_modules/packagemanager.py
```diff
@@ -0,0 +1,274 @@
...
+    reboot advice.
+    """
```
jschmid1: re 1) I guess I kept it there for other devs wondering why the hell we're reinvented the wheel.

re 2) yes, that is exactly the point where contributing to upstream will help us in the long run. lack of time prohibited me from doing that, yet. 

I'd *love* to replace this whole module with the pkg.* salt-builtins

srv/salt/_modules/packagemanager.py
```diff
@@ -0,0 +1,274 @@
...
+    def __init__(self, **kwargs):
+        self.debug = kwargs.get('debug', False)
+        self.kernel = kwargs.get('kernel', False)
+        self.reboot = kwargs.get('reboot', True)
+
+        self.platform = linux_distribution()[0].lower()
+        if "suse" in self.platform or "opensuse" in self.platform:
+            log.info("Found {}. Using {}".format(self.platform, Zypper.__name__))
+            self.pm = Zypper(**kwargs)
+        elif "fedora" in self.platform or "centos" in self.platform:
```
jschmid1: err, yes. good catch

srv/salt/_modules/packagemanager.py
jschmid1: I initially planned to make the reboot delay-time flexible as in:

```python
def reboot_in(self, time=1):
    cmd = "shutdown -r +{}".format(time)
```

but I don't see the use-case any longer.  -> changing back to _reboot(self)

srv/salt/_modules/packagemanager.py
jschmid1: from `man shutdown(8)`

[..]  If no time argument is specified, "+1" is implied.

srv/salt/ceph/cephprocesses/default.sls
```diff
@@ -0,0 +1,5 @@
+wait processes:
+ module.run:
```
jschmid1: done

srv/modules/runners/orderednodes.py
BlaineEXE: I think the format is fine. I read it as intended. I meant that the order of "NFS" isn't specified.

srv/salt/_modules/packagemanager.py
BlaineEXE: After looking at the code more, I do see why 'reinventing the wheel' is actually necessary here. Perhaps we continue as-is for now and make a note to push the functionality upstream into salt when we have cycles?

tests/unit/runners/test_orderednodes.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+from mock import patch, MagicMock
+from srv.modules.runners import orderednodes
+
+
+
+class TestOrderedNodes():
+
+    def test__preserve_order_sorted(self):
```
jschmid1: it's already there now :) doesn't hurt -> I'll keep it

tests/unit/_modules/test_packagemanager.py
jschmid1: I know that it's not __needed__ to test private methods, as you only want to validate the "output". I think in our case though it's not enough to validate the "output" because `_handle` can take multiple routes and calls methods on it's way that will try to bring the system in a certain state.
In general, I agree with the "only test the public API" pragma, in this case though I think that we should also validate the underlying private methods.

feedback/counter-arguments welcome

BlaineEXE: I agree that test this class seems kind of ... sticky. Perhaps a good middle ground is to test the methods as you have here (keep everything) and also make sure there is a test that ensures the private methods are called in the order we expect in appropriate cases. This way we can verify that each bit is doing the right thing and we can verify that the overall is doing the steps we expect in the right order.

tests/unit/_modules/test_packagemanager.py
```diff
@@ -0,0 +1,345 @@
...
+        ret.reboot_in()
+        assert po.called is True
+
+    @mock.patch('srv.salt._modules.packagemanager.linux_distribution')
+    @mock.patch('srv.salt._modules.packagemanager.Popen')
+    def test_not_implemented(self, po, dist_return):
+        """
+        Your platform is not supported
+        """
+        dist_return.return_value = ('ScientificLinux', '42.2', 'x86_64')
```
BlaineEXE: So it's like an apt wrapper for zypper? Hmm ...

That's what it seems: https://software.opensuse.org/package/zypper-aptitude

My initial thoughts are that I think we still might be fine if the system reports apt for this case sense we should theoretically (in practice may be different) just use apt commands and they get sent to zypper. It's possible that detecting when a reboot is necessary via zypper-aptitude could not follow apt's normal procedures, though. It's probably not worth risking using apt when zypper is available. Thinking a little more, the order of the statements as I wrote them is such that if a system has zypper and zypper-aptitude, it will still report zypper since that check is first.

On the other hand ... a project called pacapt exists, and it uses distro to determine package manager. See code linked. Maybe this really is a better option?: https://github.com/icy/pacapt/blob/ng/lib/00_core.sh#L61
There is a comment in the code talking about Fedora 22 moving to dnf, which is also something we'd have to support.

More research: My dev machine has zypper-aptitude installed, and this is the apt-get version.
```sh
$> apt-get --version
zypper 1.13.21
```

BlaineEXE: I still think that checking for the existence of a package manager is the better option provided we check for zypper before apt. I think this will be able to support more systems with less code/work and will support the corner case where a distro changes from one package manager to another as Fedora has done more easily than hard-coding versions. I suspect this would make this particular code need less future maintenance. Do you think this logic holds up?

If this is what we choose, perhaps we should order them as below:
1) Check for zypper (before apt so we don't get a false positive with zypper-aptitude)
2) Check for dnf (before yum in case version of fedora have a yum wrapper for dnf, not yet implemented)
3) Check for yum (not yet implemented)
4) Check for apt (last in case there is a yum/dnf wrapper as with zypper)

srv/salt/ceph/stage/prep/minion/default.sls
```diff
@@ -19,23 +24,69 @@ common packages:
...
+
+updating {{ host }}:
+  salt.state:
+    - tgt: {{ host }}
+    - tgt_type: compound
     - sls: ceph.updates
+    - failhard: True
 
-restart:
+check if restart is needed for {{ host }}:
```
BlaineEXE: And then the API starts to get really verbose too, and the long URLs kind of look like crap. I think it's fine as-is.

BlaineEXE: I think I misinterpreted the code yesterday. The `PackageManager` class already does a reboot if the package manager requests it, yeah? And this reboot is not related to package installation?

jschmid1: > I think I misinterpreted the code yesterday. The PackageManager class already does a reboot if the package manager requests it, yeah?

Yes, with one caveat. currently `zypper up` is used, while only `zypper patch` yields the correct returncode in order to determine the need for a reboot. Will be changed with two (default, non-default sls files). 

> And this reboot is not related to package installation?

It is. Based on the returncode(for zypper) and on the /var/something/reboot-required on (deb based) the module will trigger a reboot. But the packages(or specfiles) themselves tell the `Packagemanager` if a reboot is required or not (not sure on that one).

srv/salt/_modules/packagemanager.py
```diff
@@ -0,0 +1,275 @@
...
+            proc = Popen(base_command, stdout=PIPE, stderr=PIPE)
+            stdout, stderr = proc.communicate()
+            for line in stdout:
+                log.info(line)
+            for line in stderr:
+                log.info(line)
+            log.info("returncode: {}".format(proc.returncode))
+            if proc.returncode == 0:
+                if os.path.isfile('/var/run/reboot-required'):
+                    self._reboot()
```
BlaineEXE: I don't have a total grasp of how all these pieces are fitting together, so I want to make sure I'm understanding it. From my understanding, I'm seeing a possibility for issuing mulitple reboots between this code (the `PackageManager` classes) and the sls file that does upgrades.

On this line, I see that when `up` or `dup` is called, the package manager (Apt in this case, but a similar line exists for Zypper patch) issues a reboot automatically.

In the SLS file, there is also a reboot line. When `PackageManager` issues a reboot, does it set `auto_reboot` in the pillar to `False` so there won't be a double reboot?

I know @jschmid1 also mentioned needing to check old vs new kernel version since zypper has some bugs/lack of reboot reporting to work around. Would it make sense to instead of issuing a reboot on this line to instead set the `auto_reboot` to `True` and let that be handled in the SLS?

tests/unit/_modules/test_packagemanager.py
jschmid1: > . I think this will be able to support more systems with less code/work and will support the corner case where a distro changes from one package manager to another as Fedora has done more easily than hard-coding versions. I suspect this would make this particular code need less future maintenance. Do you think this logic holds up?

Right, makes sense. But let's keep that in mind and rework that once the PR is in. (track that with an issue).

tests/unit/_modules/test_packagemanager.py
jschmid1: any suggestion on how to do that?

srv/salt/_modules/packagemanager.py
jschmid1: I added `autoreboot: False|True` to globally set the default behaviour on whether deepsea has the right to 'automagically reboot' in case the system requests it. So see it as a 'user variable' rather than an internal one.

My intention was to do it the way around. I mentioned in another comment that changing to 'zypper patch' will get rid of the ceph/update/reboot sls file. 

and so technically double reboots won't be possible in zypper right now, as the 107 will never be received (due to zypper _up_'s missing ret implementation). They may be with Apt but as `Packagemanager` will cause a reboot, the next step in the orchestrate file (after the reboot) is ceph.upgrade.reboot which compares the 'installed' vs 'used' kernel, which is the same after the reboot.

tests/unit/_modules/test_packagemanager.py
BlaineEXE: Sounds like a good plan.

tests/unit/_modules/test_packagemanager.py
BlaineEXE: I'd have to look up how to do it again using pytest, but I'm quite certain you can specify in a test that you expect a method to call specific functions in a specific order. My thought would be like what I have below. Does that seem appropriate?

```
# Tests for Zypper class
Given: an operation that does not request reboot
Expect: _handle calls methods in order and reboot method is not called

Given: an op that requests reboot
Expect: _handle calls methods in order and reboot is one of them

# Tests for Apt class
Etc. for Apt

# Future tests for yum/dnf/pacman/etc. follow same form
```

srv/salt/_modules/packagemanager.py
BlaineEXE: OH! Okay. I am following now. I didn't realize the `auto_reboot` was a gate to allow that to happen at all. #samepage 
Sounds all good. Zypper makes it non-ideal it seems, but I think I have a good idea of what's happening now. Thanks!

srv/modules/runners/orderednodes.py
BlaineEXE: I think it makes more sense to update a mon+rgw with the mons (as this would do) moreso than with the rgws. So I think this seems good. Unless anyone else has another thought.

swiftgist:
merged into wip-updates-and-restart

jschmid1:
merged with #222, closing

PR #44, opened 2016-11-07
Wip test nettests parallel

oms4suse:
Add nettests:

    ping
    iperf3
    iperf3-concurrent

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,7 @@ install:
 	install -m 644 srv/salt/ceph/events/*.sls $(DESTDIR)/srv/salt/ceph/events/
 	# state files - diagnose
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/diagnose
+	install -m 644 srv/salt/ceph/diagnose/*.md $(DESTDIR)/srv/salt/ceph/diagnose
 	install -m 644 srv/salt/ceph/diagnose/*.sls $(DESTDIR)/srv/salt/ceph/diagnose
 	# state files - igw
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/igw
--- a/Makefile
+++ b/Makefile
@@ -95,6 +95,7 @@ install:
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/iperf
 	install -m 644 srv/salt/ceph/iperf/*.sls $(DESTDIR)/srv/salt/ceph/iperf/
 	install -m 644 srv/salt/ceph/iperf/systemd-iperf.service $(DESTDIR)/srv/salt/ceph/iperf/
+	install -m 644 srv/salt/ceph/iperf/*.py $(DESTDIR)/srv/salt/ceph/iperf
 	# state files - mds
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/mds
 	install -m 644 srv/salt/ceph/mds/*.sls $(DESTDIR)/srv/salt/ceph/mds/
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,9 @@ install:
 	install -m 644 srv/salt/ceph/configuration/files/ceph.conf* $(DESTDIR)/srv/salt/ceph/configuration/files/
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/events
 	install -m 644 srv/salt/ceph/events/*.sls $(DESTDIR)/srv/salt/ceph/events/
+	# state files - diagnose
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/diagnose
+	install -m 644 srv/salt/ceph/diagnose/*.sls $(DESTDIR)/srv/salt/ceph/diagnose
 	# state files - igw
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/igw
 	install -m 644 srv/salt/ceph/igw/*.sls $(DESTDIR)/srv/salt/ceph/igw/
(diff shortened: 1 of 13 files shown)


tserong:
I've run ping, iperf3 and iperf3-concurrent tests from this PR on my little cluster of test VMs (SLE 12 SP2 + latest SES4 pieces), and it seems to work fine. I've got six nodes, one is the salt master, four are cluster nodes and one is unassigned (just to demonstrate that deepsea is leaving that node alone). The tests run against five nodes (the master and the four cluster nodes), as I've got the master tagged cluster-ceph in policy.cfg, even though it's not actually a cluster node (I wonder, could I have left the master unassigned?)

Anyway, time wise, for the record, if everything is working:

* ping completes in 18 seconds. The parallelism is done such that each IP address to ping is pinged in turn simultaneously by all other minions. With my test cluster, five IP addresses were pinged, with each parallel set of pings from the other four nodes taking slightly over three seconds, hence 5 * ~3.5 = 18 seconds in total. If I'd had two networks, I'd expect it to take 36 seconds.
* iperf3 completes in 3 minutes 42 seconds.
* iperf3-concurrent completes in 2 minutes 14 seconds.

There are some oddities in the output -- at the start, ping prints:

````
|_
  - ses4-4.example.com,ses4-3.example.com,ses4-0.example.com,ses4-2.example.com
  - 192.168.12.170
|_
  - ses4-1.example.com,ses4-3.example.com,ses4-0.example.com,ses4-2.example.com
  - 192.168.12.144
|_
  - ses4-1.example.com,ses4-4.example.com,ses4-0.example.com,ses4-2.example.com
  - 192.168.12.148
|_
  - ses4-1.example.com,ses4-4.example.com,ses4-3.example.com,ses4-2.example.com
  - 192.168.12.225
|_
  - ses4-1.example.com,ses4-4.example.com,ses4-3.example.com,ses4-0.example.com
  - 192.168.12.167
````

These are, of course, the hostnames in my cluster and the IP address to ping, but where's that coming from? Can it be removed somwhow? iperf and iperf-concurrent have similar chunks of stuff at the start.

Things seem to get a bit worse if something is broken. Just for fun, I ran `iptables -I INPUT -p ICMP -j DROP` on one of my nodes to break the ping test. This resulted in lots of "[ERROR   ] {'ret': {'ses4-4.example.com': 'PING 192.168.12.170 (192.168.12.170) 56(84) bytes of data.\n\n-" junk in the output prior to the regular salt output, and made the ping run take a bit over a minute (ping invocations with 100% packet loss take 13 seconds each, apparently). I'm not so worried about the time increase, but it'd be nice to get rid of that ERROR junk in the output, as it gets reported later anyway in the regular salt function output.

The ping output is easy enough to read and make sense of, I think.

The iperf3 and iperf3-concurrent output is a little more annoying, just because there's bits of service installation and whatnot reported by salt, which you don't really care about (the interesting bit is the iperf output itself). Not sure what can be done about that though, as that's salt being salt.

I'm not sure about the /usr/bin/ceph_diagnose_iperf_client.py helper script -- could that be implemented as an execution module or something instead? The we wouldn't need to have SLS to add and remove that script from each node.

oms4suse:
Sadly the issue with the output:

    |_
      - ses4-4.example.com,ses4-3.example.com,ses4-0.example.com,ses4-2.example.com
      - 192.168.12.170
    |_
      - ses4-1.example.com,ses4-3.example.com,ses4-0.example.com,ses4-2.example.com
      - 192.168.12.144
    |_
      - ses4-1.example.com,ses4-4.example.com,ses4-0.example.com,ses4-2.example.com
      - 192.168.12.148
    |_
      - ses4-1.example.com,ses4-4.example.com,ses4-3.example.com,ses4-2.example.com
      - 192.168.12.225
    |_
      - ses4-1.example.com,ses4-4.example.com,ses4-3.example.com,ses4-0.example.com
      - 192.168.12.167

Is a function of using a runner to generate the data in an orchestrate function. This issue happens all over deepsea as it is a "feature" of salt.

oms4suse:
> I'm  not sure about the /usr/bin/ceph_diagnose_iperf_client.py helper script -- could that be implemented as an execution module or something instead? The we wouldn't need to have SLS to add and remove that script from each node.

I don’t like this either. The options I have tried but have not succeeded are:

 * custom execution module : and failed to resolve this, but did not try this work around as I did not think it was appropriate for a packaged product - https://github.com/saltstack/salt/issues/31849

 * Using "cmd.script" rather than "cmd.run" : While "cmd.script" was found, this produced an "argument error" when ever I tried to use it from orchestrate, while it worked fine from a state.sls file.

> because there's bits of service installation and whatnot reported by salt, which you don't really care about 

With salt, you cant run background process very easily, and start and stop them, for this reason I use systemd sadly this does add stages to the process, and I see no way around this.

oms4suse:
> Things seem to get a bit worse if something is broken. Just for fun, I ran iptables -I INPUT -p ICMP -j DROP on one of my nodes to break the ping test. This resulted in lots of "[ERROR ] {'ret': {'ses4-4.example.com': 'PING 192.168.12.170 (192.168.12.170) 56(84) bytes of data.\n\n-" junk in the output prior to the regular salt output, and made the ping run take a bit over a minute (ping invocations with 100% packet loss take 13 seconds each, apparently). I'm not so worried about the time increase, but it'd be nice to get rid of that ERROR junk in the output, as it gets reported later anyway in the regular salt function output.

The "ERROR junk" in the output would require to use some custom script rather than raw ping. This would then create more of the the "iperf3 and iperf3-concurrent" state changes, which you and I also don't like.

This said for "iperf3-concurrent" we could improve the error handling output by forking iperf3, and processing the output rather than the simpler "os.execvp(iperf3, arguments)".

tserong:
Thanks for the explanation. I'd still like to see if there's somehow a way of getting rid of ceph_diagnose_iperf_client.py and cleaning up some of the output, but as you say, quite a bit of this is "features" of salt, and/or you already tried various other approaches which didn't succeed.

Bearing that in mind, I'm OK with merging this PR, as I think this is a useful tool, and the README explains how it works and what to expect when using it.

(We could maybe open an issue or two for the points raised in the previous couple of comments, in case we're able to figure out a way of addressing them later)

Issue #50, opened 2016-11-24
Jijna error with salt version 2016.3.4 


supriti:
This error is not seen with salt version salt 2015.8.10. With salt version 2016.3.4, running deepsea stage 3 gives the following error: 

admin.ceph_master:
    Data failed to compile:
----------
    Rendering SLS 'base:ceph.stage.3.default' failed: Jinja error: runner() got multiple values for keyword argument 'name'
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/utils/templates.py", line 368, in render_jinja_tmpl
    output = template.render(**decoded_context)
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 989, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 754, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "<template>", line 4, in top-level template code
TypeError: runner() got multiple values for keyword argument 'name'
 
; line 4
 
---
 
{% set FAIL_ON_WARNING = salt['pillar.get']('FAIL_ON_WARNING', 'True') %}
 
{% if salt['saltutil.runner']('ready.check', name='ceph', fail_on_warning=FAIL_ON_WARNING)  == False %}    <======================
ready check failed:
  salt.state:
    - name: "Fail on Warning is True"
    - tgt: {{ salt['pillar.get']('master_minion') }}
    - failhard: True
[...]
---
retcode:
    1

jan--f:
doesn't happen with  2015.8.7 (Beryllium)

rjfd:
I think the problem is that when salt calls a runner, in the new version, the parameter `name` is reserved for internal use. To fix this, we just need to rename the first parameter of the `ready.check` and `validate.pillar` functions.

I'll open a PR with a fix.

PR #51, opened 2016-11-25
Fix runner parameter names in newer salt version (Boron)

rjfd:
The first parameter of a runner function cannot be named "name",
as it is used internally by salt.

Fixes: #50

### DIFF
--- a/srv/modules/runners/ready.py
+++ b/srv/modules/runners/ready.py
@@ -76,16 +76,16 @@ class Checks(object):
             print "{:25}: {}{}{}{}".format(attr, bcolors.BOLD, bcolors.WARNING, self.warnings[attr], bcolors.ENDC)
 
 
-def check(name, fail_on_warning=True, **kwargs):
+def check(cluster, fail_on_warning=True, **kwargs):
     """
     Check a cluster for runtime configurations that may cause issues for an
     installation.
     """
-    if name == None:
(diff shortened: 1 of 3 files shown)


rjfd:
@jan--f Can you test this please.

srv/modules/runners/ready.py
```diff
@@ -76,16 +76,16 @@ def report(self):
...
 
-def check(name, fail_on_warning=True, **kwargs):
+def check(cluster_name, fail_on_warning=True, **kwargs):
     """
     Check a cluster for runtime configurations that may cause issues for an
     installation.
     """
-    if name == None:
-        name = kwargs['name']
+    if cluster_name == None:
```
theanalyst: let's make this `cluster_name is None`

rjfd: done

jan--f:
Works for me under salt 2016.3.4 (Boron). @supriti can you check if it works for you now too please.

Maybe just 'cluster' would be better as the parameter? This would keep things consistent with the pillar data in DeepSea.

rjfd:
@jan--f you're right. Will change it to `cluster` only.

rjfd:
@jan--f pushed the rename changes. Might be worth to test again to check if I didn't screwed up anything the with the rename.

PR #54, opened 2016-11-29
Adding nfs-ganesha state for cephfs in deepsea

supriti:
In this commit: 
- Role "ganesha" is defined. Currently it is defined only for cephfs not for rgw. 
- nfs-ganesha and other packages required to run ganesha service is installed on the minion with ganesha role.
- nfs-Ganesha and nfs-ganesha-ceph package installs two conf files "ganesha.conf" and "ceph.conf" at /etc/ganesha. systemctl by default use ganesha.conf. Hence, we create a symlink for ceph.conf. 

Its still WIP. It requires at least the following changes to make it complete: 
- Better error and state handling.
- Adding more fields in ceph.conf for better flexibility
- Checking if cephfs is mounted on the server acting as nfs-ganesha server before doing anything else.
- Understanding if to extend this code further for rgw or to define a completely new role. 

Attaching two docs on how to manually setup nfs-ganesha for cephfs and rgw (the steps we are trying to automate).
[Nfs-Ganesha-CephFS.pdf](https://github.com/SUSE/DeepSea/files/619042/Nfs-Ganesha-CephFS.pdf)
[NFS-Ganesha_RGW.pdf](https://github.com/SUSE/DeepSea/files/619041/NFS-Ganesha_RGW.pdf)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,13 @@ install:
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/diagnose
 	install -m 644 srv/salt/ceph/diagnose/*.md $(DESTDIR)/srv/salt/ceph/diagnose
 	install -m 644 srv/salt/ceph/diagnose/*.sls $(DESTDIR)/srv/salt/ceph/diagnose
+	# state files - ganesha
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/ganesha
+	install -m 644 srv/salt/ceph/ganesha/*.sls $(DESTDIR)/srv/salt/ceph/ganesha/
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/ganesha/configure
+	install -m 644 srv/salt/ceph/ganesha/configure/*.sls $(DESTDIR)/srv/salt/ceph/ganesha/configure/
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/ganesha/install
+	install -m 644 srv/salt/ceph/ganesha/install/*.sls $(DESTDIR)/srv/salt/ceph/ganesha/install/
 	# state files - igw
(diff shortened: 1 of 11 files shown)


jan--f:
The history looks much better now and thanks for the quick doc overview.

I would add a couple of things to your list of changes for the furture:

- actually mount CephFS on the ganesha minions. What would be a good place for this?
- We should probably also create a designated keyring for the cephfs mount to be used by ganesha. This way we could use cephx to lock ganesha to certain data pools and paths.

Also how is client authentication handled on the gnaesha side? Are we going to provide DeepSea support fir this too?

denisok:
@jan--f not sure know about mounting CephFS. Why might we need that?

It is probably good separate state to mount CephFS (if we don't have it yet) that could be used on client nodes or if it needed on any server side node ...

supriti:
Ganesha requires cephfs to be mounted on the client where it is running. Then it exports that mounted Cephfs. For current commit, I am not mounting or checking if its mounted. It just starts ganesha server on assigned minion.

supriti:
Please ignore my last comment. We don't need to mount cephfs on the same minion as ganesha server. The only requirement is to have mds server, which is achieved by stage 4. Running "salt-run state.orch ceph.stage.ganesha" after stage 4 will start ganesha server. Its the client side role to export it further, which we don't cover with this commit.

jan--f:
I see...then consider my comments retracted too.

CephFS deployment is handled by stage 4 yes. We should probably check here is it was deployed at all though.

supriti:
[NFS-Ganesha+Deepsea](https://trello.com/c/9AEIJCne/2246-nfs-ganesha-deepsea)

theanalyst:
Adding #79 to track current changes

swiftgist:
Work continuing in #79

Issue #60, opened 2016-12-07
stage 0: kernel update requires "kernel-default" package to be installed


rjfd:
I was deploying DeepSea in a SLE12_SP2 image and when running stage 0 it stopped and restarted the node to update the kernel. The problem was that when I ran stage 0 again it tried again to install an updated version of the kernel.

After some investigation I found that the problem is due to the fact that `kernel-default` package is not installed, instead it is installed the `kernel-default-base` package. And in this case stage 0 will try to update the `kernel-default` package without any effect.

My main question to this issue is, do we really require that the host needs to have the `kernel-default` installed? or we can live with having only the `kernel-default-base`?

Either way, we need to fix stage 0 to check if `kernel-default` is installed to decide if it updated or installs, and avoid entering in a r"estart/run stage 0" loop

l-mb:
stage0 should make sure SLES is fully uptodate, not just the kernel. Then it doesn't matter what the kernel package name is.

rjfd:
@l-mb how do we check that "SLES is fully uptodate"?

swiftgist:
@rfd yeah, I ran into the same issue.  I have a fix in eric-sle12sp2.

@lmb the kernel seems to be a special case using zypper directly.  I am concerned about how to convert these to the Salt equivalents using the zypper salt module.  (As in, I don't think there's an equivalent currently.)

smithfarm:
I think @rjfd is using a JeOS image. This is good because it exposes implicit depenencies in the code.

rjfd:
@swiftgist @l-mb so regarding my question, does the kernel package influences anything in DeepSea, i.e., do we really need to have the uptodate version of `kernel-default` package, or we can have the uptodate version of `kernel-default-base` package?

swiftgist:
@rfd In general, we should have an up to date kernel.  In SES3, we had specific features for iSCSI that were not available in the base kernel.  I can see that happening again and it's somewhat a complicated issue to make behave gracefully for the customer in all cases.  

@smithfarm we're both using the JeOS which comes with kernel-default-base.  I ran into the same issue, but had only reached the point of making a branch.

rjfd:
@swiftgist does iSCSI in SES4 needs something not available in `kernel-default-base` package in SLE12_SP2?

smithfarm:
It would be nice to have an option to get along with just `kernel-default-base`, if possible, to avoid unnecessarily loading up the test VM.

swiftgist:
[...]

PR #61, opened 2016-12-07
enforce kernel-default

swiftgist:

### DIFF
--- a/srv/salt/ceph/updates/default-zypper-kernel-default.sls
+++ b/srv/salt/ceph/updates/default-zypper-kernel-default.sls
@@ -1,0 +1,22 @@
+
+
+zypper update:
+  cmd.run:
+    - name: "zypper --non-interactive  update --replacefiles --auto-agree-with-licenses"
+    - shell: /bin/bash
+    - unless: "zypper lu | grep -sq 'No updates found'"
+
+kernel install:
+  cmd.run:
+    - name: "zypper --non-interactive --no-gpg-checks install --force-resolution kernel-default"
+    - shell: /bin/bash
+    - unless: "rpm -q kernel-default"
+    - fire_event: True
+
+kernel update:
+  cmd.run:
+    - name: "zypper --non-interactive --no-gpg-checks up kernel-default"
+    - shell: /bin/bash
+    - unless: "zypper info kernel-default | grep -q '^Status: up-to-date'"
+    - fire_event: True
+


smithfarm:
Marking DNM based on discussion in #60

srv/salt/ceph/updates/default-zypper-kernel-default.sls
```diff
@@ -0,0 +1,22 @@
...
+    - unless: "zypper lu | grep -sq 'No updates found'"
+
+kernel install:
+  cmd.run:
+    - name: "zypper --non-interactive --no-gpg-checks install --force-resolution kernel-default"
+    - shell: /bin/bash
+    - unless: "rpm -q kernel-default"
+    - fire_event: True
+
+kernel update:
```
rjfd: why do you need this rule/step to update the kernel? the above `zypper update` already updates all installed packages even the kernel packages

swiftgist: @rjfd With respect to the issue in  #60, I can but it's not at the top of the list for me.  I am blocked because I do not have a Salt way of changing the kernel.

As for the 'rpm -q' is the least expensive way to make this idempotent.  The zypper commands quite literally take longer even when doing nothing and output results that tells Salt that something has changed.  The ultimate goal is that all of this will "go away" and be handled by the Salt zypper module.  In the meantime, I need a method to provision my vagrant box with the desired kernel and since I am sharing this with others, they will also need this so I do not think I should keep the change to myself.

jan--f:
lgmt!

For the default case:
Isn't only runing `zypper update` in https://github.com/SUSE/DeepSea/blob/eric-sle12sp2/srv/salt/ceph/updates/default.sls enough? Or is there another motivation for updating `kernel-default` explicitly?

jan--f:
@rjfd what do you think? any objections to a merge?

rjfd:
@jan--f yes, I don't want to merge this PR while issue #60 is still being discussed. 

Also, @swiftgist didn't reply to your question, which I also did in my review, and I want to know his opinion.

swiftgist:
@rjfd the zypper update did not update the kernel when I started working on this a few months back, so I added the explicit call which has been hardcoded to kernel-default this entire time.  From what everybody has expressed in #60, I do not think the current default.sls satisfies what is desired by the group.

However, these are new requirements to some degree and this does not need to be the end-all-be-all PR.  This PR will keep me and others that will use the vagrant box (or any JeOS image for SLE12SP2) working.  If everything must come to a halt because we do not have a consensus on a default, then DeepSea will never be released again.  This PR is to avoid changing the default.

rjfd:
@swiftgist I'm sorry, but I don't understand why this is blocking anyone. You can either update your Vagrant box by installing the `kernel-default` package, or everyone just point to this branch when testing DeepSea in their environments.

I'm using JeOS image in OVH with DeepSea, and I'm not blocked.

I don't think the master branch should be filled with commits that can potentially be reverted in the future.
But, if this is really blocking people from doing there work, @jan--f please merge this PR.

swiftgist:
@rjfd How many VMs are you running?  How many people will you be sharing your setup?  When you say "I can just update my Vagrant box by installing kernel-default".  That's what I was trying to do with DeepSea and not create a new image.  I was also trying to share this so others would not have to go through the same steps.

And as far as reverting, this commit will eventually be destroyed.  In an effort to keep DeepSea agnostic, the default.sls will hopefully be Salt modules with no specifics about OS.  However, that seems non-trivial at the moment.

PR #63, opened 2016-12-14
stage 0: update/install kernel

rjfd:
This PR contains the implementation of the third solution proposed in issue #60 .

I made a demo of how stage 0 behaves when there are hosts that do not have the `kernel-default` package installed or in use.

Demo url: https://github.com/rjfd/DeepSea/blob/wip-deepsea-kernel-default-demo/doc/demos/deepsea-kernel-default.gif

### DIFF
--- a/srv/modules/runners/kernel.py
+++ b/srv/modules/runners/kernel.py
@@ -1,0 +1,66 @@
+#!/usr/bin/python
+
+import logging
+import salt.client
+
+__log__ = logging.getLogger(__name__)
+
+def print_message(minion_list):
+    '''
+    Return a warning message about kernel-default package
+    not being installed in the system
+    '''
+    minion_list_str = ''
+    for minion in minion_list:
+        minion_list_str += ' {} \n'.format(minion)
+
+    message = '''
+############################################################
+ATTENTION: DeepSea detected that the following minion(s):
+{}
+are not running the kernel provided by kernel-default
+package.
+We strongly advise the user to use the kernel provided by
+the kernel-default package to ensure that all Ceph services
+will work correctly.
+
+If you want DeepSea to install kernel-default package,
+please rerun stage 0 as:
+
+$ CHANGE_KERNEL=YES salt-run state.orch ceph.stage.0
+
+Otherwise, if you want to keep the current running kernel
+at your own responsability, please rerun stage 0 as:
+
+$ KEEP_KERNEL=YES salt-run stage.orch ceph.stage.0
+############################################################
+    '''
+    return message.format(minion_list_str)
+
+def verify_kernel_installed(kernel_package, target_id):
+    '''
+    Verifies whether kernel_package is installed
+    Returns True if is is installed, and False otherwise
+    '''
+    __log__.debug("Verifying kernel_package=%s in target_id=%s",
+                  kernel_package, target_id)
+
+    minion_list = list()
+    local = salt.client.LocalClient()
+    res = local.cmd(target_id, 'cmd.run',
+                    ['zypper se -s kernel-* | grep "^i" | grep `uname -r | '
+                     'sed "s/-default//"` | cut -d"|" -f2'], expr_form='glob')
+    if len(res) == 0:
+        __log__.warn('Could not verify if %s package package is installed',
+                     kernel_package)
+        return minion_list
+
+    for minion in res.keys():
+        output = res[minion].strip()
+        __log__.debug('OUTPUT: %s', output)
+
+        if output != kernel_package:
+            minion_list.append(minion)
+
+    return minion_list
+
--- a/srv/salt/_modules/kernel.py
+++ b/srv/salt/_modules/kernel.py
@@ -1,0 +1,49 @@
+#!/usr/bin/python
+
+import logging
+import re
+from subprocess import Popen, PIPE
+
+__log__ = logging.getLogger(__name__)
+
+def verify_kernel_installed(kernel_package):
+    '''
+    Verifies whether kernel_package is installed
+    Returns True if is is installed, and False otherwise
+    '''
+
+    __log__.debug("Verifying kernel_package %s", kernel_package)
+
+    # zypper se -s kernel-* | grep "^i" | grep `uname -r | \
+    #                         sed "s/-default//"` | cut -d"|" -f2
+
+    proc = Popen(['uname', '-r'], stdout=PIPE)
+    uname_out = proc.stdout.read().strip()
+    match = re.search(r'(.+)-default', uname_out)
+    if match is None:
+        __log__.warn('Could not verify if %s package package is installed',
+                     kernel_package)
+        return True
+
+    kernel_ver = match.group(1).strip()
+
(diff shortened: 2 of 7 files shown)


swiftgist:
@rjfd Automation cannot rely on environment variables.  Take a look at the Salt reactor and figure out how to make the same change there.   Additionally, Stage 0 will rarely ever be run once.  Every admin on the team must now remember to set this variable; otherwise, they're stopped cold.  You may think this is not an issue, but if you apply the same strategy to the remaining hundred steps of installing Ceph, then how many variables will you have defined on the command line?  How does this solution scale to all of DeepSea.  The kernel selection is just one instance of this problem.

Part of the point in using Salt is not to use shell scripts or wrappers.  If shell scripts solves the problem, then why use Salt?

rjfd:
@swiftgist Why did you say that "automation cannot rely on environment variables"? (I'm being sincere here, I don't have experience in system administration and I want to learn what are the problems involved)

I don't know what are the capabilities of the Salt reactor, but how can I use the reactor to allow the user to specify his decision on whether to keep or change the kernel? what is your idea?

The only scenario where the user has to specify the ENV params is when executing stage 0 and there is at least one minion not using the kernel-default. Other stages will not need this parameter.
In my mind, ideally, we would decide on a default value for either change or keep the kernel, and the user could override that value in some config file.

Regarding:
> Part of the point in using Salt is not to use shell scripts or wrappers. If shell scripts solves the problem, then why use Salt?

I can imagine many different ways of building a tool to deploy a Ceph cluster. We can do it using shell scripts, Salt, Ansible, C++, nodejs. But we use Salt because we believe it's the best approach to tackle this problem.

swiftgist:
@rjfd In addition to running the Stages manually, I have been using the Salt reactor to run them automatically.  This keeps me honest.  If you have ever tried using a cron job for building software, you typically run into "environment issues".  This is similar.  However, Salt's reactor is better and worse at the same time.

Take a look at /etc/salt/master.d/reactor.conf.  Try to imagine the state diagram that you don't have with minions starting and nodes rebooting and using events to drive orchestration.  It gets complicated.  The reactor configuration file is rather simple and lacks flexibility to maintain low latency.  The effect is any complexity needs to happen in the orchestration; however, environments are lost in an asynchronous environment.

The other consideration is openATTIC.  DeepSea will effectively be an api for them.  It needs to be boring.  I understand the gripes about reading/writing the policy.cfg and not handling customizations of Ceph itself.  I do want a better solution.  However, having an admin not only remember to add a variable every time they type in the command but to inform openATTIC (and no human will ever forget) is just setting ourselves up for a disappointed, possibly angry admin.

> I don't know what are the capabilities of the Salt reactor, but how can I use the reactor to allow the user to specify his decision on whether to keep or change the kernel? what is your idea?

I don't have a new one.  The same strategy has been to use the Salt pillar.  Defining the setting there means the commands remain the same for manual, automated or openATTIC execution.  

But setting the variable is a cost that doesn't scale terribly well (in the human management way).  If in six months the defaults for DeepSea work for noone and every user has a list of things that they must specify to get started, then what are we doing?

> I can imagine many different ways of building a tool to deploy a Ceph cluster. We can do it using shell scripts, Salt, Ansible, C++, nodejs. But we use Salt because we believe it's the best approach to tackle this problem.

My point about shell scripts is that if we selected Salt, we should use it and not fallback to other comfortable solutions unless we have little choice.  I just went through this recently concerning diagnostics and a wholesale aversion to Salt and writing everything in python.

swiftgist:
@rfd Okay, I have restarted my SES3+ environment that uses kernel-default.  I went through the process and looked through the sls and kernel module.  

My bikeshedding comment is about the kernel module having so many proc calls to imitate the shell.  I think using python directly and avoiding the greps would be better.  This is purely my opinion and another opinion would be appreciated.  (I understand if this was the quick solution since it isn't the focus.)

The general steps work.  That is, if I do the following:

* Add change_kernel: YES to /srv/salt/ceph/stack/global.yml
* Add kernel_package: kernel-default-3.12.67-60.64.18.1 to /srv/salt/ceph/stack/global.yml
* Run CHANGE_KERNEL=YES salt-run state.orch ceph.stage.0

Then, my VMs upgrade from kernel-default-3.12.49-11.1 to kernel-default-3.12.67-60.64.18.1.  However, here's what's broken/unexpected.

* My reactor file for automatically running Stage 0 and 1 no longer works.
* I had to lookup and specify the available kernel
* Rerunning "CHANGE_KERNEL=YES salt-run state.orch ceph.stage.0" reboots on additional runs
* The orchestration command is now different than all other stages.

I don't know that the first one has a solution.  The third can likely be addressed.  

The crux of the issue is this though.  As an admin, why would I bother?

If I had no choice but to use DeepSea, I would likely avoid this process and simply run

salt -I cluster:ceph cmd.run "zypper --non-interactive --no-gpg-checks install --force-resolution kernel-default"
salt -I cluster:ceph cmd.run "shutdown -r now"

Once I understand how to use Salt, I can do this from memory (or possibly a quick lookup in the zypper man page which is something I already know how to do.)  The above costs more than running two Salt commands and does not benefit me as an admin.  I am doing just as many steps as I did with ceph-deploy.

Either manual solution causes a separate issue.  It breaks seamless upgrades.  One of the in progress features of DeepSea is to upgrade Ceph gracefully.  If reboots are required and require manual intervention, then this feature does not exist.

On a different note, your use of environment variables that works on minions may have solved a different problem.  It's the multiple cluster issue that I have not had time to pursue.  (I am not being facetious.)

rjfd:
> @rjfd Okay, I have restarted my SES3+ environment that uses kernel-default. I went through the process and looked through the sls and kernel module.

Hmm... this is the case where the problem does not exist. You can just run the `salt-run state.orch ceph.stage.0' (with no env variables defined) and if there is an update the kernel will be upgraded.

> My bikeshedding comment is about the kernel module having so many proc calls to imitate the shell. I think using python directly and avoiding the greps would be better.

I agree.

> The general steps work. That is, if I do the following:
> Add change_kernel: YES to /srv/salt/ceph/stack/global.yml

I don't know what is the purpose of this file, (it does not exist in DeepSea code base), why are you defining that variable?

> Add kernel_package: kernel-default-3.12.67-60.64.18.1 to /srv/salt/ceph/stack/global.yml

Same here, I don't know why you are doing this.

> Run CHANGE_KERNEL=YES salt-run state.orch ceph.stage.0

No need to run stage 0 with `CHANGE_KERNEL=YES`, the user only needs to run with `CHANGE_KERNEL=YES` when the there is a kernel installed that is not provided by the `kernel-default-base` package. In this case you only want to upgrade the current kernel which is already provided by the `kernel-default` package.

> My reactor file for automatically running Stage 0 and 1 no longer works.

I haven't prepared the code to be used with the Salt reactor

> I had to lookup and specify the available kernel

I don't understand why you had to, if you check my demo I can see that I didn't had to specify anything.

> Rerunning "CHANGE_KERNEL=YES salt-run state.orch ceph.stage.0" reboots on additional runs

This is probably because I haven't safe guard the case where only an upgrade is needed and you are telling DeepSea to change the kernel (as I said above in this scenario the user doesn't need to use any env variables when running stage 0)

PR #64, opened 2016-12-15
stage 0: fix system update and restart procedure

rjfd:
Fixes: #61

In this patch we remove the explicit `kernel-default` update procedure -- all system packages are updated by `zypper upgrade`.
Fixes the restart detection to work with any kernel package installed.

### DIFF
--- a/srv/salt/ceph/updates/default.sls
+++ b/srv/salt/ceph/updates/default.sls
@@ -6,14 +6,4 @@ zypper update:
     - shell: /bin/bash
     - unless: "zypper lu | grep -sq 'No updates found'"
 
-kernel update:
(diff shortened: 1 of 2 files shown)


swiftgist:
Let me give this a try.  I have to use my previous setup with SLE12SP1.

I'm not fond of so many shell commands including what I did originally, but this needs to be supported by the zypper salt module eventually.  If that does not happen, then I do see this becoming our own Salt module and converting most of this to python to simplify the sls file.

rjfd:
@swiftgist I agree with you. I also don't like to use that many shell commands. This should be eventually changed when we implement the support for generic distros.

Issue #70, opened 2016-12-20
Deepsea can't detect if a disk is an SSD/HDD if it sits behind a RAID Controller. 


jschmid1:
So far I heard from a couple of people wanting to deploy with a RAID setup. That won't work with the current method of detecting solid states.. 

The only tool that is capable of detecting the correct type is `smartctl megaraid`.

We either want to change the way we detect the the disktype or add some overwrite capabilities.

swiftgist:
Is this mainly for those doing pass through?  That is, the drives are not in a RAID configuration, these drives are simply connected to a RAID controller.  

If somebody is actually using RAID 3 or RAID 5, what is the "right" answer?  Suggest that the "device" be an OSD?  That seems wrong.  If it's RAID 0, then it's probably fine if not down right confusing to use both RAID and Ceph for redundancy.  Any thoughts?

jschmid1:
> Is this mainly for those doing pass through? That is, the drives are not in a RAID configuration, these drives are simply connected to a RAID controller.

I'm pretty sure that's the case here. 

> If somebody is actually using RAID 3 or RAID 5, what is the "right" answer? Suggest that the "device" be an OSD? That seems wrong.

Mh, in that case they surely don't want it to be treated as a regular OSD..

> If it's RAID 0, then it's probably fine if not down right confusing to use both RAID and Ceph for redundancy. Any thoughts?

IIRC one reason to use RAID-1 arised from a need for reliability in latency/recovery time... Never heard the usecase of RAID-0 in ceph though - 

Manual intervention might be needed here.

jschmid1:
++ If we solely rely on SMART's output we could also include checks for potential disk issues. 

++ speed ( hwinfo: real 0m2.044s vs smartctl real 0m0.445s)

smithfarm:
@jschmid1 See https://fate.suse.com/321053

jschmid1:
thanks for the pointer @smithfarm 

we might still want to consider doing a initial(before initial deployment) health check.
The regular checks should definitely go(stay) into ceph-mgr.

jschmid1:
++ adds portability as `hwinfo` is not available on other distros.

jschmid1:
fixed in #73

BlaineEXE:
[...]

PR #73, opened 2017-01-11
rewrite cephdisks / hardware detection ref #70

jschmid1:
[WIP]

### DIFF
--- a/srv/salt/_modules/cephdisks.py
+++ b/srv/salt/_modules/cephdisks.py
@@ -2,82 +2,438 @@
 
 import os
 import re
-import pprint
+import xml.etree.ElementTree as et
 from glob import glob
 from subprocess import Popen, PIPE
+import logging
+
+log = logging.getLogger(__name__)
 
 
-def list():
+class HardwareDetections(object):
     """
-    Find all unpartitioned and allocated osds.  Return hwinfo dict.
+    Custom Salt module to detect suitable disks for the ceph deployment
     """
-    drives = []
-    for path in glob('/sys/block/*/device'):
-        base = os.path.dirname(path)
-        device = os.path.basename(base)
 
-        # Skip partitioned, non-osd drives
-        partitions = glob(base + "/" + device + "*")
-        if partitions:
-            for p in partitions:
-                ids = [ re.sub('\D+', '', p) for p in partitions ]
-            if not _osd("/dev/" + device, ids):
-                continue
+    def __init__(self, **kwargs):
+        """
+        kwargs adds the possibility to pass custom raidcontroller
+        names to the class.
+        args:
+            hw_raid(bool): Manually set hw_raid True if this class can't detect it
+            hw_raid_name(str): Manually set the hw_raid_ctrls name
+            software_raid(bool): Manually set if you have sw raid and the class
+                                            fails to detect it.
 
-        # Skip removable media
+        REQUIREMENTS FOR THE PROGRAMM TO WORK:
+        gptfdisk, pciutils, smartmontools
+        """
+        self.detection_method = self.find_detection_tool(kwargs.get('detection_method', None))
+        self.hw_raid = kwargs.get('hw_raid', None)
+        self.hw_raid_name = kwargs.get('raid_controller_name', None)
+        self.software_raid = kwargs.get('sw_raid', None)
+
+    def is_removable(self, base):
+        """
+        Ask the kernel if device is a removable
+
+        args:
+            base (str): base sys path of device
+        returns:
+            bool: True if is removable
+        """
         filename = base + "/removable"
-        removable = open(filename).read().rstrip('\n')
+        with open(filename, 'r') as fd:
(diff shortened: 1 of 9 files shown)


jschmid1:
test this please

jschmid1:
retest this please

jschmid1:
retest this please

jschmid1:
retest this please

jschmid1:
test this please

jschmid1:
test this please

jschmid1:
test this please

srv/salt/_modules/cephdisks.py
```diff
@@ -2,82 +2,429 @@
...
+        kwargs adds the possibility to pass custom raidcontroller
+        names to the class.
+        args:
+            hw_raid(bool): Manually set hw_raid True if this class can't detect it
+            hw_raid_name(str): Manually set the hw_raid_ctrls name
+            software_raid(bool): Manually set if you have sw raid and the class
+                                            fails to detect it.
+        """
+        self.overwrite_detection_method = kwargs.get('detection_method', None)
+        self.detection_method = self.find_detection_tool()
```
jan--f: Do you have a reason to add the ```overwrite_detection_method``` field to the object? It is only used by ```find_detection_tool```, so passing it as a method parameter seems more suitable to me.

    self.detection_method = self.find_detection_tool(kwargs.get('detection_method', None))

srv/salt/_modules/cephdisks.py
```diff
@@ -2,82 +2,429 @@
...
+                exe_file = os.path.join(path, program)
+                if is_exe(exe_file):
+                    return exe_file
+        if nofail is True:
+            return None
+        elif nofail is False:
+            raise StandardError("Can't find the tool: {}. Please Install it in order to resume.".format(program))
+        else:
+            raise StandardError("nofail needs to be bool(True) or bool(False)")
+
```
jan--f: There's a bunch of unnecessary whitespaces here. I'm always a fan of removing these.

srv/salt/_modules/cephdisks.py
jschmid1: good catch, will update that.

srv/salt/_modules/cephdisks.py
jschmid1: done

jan--f:
@jschmid1 can you please rebase onto master. Currently your branch fails in stage 1 with an exeption: ```NameError: global name '_CephCluster__create_dirs' is not defined```

jschmid1:
done @jan--f

jan--f:
ok get to stage 3 after now. Don't get a healthy cluster though...stage 3 is stuck for the past ~2h, MONs are up, no OSDs.
```
salt-run jobs.active
20170308115416927433:
    ----------
    Arguments:
        - ceph.pool
        |_
          ----------
          __kwarg__:
              True
          concurrent:
              False
          queue:
              False
          saltenv:
              base
    Function:
        state.sls
    Returned:
    Running:
        |_
          ----------
          admin:
              16308
    StartTime:
        2017, Mar 08 11:54:16.927433
    Target:
        admin
    Target-type:
        glob
    User:
        salt
```

On a purged cluster...will destroy and recreate.

swiftgist:
I initially ran into an issue since I was overriding packages/common where a missing lspci fails quietly.  I am debating what the default behavior should be.  Part of me would like the message to be more prominent, but I know it's a separate module.  The other option could be to return what was found.  (I don't know why a Linux distribution would not have lspci (maybe repo trouble), but some list would be better than no list.)

I will say that cephdisks.list_drives bugs me. :)  I think the issue you wanted to avoid was overriding keywords.  I should have done what's listed at https://docs.saltstack.com/en/latest/ref/modules/#aliasing-functions.

jan--f:
Ok works for me. I'd still like to see most methods in cephdisks.py as private though. Unless there is a good reason not to?

jschmid1:
>  The other option could be to return what was found. (I don't know why a Linux distribution would not have lspci (maybe repo trouble), but some list would be better than no list.)

The vagrant box does not come with lspci/lsscsi preinstalled, so I had to add it to packages.common.
Usually these packages should be present.

> I will say that cephdisks.list_drives bugs me. :) I think the issue you wanted to avoid was overriding keywords. I should have done what's listed at https://docs.saltstack.com/en/latest/ref/modules/#aliasing-functions.

Correct, I wanted to avoid overwriting _list_ 

-> looks good!
```
__func_alias__ = {
    'set_': 'set',
    'list_': 'list',
}
```

I'll adapt to that in the next commit.

jschmid1:
> I initially ran into an issue since I was overriding packages/common where a missing lspci fails quietly.

Oh, true. Haven't thought about the fact that common.packages is pointing to default-zypper.sls rather than default.sls hence made my changes in the latter..

Will fix that.

We still should track the behavior of pkg.install. It should support zyppers  "--no-gpg-checks" eventually to make our lives easier.

PR #78, opened 2017-01-21
Add disengage.safety and ceph.purge

swiftgist:
Add a runner to allow the resetting of a cluster and DeepSea configuration.  Require another runner to prevent accidental purging.


swiftgist:
Works for Jan and José

Issue #83, opened 2017-01-26
add ability to add warnings to validate


theanalyst:
Currently we have the ability to only pass fail in validate, sometimes (for eg. in case of fqdn) we'd like to have an option to warn and pass

theanalyst:
Working on this

PR #84, opened 2017-01-26
validate: add ability to warn & warn when fqdn is localhost

theanalyst:
In certain cases it makes sense to warn the user instead of erroring
out, adding the ability for this.

Fixes: #83

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -488,14 +488,16 @@ class Validate(object):
         """
         Verify that fqdn matches minion id
         """
-        for node in self.grains.keys():
-            if self.grains[node]['fqdn'] != node:
-                msg = "fqdn {} does not match minion id {}".format(self.grains[node]['fqdn'], node)
-                if 'fqdn' in self.errors:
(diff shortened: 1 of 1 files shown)


theanalyst:
WIP: still pending testing

theanalyst:
Current output:

```sh
salt-run validate.pillars 
fsid                     : valid
public_network           : valid
public_interface         : valid
cluster_network          : valid
cluster_interface        : valid
monitors                 : valid
storage                  : valid
master_role              : valid
mon_host                 : valid
mon_initial_members      : valid
time_server              : valid
fqdn                     : ['fqdn localhost does not match minion id tf-salt-master', 'fqdn localhost does not match minion id tf-salt-minion-1', 'fqdn localhost does not match minion id tf-salt-minion-0', 'fqdn localhost does not match minion id tf-salt-minion-3', 'fqdn localhost does not match minion id tf-salt-minion-2']
```

theanalyst:
An easy way to test is probably doing `salt "*" grains.set fqdn localhost` and then running validate

PR #89, opened 2017-01-27
tests: add Travis CI integration

theanalyst:
Runs just tox at the moment, also setting pip cache so that we don't
download packages between builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: python
+python:
+  - "2.7"
+install: pip install tox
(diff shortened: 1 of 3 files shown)


theanalyst:
Requires someone with admin access to the repo to switch travis on from travis-ci.org after merging

theanalyst:
Needs atleast one test case as py.test seems to be unhappy with no tests

jschmid1:
looks good to me!

PR #93, opened 2017-02-01
rpm: add python-ipaddress to the rpm requires

theanalyst:
We use ipaddress python module in validate, while it might be installed
in many systems there is no explicit requires and it would lead to
interesting failures if not installed


theanalyst:
PS I really don't have a great experience modifying an rpm spec file, asking @smithfarm @swiftgist for review

PR #96, opened 2017-02-03
stage 1,3: support for multiple public/cluster networks

rjfd:
The purpose of this patch is to generalize the public/cluster network Ceph config option to support more than one network, as discussed in http://lists.suse.com/pipermail/deepsea-users/2016-December/000022.html

The algorithm for assigning the public/cluster networks work as follows:

1. assign the network with more hosts as public network
2. assign the second network with more hosts (but with at least 2 hosts) as cluster network
3. any network with a single host is assigned as a public network
4. in the end, if cluster network does not have any assignment than it inherits the same networks as the public network

### DIFF
--- a/srv/modules/runners/populate.py
+++ b/srv/modules/runners/populate.py
@@ -463,7 +463,7 @@ class CephRoles(object):
 
         self.root_dir = settings.root_dir
         self.networks = self._networks(self.servers)
-        self.public_network, self.cluster_network = self.public_cluster(self.networks)
+        self.public_networks, self.cluster_networks = self.public_cluster(self.networks)
 
         #self.master_contents = {}
         self.available_roles = []
@@ -622,11 +622,12 @@ class CephRoles(object):
         """
         Find the public interface for a server
         """
-        public_net = ipaddress.ip_network(u'{}'.format(self.public_network))
-        for entry in self.networks[public_net]:
-            if entry[0] == server:
-                log.debug("Public interface for {}: {}".format(server, entry[2]))
-                return entry[2]
+        for public_network in self.public_networks:
+            public_net = ipaddress.ip_network(u'{}'.format(public_network))
+            for entry in self.networks[public_net]:
+                if entry[0] == server:
+                    log.debug("Public interface for {}: {}".format(server, entry[2]))
+                    return entry[2]
         return ""
 
 
@@ -641,8 +642,12 @@ class CephRoles(object):
             filename = "{}/cluster.yml".format(cluster_dir)
             contents = {}
             contents['fsid'] = str(uuid.uuid3(uuid.NAMESPACE_DNS, os.urandom(32)))
-            contents['public_network'] = self.public_network
-            contents['cluster_network'] = self.cluster_network
+
+            public_networks_str = ", ".join([str(n) for n in self.public_networks])
+            cluster_networks_str = ", ".join([str(n) for n in self.cluster_networks])
+
+            contents['public_network'] = public_networks_str
+            contents['cluster_network'] = cluster_networks_str
             contents['available_roles'] = self.available_roles
 
             self.writer.write(filename, contents)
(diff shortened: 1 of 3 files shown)


srv/modules/runners/validate.py
```diff
@@ -107,6 +107,20 @@ def _clusters(self):
             clusters.setdefault(cluster, []).append(minion)
         return clusters
 
+
+class Util(object):
+    @staticmethod
+    def parse_list_from_string(list_str, delim=','):
```
theanalyst: why not a list comprehension or something little more simpler there,,
```py
[x for x in xs.split(",") if x != '']  
``` 
for eg

theanalyst: Also a test case would be nice to have

srv/modules/runners/populate.py
```diff
@@ -671,21 +681,25 @@ def public_cluster(self, networks):
...
 
         priorities = sorted(priorities, cmp=network_sort)
-        if len(priorities) == 1:
-            return str(priorities[0][1]), str(priorities[0][1])
-        else:
-            return str(priorities[0][1]), str(priorities[1][1])
+        public_networks = list()
+        cluster_networks = list()
+
+        for i in range(0, len(priorities)):
```
theanalyst: a more pythonic way would be to use enumerate here

srv/modules/runners/populate.py
```diff
@@ -619,8 +620,17 @@ def cluster_config(self):
             filename = "{}/cluster.yml".format(cluster_dir)
             contents = {}
             contents['fsid'] = str(uuid.uuid3(uuid.NAMESPACE_DNS, os.urandom(32)))
-            contents['public_network'] = self.public_network
-            contents['cluster_network'] = self.cluster_network
+
+	    public_networks_str = str(self.public_networks[0])
+            for i in range(1, len(self.public_networks)):
```
theanalyst: maybe `",".join(public_networks)` ?

srv/modules/runners/populate.py
jan--f: Yeah I agree.

    for priority in priorities:

would me much nicer here.

swiftgist:
We need to consider upgrading from a previous version where public_network and cluster_network are single valued.

rjfd:
@theanalyst @jan--f addressed your comments, added a couple of tests, and rebased on top of master.

rjfd:
@swiftgist in the validate runner the parsing of the public/cluster networks field is generic enough to parse single valued values.
So, I think there should not be any consequences for the upgrade

rjfd:
@theanalyst do you still have any concern with this PR or is it good to go?

swiftgist:
@rjfd I wanted to give this a spin but have not had a chance yet.

swiftgist:
@rjfd Okay, it seems to work fine, but I have a question. :)  What are the side effects of

public_network = 172.16.21.0/24, 192.168.21.0/24, 192.168.121.0/24

in my ceph.conf.  I had been under the impression that monitors and OSDs would bind to all addresses but that does not seem to be the case.  Would I be confused if I saw this as a first time Ceph admin wondering why I have so many public networks?

rjfd:
@swiftgist good question. I don't know if mons will bind to all addresses that belong to one of the public network list. Maybe @jecluis can teach us how that process works in this case.

Regarding the length of the public_network list, since we are trying to automatically figure out what networks should be considered public, I don't think there's much we can do right now. What I think it could help was to develop some easy way that the sysadmin could tell us which networks "shouldn't" be considered for public or cluster networks. Some kind of "exclusive-set".

swiftgist:
@rjfd What do you think of this?

        for idx, (quantity, network) in enumerate(priorities):
            if idx == 1 and quantity > 1:
                cluster_networks.append(network)
            if idx == 0 and quantity > 1:
                public_networks.append(network)
            if quantity == 1:
                public_networks.append(network)

Would that still match the intent for OpenStack and OVH systems?  This gives me a single public network.

jecluis:
> I don't know if mons will bind to all addresses that belong to one of the public network list

The monitors will bind to the first local address (as returned by `getifaddrs(3)`) that matches a network in `public_network`.

In essence, if you have multiple entries in `public_network`, all of them with local addresses, we will pick the first `public_network` on the list and bind to the corresponding local address. Those networks without a corresponding local address will be ignored.

This may get a little trickier as you start getting `mon_host` in the mix though, but I'll spare you the gory details until you actually need them.

rjfd:
@swiftgist that algorithm will work for OVH environments, but I think there is no good excuse to exclude other networks that have more than one node.

jan--f:
My test run fails here: https://github.com/rjfd/DeepSea/blob/48896fcf03310522b24286b579408d22f04775ab/srv/modules/runners/populate.py#L563
with ```NameError: global name 'add_pub_interface' is not defined```. I guess ```self.add_pub_interface``` should be called instead?

rjfd:
@jan--f that problem is unrelated with this PR and it was already fixed in the master branch. I'll rebase this PR on top of master so you can test it again.

rjfd:
@jan--f rebase done

swiftgist:
@rjfd I can think of a few reasons.  

* Expectations is one.  Aside from the OpenStack environments, I believe a single public network is the norm for both POCs and production environments.  

* Another is consistency with how the ceph.conf has appeared to all of those currently using SES.  

* Some failures will go unnoticed and result in having to setup part of the Ceph cluster again.  Think intended public network interface is misbehaving, but DeepSea magically configures the storage/monitor node to use another interface.  The cluster may work, but it's not what the admin wanted.  I expect the validation to okay the setup so this will be discovered the hard way.  

* Many gateways will have additional interfaces intended as dedicated networks between specific clients.  Including the aggregate of all subnets on 50+ nodes is not helping anyone.  Requiring a blacklist is giving the admins more to do that is not necessary.

As far as those that have two separate subnets for a public network, I would expect this to happen over time and would be uncommon.  This group can override the Salt pillar configuration which is how everything else is configured in Salt.

rjfd:
@swiftgist alright, those reasons make sense to me. I'll update the PR accordingly

rjfd:
@swiftgist pushed the new changes.

PR #100, opened 2017-02-09
Disable unsupported kernel features for rbd images

swiftgist:
With the SES5 repo and kernel 4.4.38-93-default, iSCSI gateways fail to configure with the demo image.  Set rbd images to currently supported features.


srv/salt/ceph/configuration/files/ceph.conf-shared.j2
```diff
@@ -8,6 +8,7 @@ auth_client_required = cephx
 filestore_xattr_use_omap = true
 public_network = {{ salt['pillar.get']('public_network') }}
 cluster_network = {{ salt['pillar.get']('cluster_network') }}
+rbd default features = 3
```
rjfd: In SES5 we already support textual values for the RBD features, like this:
`rbd default features = layering,striping`

But if this configuration needs to work with versions <SES5 then it looks good to me

jan--f:
Yep looks good. This should find its way into the SES docs once its released.

PR #101, opened 2017-02-09
rpm: add hwinfo to requires

theanalyst:
We use hwinfo in stage1, in cephdisks module. Adding an explicit
requirement so that we install hwinfo with deepsea itself

### DIFF
--- a/deepsea.spec
+++ b/deepsea.spec
@@ -32,6 +32,7 @@ BuildRequires:  salt-master
 Requires:       salt-master
 Requires:       salt-minion
 Requires:       python-ipaddress
+Requires:       hwinfo
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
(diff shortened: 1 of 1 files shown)


swiftgist:
Adding this to the spec file won't help where you need it.  The hwinfo is needed on all SUSE minions.  (We are assuming lshw and udevadm for non-SUSE minions. )

I guess the solution if this bit you is to add it to ceph.packages.common with a SUSE conditional so that it does get installed during Stage 0.  Honestly though, I'm not crazy about adding arch conditionals in the sls files.  

I can't find the apache example for pkg.installed at the moment.  Not only does pkg.installed need to call the "right" tool, but the package name itself needs to be correct.  What do you think of that direction?

jan--f:
@abhi can we close this? Seems like your issue needs a different solution.

Issue #81, opened 2017-01-26
stage 2 fails silently if /srv/pillar/ceph/stack/ceph is not writeable


jan--f:
Had the wrong owner/permission on /srv/pillar/ceph/stack/ceph. The push runner fails fairly silently if it can not write to this directory.
With `-l debug` one gets
```
[INFO] Exception occurred in runner push.proposal: Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/client/mixins.py", line 357, in low
    data['return'] = self.functions[fun](*args, **kwargs)
  File "/srv/modules/runners/push.py", line 72, in proposal
    pillar_data.output(common)
  File "/srv/modules/runners/push.py", line 120, in output
    self._custom(custom)
  File "/srv/modules/runners/push.py", line 152, in _custom
    os.makedirs(path_dir)
  File "/usr/lib64/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 13] Permission denied: '/srv/pillar/ceph/stack/ceph/minions'
```

Maybe the push runner can fail a bit more dramatically. Runing stage 3 after doesn't give much info either.

PR #105, opened 2017-02-10
fail hard if stage 2 fails to create directory

jan--f:
Fixes: #81

### DIFF
--- a/srv/modules/runners/push.py
+++ b/srv/modules/runners/push.py
@@ -72,6 +72,16 @@ def proposal(filename = "/srv/pillar/ceph/proposals/policy.cfg", dryrun = False)
     pillar_data.output(common)
     return True
 
+def __create_dirs(path, root):
+    try:
+        os.makedirs(path)
+    except OSError as err:
+        if err.errno == errno.EACCES:
+            log.exception('''
+            ERROR: Cannot create dir {}
+            Please make sure {} is owned by salt
+            '''.format(path, root))
+            raise err
 
 class PillarData(object):
     """
@@ -135,7 +145,7 @@ class PillarData(object):
         """
         path_dir = os.path.dirname(filename)
         if not os.path.isdir(path_dir):
-            os.makedirs(path_dir)
+            __create_dirs(path_dir, self.pillar_dir)
         log.info("Writing {}".format(filename))
         if not self.dryrun:
             with open(filename, "w") as yml:
@@ -149,7 +159,7 @@ class PillarData(object):
         """
         path_dir = os.path.dirname(custom)
         if not os.path.isdir(path_dir):
-            os.makedirs(path_dir)
+            __create_dirs(path_dir, self.pillar_dir)
         if not self.dryrun:
             if not os.path.isfile(custom):
                 log.info("Writing {}".format(custom))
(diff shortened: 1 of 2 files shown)


srv/modules/runners/push.py
```diff
@@ -72,6 +72,14 @@ def proposal(filename = "/srv/pillar/ceph/proposals/policy.cfg", dryrun = False)
     pillar_data.output(common)
     return True
 
+def __create_dirs(path, root):
+    try:
+        os.makedirs(path)
+    except OSError as err:
+        if err.errno == errno.EACCES:
+            print "ERROR: Cannot create dir {}".format(path)
```
theanalyst: does the print statement also get displayed when raising the exception ? (or do we also need a log.error? )

jan--f: It does get printed. But I agree...I should call log here.

rjfd:
@jan--f please squash the first and third commits.

srv/modules/runners/push.py
```diff
@@ -72,6 +72,16 @@ def proposal(filename = "/srv/pillar/ceph/proposals/policy.cfg", dryrun = False)
     pillar_data.output(common)
     return True
 
+def __create_dirs(path, root):
```
jschmid1: since we have the exact same method here and in `populate.py:757` we might consider to create a helper lib.

jan--f: I'd be all for that. Just haven't found an answer where such a lib would live. '''/srv/modules/runners''' sounds like a bad idea to me, since salt would pick this up as a runner.

jan--f: @swiftgist Any idea what the convention would be, if there is any?

swiftgist: I haven't found it yet.  I suspect we could have something live under /usr/lib/python2.7/site-packages/salt, but that does not help with packaging.  I suppose the closest I have come with a common function for runners is a module.  This might seem strange or abusive, but we have more than this create_dirs example that is common across runners.

swiftgist:
Double check the behavior of the orchestration file for Stage 2.  I had to rely on the side effect of printing in other cases from runners since the orchestration will not capture the error effectively and only gives a stack trace.  I have also been avoiding stack traces since finding the errors are difficult.

jan--f:
hmm it does print ok-ish. The runner still doesn't fail as I'd like. What do you think @swiftgist ?
```
[ERROR   ] 
            ERROR: Cannot create dir /srv/pillar/ceph/proposals/cluster-ceph/cluster
            Please make sure /srv/pillar/ceph/proposals is owned by salt
            
  File "/srv/modules/runners/populate.py", line 759, in _create_dirs
    os.makedirs(path)
  File "/usr/lib64/python2.7/os.py", line 150, in makedirs
    makedirs(head, mode)
  File "/usr/lib64/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 13] Permission denied: '/srv/pillar/ceph/proposals/cluster-ceph'
retcode:
    0
admin_master:
  Name: minions.ready - Function: salt.runner - Result: Changed Started: - 15:01:49.692495 Duration: 538.227 ms
  Name: populate.proposals - Function: salt.runner - Result: Changed Started: - 15:01:50.230955 Duration: 804.119 ms

Summary for admin_master
------------
Succeeded: 2 (changed=2)
Failed:    0
------------
Total states run:     2
Total run time:   1.342 s
```

swiftgist:
That seems fine.  The error message is at the top and the stack trace isn't too long to cause things to scroll off.  Maybe log.exception appears more prominent.  If that's the case, we should make the same change in other runners.

jan--f:
@jschmid1 should we defer the code duplication until we have a good solution for this?

PR #106, opened 2017-02-10
populate: use self while calling add_pub_interface

theanalyst:
Since it is a method in the class we need to call it with self.
otherwise we'll end up with a NameError

### DIFF
--- a/srv/modules/runners/populate.py
+++ b/srv/modules/runners/populate.py
@@ -560,7 +560,7 @@ class CephRoles(object):
         Create a file for mon_host and mon_initial_members
         """
         minion_dir = "{}/role-mon/stack/default/{}/minions".format(self.root_dir, self.cluster)
-        add_pub_interface(minion_dir)
+        self._add_pub_interface(minion_dir)
 
     def igw_members(self):
         """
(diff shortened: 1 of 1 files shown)


theanalyst:
Missed this in #95 review, we really need a linter to check these :)

srv/modules/runners/populate.py
```diff
@@ -569,14 +569,14 @@ def igw_members(self):
...
 
     def ganesha_members(self):
         """
         Create a file for ganesha hosts.
         """
         minion_dir = "{}/role-ganesha/stack/default/{}/minions".format(self.root_dir, self.cluster)
-        add_pub_interface(minion_dir)
+        self.add_pub_interface(minion_dir)
 
     def add_pub_interface(self, minion_dir):
```
jan--f: maybe make ```add_pub_interface``` "private"?

theanalyst: ie. `_add_pub_interface`?

theanalyst: @jan--f updated

rjfd:
@theanalyst I agree we should have pylint running on travis, but before we activate such thing we have to cleanup all the python code according to the linter rules.
Can you open an issue to track the use of pylint?

theanalyst:
@rjfd opened https://github.com/SUSE/DeepSea/issues/107

srv/modules/runners/populate.py
jan--f: Is it is only used within the class even ```__add_pub_interface```. It won't be truly private but at least the intent is clear.

PR #109, opened 2017-02-10
populate: dont use double underscored functions from outside

theanalyst:
Since python mangles anything inside a class with a double underscore
assuming private methods these would again raise NameErrors.

As an example:

```py
In [1]: def __foo():
   ...:     return 1
   ...:

In [2]: class A():
   ...:     def __init__(self):
   ...:         self.a = 1
   ...:     def bar(self):
   ...:         return __foo() + self.a
   ...:

In [3]: a = A()

In [4]: a.bar()
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-4-4b57aad064bc> in <module>()
----> 1 a.bar()

<ipython-input-2-4bba0c494767> in bar(self)
      3         self.a = 1
      4     def bar(self):
----> 5         return __foo() + self.a
      6

NameError: global name '_A__foo' is not defined
```

### DIFF
--- a/srv/modules/runners/populate.py
+++ b/srv/modules/runners/populate.py
@@ -131,7 +131,7 @@ class CephStorage(object):
         """
         model_dir = "{}/{}/stack/default/{}/minions".format(self.root_dir, name, self.cluster)
         if not os.path.isdir(model_dir):
-            __create_dirs(model_dir, self.root_dir)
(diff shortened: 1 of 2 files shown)


jan--f:
dang...yeah absolutely correct. Forgot to push that last part.

PR #112, opened 2017-02-12
kernel.replace module

swiftgist:
Remove the hardcoded zypper update and allow the changing of the kernel from a list of candidates per OS.


theanalyst:
What about cases when we need no kernel updates? ie. for instance an rgw only cluster has no dependencies on updating the kernel, how does this syntax support such a case?

swiftgist:
@theanalyst If somebody really did want certain roles to have one kernel and not another, the admin can make another sls file and set updates_init only to that role.  For instance, this would be the same behavior if an admin wishes to disable this completely because SUMA is used for kernel management.

swiftgist:
@jan--f Double checked.  Makefile and spec file are using *.sls, so no need to explicitly remove

theanalyst:
I'm ok with the idea in general, just that an easy way to disable kernel updates via a `update_kernel` flag set to yes by default could be incorporated in future maybe?

jan--f:
@theanalyst I had the same thought. Unfortunately it ends up being no less complex then using a custom method for updates. Orchestration files like stage 1 don't accept parameters, so such an option had to go into the pillar. So we could just provide a no-kernel-update.sls or so and have people set ```updates_init``` accordingly. Same effect, no extra config option (aka complexity).

PR #121, opened 2017-02-27
cherrypy: add state files to Makefile and RPM spec files

rjfd:
Adding state files to the Makefile (and RPM spec file), to allow the installation and usage of the salt-api.


swiftgist:
@rjfd Thanks for this.  I noticed it after the announcement.  Would you add .../cherrypy/files/*.j2 as well?

rjfd:
@swiftgist I'm already including those by using the glob `*.conf*`, but if you think it's better to explicitly add a new entry with `*.j2`, I'll add it.

PR #126, opened 2017-03-03
Makefile: add hint to restart salt-master

jan--f:
Ran into this more then once. Its not obvious after installing via make install, since it half works but salt-master doesn't pick up the stack.cfg. This becomes a problem after stage 2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -333,12 +333,15 @@ install:
 	ln -sf services		$(DESTDIR)/srv/salt/ceph/stage/4
 	ln -sf removal		$(DESTDIR)/srv/salt/ceph/stage/5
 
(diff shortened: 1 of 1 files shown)


jan--f:
@swiftgist I was also wondering: adding a ```chown -R salt /srv/pillar/ceph``` is probably at odds with package build isn't it?

swiftgist:
@jan--f Were you think of just adding it to the install target in the Makefile?  I think it may be okay to restart the Salt master there as well since anyone doing this type of install would have to restart it.  What do you think?

jan--f:
@swiftgist I'm worried what OBS will do with that. or can the install command enforce file ownership too?

jan--f:
closing...superseded by #131

Issue #128, opened 2017-03-07
validation fails when mds, rgw and ganesha roles are not present


jan--f:
The ganesha validation checks for mds or rgw (and fails if missing) depsite nor ganesha role being deployed. Validation should check for missing mds or rgw only when ganesha is actually deployed.

PR #129, opened 2017-03-07
validate: ganesha validation is too strict

jan--f:
Ganesha validation should only fail on missing mds or rgw node if
ganesha is actually deployed.
Also adds some refactoring to simplify pillar iteration.
Fixes: #128 

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -278,37 +278,43 @@ class Validate(object):
 
     def ganesha(self):
         """
-        Nodes may only be assigned one ganesha role.  Ganesha depends on 
+        Nodes may only be assigned one ganesha role.  Ganesha depends on
         cephfs or radosgw.
         """
         ganesha_roles = []
         role_mds = False
         role_rgw = False
+        role_ganesha = False
 
-        for node in self.data.keys():
-            if ('roles' in self.data[node]):
-                if('ganesha_configurations' in self.data[node]):
-                    ganesha_roles = list(set(self.data[node].get("roles")) &
-                                        set(self.data[node].get("ganesha_configurations")))
+        for node, data in self.data.items():
+            if ('roles' in data):
+                if('ganesha_configurations' in data):
+                    ganesha_roles = list(set(data.get("roles")) &
+                                        set(data.get("ganesha_configurations")))
                     if len(ganesha_roles) > 1:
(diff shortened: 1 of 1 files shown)


srv/modules/runners/validate.py
```diff
@@ -278,37 +278,41 @@ def storage(self):
...
-                    ganesha_roles = list(set(self.data[node].get("roles")) &
-                                        set(self.data[node].get("ganesha_configurations")))
+        for node, data in self.data.items():
+            if ('roles' in data):
+                if('ganesha_configurations' in data):
+                    ganesha_roles = list(set(data.get("roles")) &
+                                        set(data.get("ganesha_configurations")))
                     if len(ganesha_roles) > 1:
                         msg = "minion {} has {} roles. Only one permitted".format(node, ganesha_roles)
                         self.errors.setdefault('ganesha',[]).append(msg)
```
supriti: We can set role_ganesha = True if len(ganesa_roles) is not greater than 1.

srv/modules/runners/validate.py
```diff
@@ -278,37 +278,41 @@ def storage(self):
...
+                                set(data.get("rgw_configurations")))):
                             role_rgw=True
 
-        if not (role_mds or role_rgw):
+                if not role_ganesha:
+                    role_ganesha = 'ganesha' in data['roles']
+
+        if not (role_mds or role_rgw) and role_ganesha:
             msg = "Ganesha requires either mds or rgw node in cluster."
             self.errors['ganesha'] = msg
```
supriti: We can set role_ganesha True in the previous loop. 
Maybe rewrite as. 
if role_ganesha:
  if(role_rgw or role_mds):
    self.passed['ganesha'] = "valid"
  else:
     msg = "Ganesha requires either mds or rgw node in cluster."
     self.errors['ganesha'] = msg

supriti: To me, the code flow seems more intuitive this way. But correctness wise, both approach are ok.

srv/modules/runners/validate.py
jan--f: Yeah I din't quite understand that part. Why does this code binary-and the two sets?

supriti: User can define custom roles in ganesha_configurations. We are taking intersection of ganesha_configurations and defined roles for node.

PR #131, opened 2017-03-08
Makefile/spec: complete installation with make and rpm

jan--f:
Until now the Makefile only copied files to the correct location.
Further steps were necessary to get a functional installation. This
commit adds the necessary step while keeping the build process working.
Also a chown was added to the rpm's post section to complete
installation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ usage:
 	@echo -e "\tmake install\tInstall DeepSea on this host"
 	@echo -e "\tmake rpm\tBuild an RPM for installation elsewhere"
 
-install:
+copy-files:
 	# salt-master config files
 	install -d -m 755 $(DESTDIR)/etc/salt/master.d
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -338,6 +338,9 @@ install:
 	ln -sf services		$(DESTDIR)/srv/salt/ceph/stage/4
 	ln -sf removal		$(DESTDIR)/srv/salt/ceph/stage/5
 
+install: copy-files
+	chown -R salt /srv/pillar/ceph
```
rjfd: @jan--f don't you also want to add this line:

```
sed -i "s/_REPLACE_ME_/`hostname -f`/" /srv/pillar/ceph/master_minion.sls
```

to before the chown command?

jan--f:
test build can be found here https://build.opensuse.org/package/show/home:jfajerski/deepsea

Makefile
jan--f: I absolutely did forget that...fixed.

PR #135, opened 2017-03-09
Enable openattic services

swiftgist:

### DIFF
--- a/srv/salt/ceph/openattic/default.sls
+++ b/srv/salt/ceph/openattic/default.sls
@@ -5,3 +5,13 @@ install openattic:
     - pkgs:
       - openattic
 
+enable openattic-rpcd:
+  service.running:
+    - name: openattic-rpcd
+    - enable: True
+
+enable openattic-systemd:
+  service.running:
+    - name: openattic-systemd
+    - enable: True
+


srv/salt/ceph/openattic/default.sls
```diff
@@ -5,3 +5,13 @@ install openattic:
     - pkgs:
       - openattic
 
+enable openattic-rpcd:
```
rjfd: openattic-rpcd service was already removed from the development branch in openATTIC, so it will not exist in the next openATTIC version.

Also, doesn't the openATTIC RPM installation automatically starts these services?

jan--f: I had the same thought. Salt just makes sure that the service is started...so no harm done I figured.

LenzGr: The automatic start/enablement was removed from the SES RPM spec file, as this is against the packaging guidelines. We're looking into adding this to `oaconfig` instead, e.g. creating `oaconfig enable` that will make sure that all required services (Apache, Nagios, PostgreSQL) are enabled upon reboot.

LenzGr: FYI, openattic-rpcd will be removed in 2.0.19, so this likely means we have to make sure that DeepSea is aware of these changes as well.

rjfd:
@swiftgist With the changes of this PR only, openATTIC is still not installed in any stage, right?

swiftgist:
@rjfd On the master, openATTIC is not currently installed.  On SES4, openATTIC is installed during Stage 4.  

Although openattic-rpcd will be removed in 2.0.19, it still exists in the currently distributed openATTIC.   I plan to cherrypick this commit to create an MR for SES4.  When 2.0.19 is released, the rpcd could be removed or changed to a conditional.  

After some discussions with Lenz, openATTIC should be added back to Stage 4 but also have a dedicated role.  I think that's worth a separate PR.

srv/salt/ceph/openattic/default.sls
LenzGr: FYI, 2.0.19 has been published - please remove any references to `openattic-rpcd` in future DeepSea releases.

LenzGr: See #198 for a PR that performs this removal.

PR #136, opened 2017-03-09
cephservices

swiftgist:
Correction for delay, add _timeout and unit tests

### DIFF
--- a/srv/modules/runners/cephservices.py
+++ b/srv/modules/runners/cephservices.py
@@ -120,7 +120,7 @@ def wait(**kwargs):
         if check(**kwargs):
             log.debug("Services are up")
             return True
-        time.sleep(settings['delay'])
+        time.sleep(current_delay)
         if current_delay < 60:
             current_delay += settings['delay']
         else:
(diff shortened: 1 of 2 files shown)


srv/modules/runners/cephservices.py
```diff
@@ -108,22 +109,34 @@ def wait(**kwargs):
...
+        time.sleep(current_delay)
         if current_delay < 60:
-            current_delay += delay
+            current_delay += settings['delay']
         else:
             current_delay = 60
     log.error("Timeout expired")
     raise RuntimeError("Timeout expired")
 
+def _timeout(cluster='ceph', **kwargs):
```
jan--f: What is the motivation for passing ```**kwargs``` here? Its never accessed. I can see why the ```check()``` gets it (though is doesn't use it either).

jschmid1:
not part of this PR but:

https://github.com/SUSE/DeepSea/blob/8615903a61de3ddd3105adeb5d6e23c3ad63c729/srv/modules/runners/cephservices.py#L92

'expected' is assigned to but never used

PR #137, opened 2017-03-09
filequeue

swiftgist:
Change parameter, add unit tests

### DIFF
--- a/tests/unit/runners/test_filequeue.py
+++ b/tests/unit/runners/test_filequeue.py
@@ -66,7 +66,7 @@ class TestFileQueue():
         touched = fq.touch("blue")
         files = fq.ls()
         shutil.rmtree(dirpath)
-        assert set(files) == set(['red', 'blue'])
+        assert sorted(files) == (['blue', 'red'])
 
     @patch('srv.modules.runners.filequeue.FileQueue._fire_event', autospec=True)
     def test_items(self, fire_event, dirpath):
--- a/tests/unit/runners/test_filequeue.py
+++ b/tests/unit/runners/test_filequeue.py
@@ -1,0 +1,58 @@
+
+# Import Salt Testing libs
+from salttesting import skipIf, TestCase
+from salttesting.helpers import ensure_in_syspath
+
+import tempfile
+import shutil
+import os
+import pytest
+from salttesting.mock import NO_MOCK, NO_MOCK_REASON, MagicMock, patch, call
+from srv.modules.runners import filequeue
+
+class FilequeueTestCase(TestCase):
+    '''
+    This class contains a set of functions that test salt.modules.fib.
+    '''
+    def test_default_queue(self):
+        '''
+        Verify that the default queue exists
+        '''
+        dirpath = tempfile.mkdtemp()
+        fq = filequeue.FileQueue(root_dir=dirpath)
+        queue = os.path.isdir("{}/default".format(dirpath))
+        shutil.rmtree(dirpath)
+        self.assertEqual(queue, True)
+
+    def test_named_queue(self):
+        '''
+        Verify that the 'lunch' queue exists
+        '''
+        dirpath = tempfile.mkdtemp()
+        fq = filequeue.FileQueue(root_dir=dirpath, queue='lunch')
+        queue = os.path.isdir("{}/lunch".format(dirpath))
+        shutil.rmtree(dirpath)
+        self.assertEqual(queue, True)
+
+    def test_dirs(self):
+        '''
+        Verify that dirs returns the queues
+        '''
+        dirpath = tempfile.mkdtemp()
+        fq = filequeue.FileQueue(root_dir=dirpath, queue='lunch')
+        names = fq.dirs()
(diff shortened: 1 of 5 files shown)


tests/unit/runners/test_filequeue.py
```diff
@@ -0,0 +1,172 @@
...
+        '''
+        Verify that ls returns filenames
+        '''
+        dirpath = tempfile.mkdtemp()
+        fq = filequeue.FileQueue(root_dir=dirpath)
+        touched = fq.touch("red")
+        touched = fq.touch("blue")
+        files = fq.ls()
+        shutil.rmtree(dirpath)
+        assert files == ['red', 'blue']
```
jan--f: Getting a test failure here.
```
assert ['blue', 'red'] == ['red', 'blue']
```

So its just an ordering issue. Either sort both lists before the comparison or compare sets.

tests/unit/runners/test_filequeue.py
```diff
@@ -0,0 +1,172 @@
...
+    '''
+    def test_default_queue(self):
+        '''
+        Verify that the default queue exists
+        '''
+        dirpath = tempfile.mkdtemp()
+        fq = filequeue.FileQueue(root_dir=dirpath)
+        queue = os.path.isdir("{}/default".format(dirpath))
+        shutil.rmtree(dirpath)
+        assert queue == True
```
jschmid1: comparison to True should be done with 'is'

tests/unit/runners/test_filequeue.py
```diff
@@ -0,0 +1,172 @@
...
+class TestFileQueue():
+    '''
+    This class tests the FileQueue class for the salt runner
+    '''
+    def test_default_queue(self):
+        '''
+        Verify that the default queue exists
+        '''
+        dirpath = tempfile.mkdtemp()
+        fq = filequeue.FileQueue(root_dir=dirpath)
```
jschmid1: fq is assigned but never used

tests/unit/runners/test_filequeue.py
```diff
@@ -0,0 +1,172 @@
...
+        assert touched and exists 
+
+    @patch('srv.modules.runners.filequeue.FileQueue._fire_event', autospec=True)
+    def test_ls(self, fire_event):
+        '''
+        Verify that ls returns filenames
+        '''
+        dirpath = tempfile.mkdtemp()
+        fq = filequeue.FileQueue(root_dir=dirpath)
+        touched = fq.touch("red")
```
jschmid1: and no need to assign 'touched'

tests/unit/runners/test_filequeue.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def test_named_queue(self):
+        '''
+        Verify that the 'lunch' queue exists
+        '''
+        dirpath = tempfile.mkdtemp()
+        fq = filequeue.FileQueue(root_dir=dirpath, queue='lunch')
+        queue = os.path.isdir("{}/lunch".format(dirpath))
+        shutil.rmtree(dirpath)
+        assert queue == True
```
jschmid1: same as: https://github.com/SUSE/DeepSea/pull/137/files#r105359743

tests/unit/runners/test_filequeue.py
```diff
@@ -0,0 +1,172 @@
...
+
+class TestFileQueue():
+    '''
+    This class tests the FileQueue class for the salt runner
+    '''
+    def test_default_queue(self):
+        '''
+        Verify that the default queue exists
+        '''
+        dirpath = tempfile.mkdtemp()
```
jschmid1: instead of wirting `dirpath = tempfile.mkdtemp()` in every method, pytest suggests to use _fixtures_

http://doc.pytest.org/en/latest/fixture.html

jschmid1: same can be done for `filequeue.FileQueue(root_dir=dirpath)`

swiftgist:
Fixed the ordering issue and added a fixture.  

It seems like a toss up to me with respect to a fixture in this case.  While it saves a few lines, I am adding a new parameter to each function.  I think the amount of time to figure out either is about the same.  If the fixture has a few steps, I can see the advantage.

If this were more complicated, I could setup a context or wrapper class that would preserve dirpath between the setup and teardown.  It seems though pytest is considering that method to be the "classical" way.  However, I do not think adding any of this will improve readability and would only increase complexity in this instance.

tests/unit/runners/test_filequeue.py
```diff
@@ -0,0 +1,166 @@
...
+    def test_ls(self, fire_event, dirpath):
+        '''
+        Verify that ls returns filenames
+        '''
+        fq = filequeue.FileQueue(root_dir=dirpath)
+        touched = fq.touch("red")
+        touched = fq.touch("blue")
+        files = fq.ls()
+        shutil.rmtree(dirpath)
+        assert set(files) == set(['red', 'blue'])
```
jan--f: The ordering makes this test fail still.

PR #141, opened 2017-03-10
stage 3: add "dev_env" flag to enable/disable 4 node restriction

rjfd:
This PR allows the user to specify that wants to deploy a Ceph cluster in a development environment, and thus avoiding the minimum number of storage nodes restriction.

The specification of then type of environment is done using an env variable called  `DEV_ENV` 

Example: `$ DEV_ENV=true salt-run state.orch ceph.stage.deploy` will run stage 3 with the environment set to "dev".

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -142,10 +142,20 @@ class Validate(object):
         self.data = data
         self.grains = grains
         self.printer = printer
+        self.in_dev_env = self.__dev_env()
         self.passed = OrderedDict()
         self.errors = OrderedDict()
         self.warnings = OrderedDict()
         self._minion_check()
+
+    def __dev_env(self):
+        if 'DEV_ENV' in os.environ:
+            return os.environ['DEV_ENV'].lower() != 'false'
+        elif len(self.data.keys()) > 1:
+            any_minion = self.data.keys()[0]
+            if 'DEV_ENV' in self.data[any_minion]:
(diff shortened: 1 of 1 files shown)


swiftgist:
Were you thinking of future settings for DEPLOY_ENV?  Something besides 'dev'?  Otherwise, I think using the same label (DEV_ENV environment variable matches dev_env validate variable) is enough.

I do know that @theanalyst would like to use the pillar.  What do you think of moving the environ.get into the init? Here's what I'm thinking:

```
    self.dev_env = self._dev_env()


    def _dev_env(self):
        if os.environ.get('DEV_ENV'):
            return os.environ.get('DEV_ENV')
        any_minion = self.data.keys()[0]
        if 'DEV_ENV' in self.data[any_minion]:
            return self.data[any_minion]['DEV_ENV']
        return False
```
That should let the environment override the pillar.  As far as the pillar goes, I think setting it in global.yml would not upset anyone that would use this way.

The only other question is "Would you want the validate message to say something other than valid?"  To let yourself know that the setting is in effect?  I am thinking about others that might use it and we may disable other validations at some point.  Seeing 'dev_env' in green might help.

rjfd:
@swiftgist good suggestion! I changed the implementation to just use "DEV_ENV" env var, and now we also look at the pillar for DEV_ENV in case no env var is defined.
If DEV_ENV is set then we output to the user the info that is running in DEV_ENV

rjfd:
I also updated the PR's description to match the new implementation.

PR #143, opened 2017-03-13
[DNM] Purge: remove cache folder for ganesha and rgw

supriti:

### DIFF
--- a/srv/salt/ceph/rescind/ganesha/default.sls
+++ b/srv/salt/ceph/rescind/ganesha/default.sls
@@ -1,6 +1,9 @@
 ganesha nop:
   test.nop
 
+/srv/salt/ceph/ganesha/cache/:
+    file.absent
+
 {% for role in salt['pillar.get']('ganesha_configurations', [ 'ganesha' ]) %}
 
 /var/lib/ceph/radosgw/ceph-{{role}}.{{ grains['host'] }}:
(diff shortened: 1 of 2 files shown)


swiftgist:
@supriti If I remove the role from one node, but leave others (e.g. moving from four to three gateways), won't I lose the cached values for the remaining gateways?

supriti:
@swiftgist Yes this PR is not correct. I intended to update it, but did not get time. 
My original understanding of rescind was not correct. 

For rescind, we should be removing only the roles that are non-existent.
Step1:
To get list of such roles.
- Get list of currently defined ganesha role:
  using module "ganesha.configurations"
- Iterate over the cache folder and parse the .conf file.
ex: ganesha.data1.conf --> role "ganesha" on node "data1".
- Store the roles defined in cache folder, but not part of ganesha_configurations. 

Step 2:
Remove all the corresponding state for non-existent ganesha roles:
- Delete files from cache
- Uninstall nfs-ganesha package from the host
- Uninstall the keyring files in the target minion
- Delete keyrings from ceph auth.

I think for rgw also we would need something similar.
I will send an updated PR.

jan--f:
Marking Do Not Merge for now.

supriti:
Closing this PR as its not correct. Will open a new one later.

PR #144, opened 2017-03-13
stage 0: force ordering of salt state items

rjfd:
This PR forces the ordering of salt states applied to the minions in stage 0 to guarantee that mine functions have all the package dependencies installed before executing.

### DIFF
--- a/srv/salt/ceph/stage/prep/minion/default.sls
+++ b/srv/salt/ceph/stage/prep/minion/default.sls
@@ -9,11 +9,6 @@ sync:
     - tgt: '*'
     - sls: ceph.sync
 
-mines:
-  salt.state:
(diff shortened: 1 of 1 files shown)


swiftgist:
We should reorder the file.  I found that when adding requires to states, the order is not what you think it might be.  All the requires are run first with any of their dependencies and then any statements without requires are run.  I think it isn't intuitive.  When I started adding requires to everything, separating sls files became painful since a requires cannot depend on something outside of a state file.  If no stanza has a requires then the file will execute top down.  So, I started going with that strategy.

rjfd:
@swiftgist right, and that way is much more cleaner. Removed the requires, changed the order and rebased commit.

PR #146, opened 2017-03-13
Remove duplicate ganesha_members

supriti:


swiftgist:
The call to ganesha_members is still there and it only sets the minion directory and returns.  Shouldn't it be gone as well?

supriti:
Messed up my branch. Opening a new PR.

Issue #148, opened 2017-03-14
deepsea depends on python-netaddr on opensuse 42.2


jan--f:
The net runner imports netaddr. On opensuse 42.2 this requires the installation of python-netaddr.

For some reason this is not a problem in Tumbleweed, although the python-netaddr package is not installed.

jschmid1:
closed via #149

PR #149, opened 2017-03-14
spec file: add python-netaddr dependency for Opensuse 42.2

jan--f:

### DIFF
--- a/deepsea.spec
+++ b/deepsea.spec
@@ -32,6 +32,9 @@ BuildRequires:  salt-master
 Requires:       salt-master
 Requires:       salt-minion
 Requires:       python-ipaddress
+%if 0%{?leap_version} == 420200
+Requires:       python-netaddr
+%endif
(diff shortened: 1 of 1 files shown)


jan--f:
Fixes: #148

deepsea.spec
```diff
@@ -32,6 +32,9 @@ BuildRequires:  salt-master
 Requires:       salt-master
 Requires:       salt-minion
 Requires:       python-ipaddress
+%if 0%{?leap_version} == 420200
```
rjfd: Should we also check if this is necessary in SLE?

jan--f: sure but lets do it in another PR. This is blocking me in Teuthology.

jan--f: @rjfd btw I checked and in sle the salt-master brings in python-netaddr as a dependency
so we should keep an eye on that when sle updates the salt-master.

Issue #147, opened 2017-03-13
Ganesha+RGW: Access_key_Id and Secret_Access_Key is always "None"


supriti:
rgw.add_users is called in stage 4. 
But we call ganesha.config in stage 2. Hence, the keys are not populated.

To resolve this issue, we need to move ganesha.config to stage 4.

PR #153, opened 2017-03-14
Move ganesha config from stage 2 to 4. 

supriti:
rgw creates users at stage 4, which is passed on to ganesha config.

### DIFF
--- a/srv/salt/ceph/ganesha/configure/default.sls
+++ b/srv/salt/ceph/ganesha/configure/default.sls
@@ -1,8 +1,7 @@
 {% for role in salt['ganesha.configurations']() %}
 /etc/ganesha/ganesha.conf:
   file.managed:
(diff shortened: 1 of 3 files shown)


supriti:
Fixes issue #147

swiftgist:
On the first run of a fresh installation, I get the error 

```
----------
   ID: /etc/ganesha/ganesha.conf
   Function: file.managed
   Result: False
   Comment: Unable to manage file: none of the specified sources were found   
```

Jan suggested changing the source list from an array to a value in /srv/salt/ceph/ganesha/configure/default.sls.  This seems to work.


```
    - source: salt://ceph/ganesha/cache/{{ role }}.{{ salt['grains.get']('host') }}.conf
```

supriti:
Pushed new changes

PR #154, opened 2017-03-15
Add cephservices module

swiftgist:
This module is necessary for the upgrade/restart orchestration #43


srv/salt/_modules/cephservices.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/python
+
+import salt.client
+import pprint
```
jschmid1: pprint, os, sys are imported but unused

PR #156, opened 2017-03-21
Fix certificate of origin

BlaineEXE:
Replace the current Developer's Certificate of Origin 1.1 that
questionably may have had modifications with the verbatim version found
at https://developercertificate.org/.

From above website:
> Everyone is permitted to copy and distribute verbatim copies of this
> license document, but changing it is not allowed.

In addition, enclose the Certificate in a code block, as well as the
signed-off-by example so that the sections stand out more easily to the
reader, whether of the md file itself or of the rendered version.

Also fix header H1-H2 mix-up.


theanalyst:
Thanks!

PR #159, opened 2017-03-22
Add restart functionality for ganesha

jschmid1:

### DIFF
--- a/srv/salt/ceph/ganesha/restart/default.sls
+++ b/srv/salt/ceph/ganesha/restart/default.sls
@@ -1,0 +1,20 @@
+restart:
+  cmd.run:
+    - name: "systemctl restart nfs"
+    - unless: "systemctl is-failed nfs"
+    - fire_event: True
+
+  cmd.run:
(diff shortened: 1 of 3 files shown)


srv/salt/ceph/ganesha/restart/default.sls
```diff
@@ -0,0 +1,20 @@
+restart:
+  cmd.run:
```
jan--f: Sorry, I'm out of the loop a little bit...is there a reason for not using the service module? As in service.restart?

jan--f: hmm nevermind. I guess the answer is as usual "it doesn't do what we need" :(

swiftgist:
I believe the state file can call the service module, but that needs to be verified.  I suggest we create a separate issue if that's fine with each of you.

jan--f:
@swiftgist absolutely

jschmid1:
@swiftgist fine for me

swiftgist:
Created #178

PR #160, opened 2017-03-22
spec: leap_version conditional is not reliable.

jan--f:
use sle_version and is_opensuse instead.

### DIFF
--- a/deepsea.spec
+++ b/deepsea.spec
@@ -32,7 +32,7 @@ BuildRequires:  salt-master
 Requires:       salt-master
 Requires:       salt-minion
 Requires:       python-ipaddress
-%if 0%{?leap_version} == 420200
+%if 0%{?sle_version} == 120200 && 0%{?is_opensuse} == 1
 Requires:       python-netaddr
 %endif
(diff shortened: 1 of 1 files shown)


jan--f:
leap_version did not work on our OVH VMs for some reason. This should always work.
SLE does not need this dependency, since the salt-master depends on python-netaddr. We might want to keep this in mind for an update to salt-master though.

tserong:
That change looks fine to me (see the table at https://en.opensuse.org/openSUSE:Build_Service_cross_distribution_howto#Detect_a_distribution_flavor_for_special_code), although the fact that %leap_version isn't set on Leap 42.2 is, IMO, a bug in the rpm package itself which *should* define that macro on Leap systems.

jan--f:
Agreed. According to https://bugzilla.opensuse.org/show_bug.cgi?id=985527 it is fixed. Not sure why I don't get in in an OVH openstack VM.

jan--f:
Its a known issue. Will be fixed in 42.3 at the latest. Until then this workaround is recommended,

PR #162, opened 2017-03-23
extend cephprocesses

jschmid1:
That PR extends cephprocesses.py and allows more fine-grained control over the queried services

If you call `cephprocesses.check()` it will execute status.pid for every service that is registered on that node. There is no way to e.g. _only_ check if monitors are running if the targeted node has also a MDS assigned. By overwriting the `roles`  you can ignore colocated services.

Example: status.pid will return you None if executed on a _hot-stand-by mds_, therefore the 'mon-node' will be considered as failed. That should eventually be tackled in a separate PR.

`salt-run cephprocesses.check roles=['mon']` 

will now _ONLY_ check for the status.pid on every node that has the role 'mon' assigned.

`salt-run cephprocesses.check roles=['mon','mds']` 

will now _ONLY_ check for the status.pid on every node that has the role 'mon' and 'mds' assigned.

`salt-run cephprocesses.check `

 will now check for the status.pid on every node for every assigned role.

### DIFF
--- a/srv/modules/runners/cephprocesses.py
+++ b/srv/modules/runners/cephprocesses.py
@@ -26,7 +26,7 @@ def check(cluster='ceph', roles=[], tolerate_down=0):
     search = "I@cluster:{}".format(cluster)
 
     if not roles:
-        roles_d, roles = _cached_roles(search)
+        roles = _cached_roles(search)
 
     status = _status(search, roles)
 
@@ -90,7 +90,7 @@ def _cached_roles(search):
                 roles.setdefault(role, []).append(minion)
 
     log.debug(pprint.pformat(roles))
-    return roles, roles.keys()
+    return roles.keys()
 
 
 def wait(cluster='ceph', **kwargs):
--- a/srv/modules/runners/cephprocesses.py
+++ b/srv/modules/runners/cephprocesses.py
@@ -64,7 +64,7 @@ def _status(search, roles):
         role_search = search + " and I@roles:{}".format(role)
         status[role] = local.cmd(role_search,
                                  'cephprocesses.check',
-                                 [],
+                                 [roles.keys()],
                                  expr_form="compound")
 
     log.debug(pprint.pformat(status))
--- a/srv/modules/runners/cephprocesses.py
+++ b/srv/modules/runners/cephprocesses.py
@@ -28,6 +28,7 @@ def check(cluster='ceph', **kwargs):
     if 'roles' in kwargs:
         roles = { k:roles[k] for k in kwargs['roles'] }
 
+	
     status = _status(search, roles)
     log.debug("roles: {}".format(pprint.pformat(roles)))
     log.debug("status: {}".format(pprint.pformat(status)))
@@ -45,6 +46,10 @@ def check(cluster='ceph', **kwargs):
     for role in status.keys():
         for minion in status[role]:
             if status[role][minion] == False:
+                # Currently no checking for passive mds.
+                # status.pid returns None as the process is idling.
+                # therefore it will also report that i.e mon is not running
+                # if mon and mds are on the same node
                 log.error("ERROR: {} process on {} is not running".format(role, minion))
                 ret = False
 
@@ -60,6 +65,13 @@ def _status(search, roles):
 
     status = {}
     local = salt.client.LocalClient()
+
+    ignore_roles = [ 'admin', 'master' ]
+
+    for ig_role in ignore_roles:
+      if roles.has_key(ig_role) and roles:
+        roles.pop(ig_role)
+
     for role in roles:
         role_search = search + " and I@roles:{}".format(role)
         status[role] = local.cmd(role_search,
--- a/srv/modules/runners/cephprocesses.py
+++ b/srv/modules/runners/cephprocesses.py
@@ -18,30 +18,23 @@ A secondary purpose is a utility to check the current state of all processes.
 """
 
 
-def check(cluster='ceph', **kwargs):
+def check(cluster='ceph', roles=[]):
     """
     Query the status of running processes for each role.  Also, verify that
     all minions assigned roles do respond.  Return False if any fail.
     """
     search = "I@cluster:{}".format(cluster)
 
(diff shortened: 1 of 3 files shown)


jschmid1:
need to adapt tests. sec

jan--f:
The runner looks good.
Something is fishy with the module though. On a cluster with no rgw role:
```
 salt '*' cephprocesses.check roles=['rgw']
data4:
    True
data6:
    True
mon3:
    True
data3:
    True
data1:
    True
data5:
    True
mon1:
    True
admin:
    True
mon2:
    True
data2:
    True
```
Not sure if the module should be called on its own, but that is surely not the expected output.

jschmid1:
different behavior for:

salt-minion 2016.11.3 (Carbon) vs salt-minion 2015.8.12 (Beryllium)

srv/modules/runners/cephprocesses.py
```diff
@@ -27,6 +27,10 @@ def check(cluster='ceph', **kwargs):
 
     roles = _cached_roles(search)
```
BlaineEXE: It may be useful in this pull request to refactor this whole block from line 28-34 in the updated file.

Using the name `roles` to hold the cached roles and overwriting the variable (but not the cached roles) in the block below is confusing to read. For myself, as a late-comer, it is easy to misunderstand, and I could easily make a mistake understanding the code. Additionally, the test `if 'roles' in kwargs:` is actually unnecessary. Additionally, `k` is a non-descriptive variable name and should be changed.

This code also loops over kwargs['roles'] twice. These may as well be combined, although I don't think performance is a particular concern here.

```py
cached_roles = _cached_roles(search)
roles = {}
for role in kwargs.get('roles', {}):
    if role not in cached_roles:
        log.error('...')
        return False
    roles[role] = cached_roles[role]
```

srv/salt/_modules/cephprocesses.py
```diff
@@ -26,13 +26,24 @@ def check():
 
     ret = True
     if 'roles' in __pillar__:
```
BlaineEXE: There is too much nesting here, which  can make code brittle. I would suggest using a Fail-Fast technique here to reduce this.

There is also a fair bit of code reuse between the if and else blocks with the exception of the print statements. These statements don't indicate that they are pillar-only, so why not print them out for both cases or never at all?
 Additionally, using `result` or `ret` (or their variants) as a variable names is very poor process.

Without checking too much for correctness of my snippet ...
```py
can_continue = True
if 'roles' not in __pillar__:
    return True # can continue
roles = __pillar__['roles']
if only_for:
    roles = only_for['roles']
for role in roles:
    log.info('...')
    for process in processes[role]:
        pid = __salt__['status.pid'](process)
        print "ON ROLE: {}".format(role)
        print "RESULT: {}".format(pid)
        if not pid:
            log.error('...')
            can_continue = False
return can_continue
```

`if not pid` checks if `pid` is empty, and it works, but I think a better test altogether can be constructed. This relies on salt to always output a null string with no extra spaces or newlines for the rest of time, and it relies on `status.pid` to always return a pid or nothing for the rest of time. To protect ourselves from any possible modifications to salt in the future, we are better of testing whether the result we get is one and only one integer value and returning error otherwise.

```py
if not pid.isdigit():
    log.error('...')
```

srv/modules/runners/cephprocesses.py
```diff
@@ -60,11 +69,18 @@ def _status(search, roles):
 
     status = {}
     local = salt.client.LocalClient()
+
+    ignore_roles = [ 'admin', 'master' ]
+
+    for ig_role in ignore_roles:
+      if roles.has_key(ig_role) and roles:
```
theanalyst: probably reversed ie, if `roles and ..` ? if roles doesn't exist, roles.has_key's behaviour would be undefined

srv/salt/_modules/cephprocesses.py
```diff
@@ -12,27 +12,39 @@
...
+                 'ganesha': ['ganesha.nfsd', 'rpcbind', 'rpc.statd']}
 
     ret = True
+
+    def check_process(role):
+        for process in processes[role]:
+            result = __salt__['status.pid'](process)
+            log.info("Pid for process {} is {}".format(process, result))
+            if result == '':
+                log.error("ERROR: process {} for role {} is not running".format(process, role))
```
theanalyst: we could move the log.error to the check_process' ret call and just return `result == ''` here

srv/modules/runners/cephprocesses.py
BlaineEXE:
```py
for ig_role in ignore_roles:
    if not roles:
        # do something because roles dict does not exist
    roles.pop(ig_role, None) # returns None if role doesn't exist & therefore doesn't throw error
```

srv/salt/_modules/cephprocesses.py
```diff
@@ -12,27 +12,39 @@
...
+                 'mds': ['ceph-mds'],
+                 'igw': ['lrbd'],
+                 'rgw': ['radosgw'],
+                 'ganesha': ['ganesha.nfsd', 'rpcbind', 'rpc.statd']}
 
     ret = True
+
+    def check_process(role):
+        for process in processes[role]:
+            result = __salt__['status.pid'](process)
```
BlaineEXE: `result` should be avoided as a variable name. `pid` seems more appropriate here.

srv/salt/_modules/cephprocesses.py
```diff
@@ -12,27 +12,39 @@
...
+            log.info("Pid for process {} is {}".format(process, result))
+            if result == '':
```
BlaineEXE: Salt hasn't proven itself to have stable internals. We should be prepared for Salt to return a string that has whitespace but no content, a `None` value, or error text.

I might suggest testing instead that the result is (or is not) in the form we expect of a pid, which is conveniently just a positive integer.

```py
if not pid.isdigit():
    log.error('...')
```

swiftgist:
merged into wip-updates-and-restart

jschmid1:
merged with #222, closing

PR #165, opened 2017-03-24
push runner: policy.cfg strip comments at end of line

jan--f:
If policy.cfg has a line like
profile-1Disk40GB-1/cluster/*sls #hey there
push would throw an exception. This commit strips comments before
further processing.
Fixes bsc#1030884.

### DIFF
--- a/srv/modules/runners/push.py
+++ b/srv/modules/runners/push.py
@@ -211,6 +211,8 @@ class PillarData(object):
         common = {}
         with open(filename, "r") as policy:
             for line in policy:
+                # strip comments from the end of the line
+                line = re.sub(' #.*$', '', line)
                 line = line.rstrip()
                 if (line.startswith('#') or not line):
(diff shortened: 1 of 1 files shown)


srv/modules/runners/push.py
```diff
@@ -211,6 +211,8 @@ def organize(self, filename):
         common = {}
         with open(filename, "r") as policy:
             for line in policy:
+                # strip comments from the end of the line
+                line = re.sub(' #.*$', '', line)
```
tserong: I know this has already been merged (sorry), but is the leading space in that RE going to be a problem? Like, if someone has a tab character before the comment, or (improbably) starts a comment with no whitespace between the policy line and the comment?

jan--f: right...nice catch. I'm wouldn't be worried about no space...that might be valid. But tabs or multiple spaces is def something that should be filtered. Patch coming up...

PR #167, opened 2017-03-24
push: eat any white space preceding a comment

jan--f:
This takes care of tabs as well. Multiple white spaces get removed by
the following rstrip. '#' without a separating white space are not
stripped, since they might belong to an 're=' modifier. Comments at the
end of lines should be separated by a white space.

### DIFF
--- a/srv/modules/runners/push.py
+++ b/srv/modules/runners/push.py
@@ -212,7 +212,7 @@ class PillarData(object):
         with open(filename, "r") as policy:
             for line in policy:
                 # strip comments from the end of the line
-                line = re.sub(' #.*$', '', line)
+                line = re.sub('\s+#.*$', '', line)
                 line = line.rstrip()
                 if (line.startswith('#') or not line):
                     log.debug("Ignoring '{}'".format(line))


swiftgist:
okay, now you're making me pedantic. :)  what if there's more than one whitespace?

srv/modules/runners/push.py
```diff
@@ -212,7 +212,7 @@ def organize(self, filename):
         with open(filename, "r") as policy:
             for line in policy:
                 # strip comments from the end of the line
-                line = re.sub(' #.*$', '', line)
+                line = re.sub('\W#.*$', '', line)
```
tserong: Shouldn't that be `\s`?

jan--f: Correct...```\W``` just happened to work too.

jan--f:
@swiftgist If there are more white spaces the ```rstrip()``` in line 216 will remove those.

tserong:
The changes to srv/salt/_modules/proposal.py and junit-py27.xml seem unrelated...?

jan--f:
Argh...should've had a coffee before. not supposed to be in there at all.

jan--f:
@tserong thx for keeping an eye on my fat fingers.

tserong:
:-)

PR #173, opened 2017-03-27
Fix for re-run of stage.3 in case storage:data+journals is used

Martin-Weiss:
When re-running stage.3 a prepare and activate is done even if the OSD is already created and on disk.

### DIFF
--- a/srv/salt/ceph/osd/default.sls
+++ b/srv/salt/ceph/osd/default.sls
@@ -23,13 +23,13 @@ activate {{ device }}:
 prepare {{ data }}:
   cmd.run:
     - name: "ceph-disk -v prepare --fs-type xfs --data-dev --journal-dev --cluster {{ salt['pillar.get']('cluster') }} --cluster-uuid {{ salt['pillar.get']('fsid') }} {{ data }} {{ journal }}"
-    - unless: "fsck {{ data }}"
+    - unless: "fsck {{ data }}1"
     - fire_event: True
 
 activate {{ data }}:
   cmd.run:
     - name: "ceph-disk -v activate --mark-init systemd --mount {{ data }}"
-    - unless: "grep -q ^{{ data }} /proc/mounts"
+    - unless: "grep -q ^{{ data }}1 /proc/mounts"
     - fire_event: True
 
 {% endfor %}


tserong:
I haven't checked, so apologies if I'm wrong, but sticking a "1" on the end might not work in the case of using /dev/disk/by-path or by-id device names.

Martin-Weiss:
You are right - there we would require to add "-part1" or similar:

Example from /dev/disk/by-path/

lrwxrwxrwx 1 root root   9 Mar 19 15:39 pci-0000:00:17.0-scsi-1:0:0:0 -> ../../sdb
lrwxrwxrwx 1 root root  10 Mar 19 15:39 pci-0000:00:17.0-scsi-1:0:0:0-part1 -> ../../sdb1
lrwxrwxrwx 1 root root  10 Mar 19 15:39 pci-0000:00:17.0-scsi-1:0:0:0-part2 -> ../../sdb2
lrwxrwxrwx 1 root root  10 Mar 19 15:39 pci-0000:00:17.0-scsi-1:0:0:0-part3 -> ../../sdb3

Example from /dev/disk/by-id/

 lrwxrwxrwx 1 root root    9 Mar 19 15:39 ata-Hitachi_HDS722020ALA330_JK1175YAK70HWX -> ../../sdb
lrwxrwxrwx 1 root root   10 Mar 19 15:39 ata-Hitachi_HDS722020ALA330_JK1175YAK70HWX-part1 -> ../../sdb1
lrwxrwxrwx 1 root root   10 Mar 19 15:39 ata-Hitachi_HDS722020ALA330_JK1175YAK70HWX-part2 -> ../../sdb2
lrwxrwxrwx 1 root root   10 Mar 19 15:39 ata-Hitachi_HDS722020ALA330_JK1175YAK70HWX-part3 -> ../../sdb3

Question: it seems that this is also not handled properly in the other areas of the code, then - do I understand this right?

tserong:
As discussed the storage:data+journals case probably needs a `readlink` added as is used in the storage:osds case, which will fold the device back to the simple form that works with the 1 suffix. Probably should merge this and #175.

swiftgist:
Still haven't had the chance to verify this on real hardware.  Trying to reconcile how this has been working without the suffixes.

jan--f:
@Martin-Weiss also closing for the same reason as #175.

PR #175, opened 2017-03-27
The activate for journal on separate disk seems to be missing the par…

Martin-Weiss:
…tition number

I have added the "1" as otherwise the activate will fail in a setup where journal is on a separate disk.
(in case of storage:data+journals)

### DIFF
--- a/srv/salt/ceph/osd/default.sls
+++ b/srv/salt/ceph/osd/default.sls
@@ -28,7 +28,7 @@ prepare {{ data }}:
 
 activate {{ data }}:
   cmd.run:
-    - name: "ceph-disk -v activate --mark-init systemd --mount {{ data }}"
+    - name: "ceph-disk -v activate --mark-init systemd --mount {{ data }}1"
     - unless: "grep -q ^{{ data }} /proc/mounts"
     - fire_event: True
 


tserong:
I haven't checked, so apologies if I'm wrong, but sticking a "1" on the end might not work in the case of using /dev/disk/by-path or by-id device names.

Martin-Weiss:
You are right - there we would require to add "-part1" or similar:

Example from /dev/disk/by-path/

lrwxrwxrwx 1 root root 9 Mar 19 15:39 pci-0000:00:17.0-scsi-1:0:0:0 -> ../../sdb
lrwxrwxrwx 1 root root 10 Mar 19 15:39 pci-0000:00:17.0-scsi-1:0:0:0-part1 -> ../../sdb1
lrwxrwxrwx 1 root root 10 Mar 19 15:39 pci-0000:00:17.0-scsi-1:0:0:0-part2 -> ../../sdb2
lrwxrwxrwx 1 root root 10 Mar 19 15:39 pci-0000:00:17.0-scsi-1:0:0:0-part3 -> ../../sdb3

Example from /dev/disk/by-id/

lrwxrwxrwx 1 root root 9 Mar 19 15:39 ata-Hitachi_HDS722020ALA330_JK1175YAK70HWX -> ../../sdb
lrwxrwxrwx 1 root root 10 Mar 19 15:39 ata-Hitachi_HDS722020ALA330_JK1175YAK70HWX-part1 -> ../../sdb1
lrwxrwxrwx 1 root root 10 Mar 19 15:39 ata-Hitachi_HDS722020ALA330_JK1175YAK70HWX-part2 -> ../../sdb2
lrwxrwxrwx 1 root root 10 Mar 19 15:39 ata-Hitachi_HDS722020ALA330_JK1175YAK70HWX-part3 -> ../../sdb3

Question: it seems that this is also not handled properly in the other areas of the code, then - do I understand this right?

tserong:
As discussed the storage:data+journals case probably needs a `readlink` added as is used in the storage:osds case, which will fold the device back to the simple form that works with the 1 suffix. Probably should merge this and #173.

chabowski:
Von meinem Samsung Galaxy Smartphone gesendet.


-------- Ursprüngliche Nachricht --------
Von: Tim Serong <notifications@github.com>
Datum: 31.03.17 14:29 (GMT+01:00)
An: SUSE/DeepSea <DeepSea@noreply.github.com>
Cc: Subscribed <subscribed@noreply.github.com>
Betreff: Re: [SUSE/DeepSea] The activate for journal on separate disk seems to be missing the par… (#175)


As discussed the storage:data+journals case probably needs a readlink added as is used in the storage:osds case, which will fold the device back to the simple form that works with the 1 suffix. Probably should merge this and #173<https://github.com/SUSE/DeepSea/pull/173>.

swiftgist:
I am wanting to double check this on the Blueshark cluster, but some setup is required.  I am wanting to understand how this didn't cause an issue before.

jan--f:
@Martin-Weiss the ods deployment was recently completely overhauled. I'll close this, since its not applicable anymore and I believe your issue should be solved too.

PR #185, opened 2017-04-03
Add linter

### LINKED ISSUE #107: Use pylint or any other linter for syntax checks
As we're adding more and more code in the runner, we tend to error, a linter should be able to catch many of them, though at this moment we need a lot more code style fixes before we can have a linter run clean. 

We could use salt-pylint and borrow salt's own pylintrc config which should allow us to not report syntax errors on things salt monkey patches in the global namespace (for eg `__salt__` and `__opts__` builtins to name a few)

cc @rjfd @jschmid1 @swiftgist @jan--f

BlaineEXE:
Add a linter to the project. In this case, we are using saltpylint as per discussion in #107. As part of this effort:

- Add linter config file copied from saltstack/salt (with a few modifications)
- Add requirements.txt which includes saltpylint and all of our related needs
- Add a script to create and bootstrap the venv so it's easier for other contributers to get the linter environ up and running quickly

Closes #107

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,439 @@
+[MASTER]
+# Specify a configuration file.
+#rcfile=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Profiled execution.
+profile=no
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# List of plugins (as comma separated values of python modules names) to load,
+# usually to register additional checkers.
+load-plugins=saltpylint.pep8,
+  saltpylint.pep263,
+  saltpylint.strings,
+  saltpylint.fileperms,
+  saltpylint.py3modernize,
+  saltpylint.smartup,
+  saltpylint.minpyver,
+  saltpylint.blacklist,
+  saltpylint.thirdparty
+
+# Use multiple processes to speed up Pylint.
+# Don't bump this values on PyLint 1.4.0 - Know bug that ignores the passed --rcfile
+jobs=1
+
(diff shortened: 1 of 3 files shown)


BlaineEXE:
As an aside: It may become necessary in the future to specify in the requirements.txt a specific version or range of versions of Salt if we discover that DeepSea is dependent on specific Salt versions.

BlaineEXE:
I changed line length to 100. I think this seems a nice number in between 80 and 120. Also note that pylint should ignore lines longer than that if they are a web url.

jschmid1:
damn, came too late for the party..

With the introduction of virtualenv we kinda have a redundancy with tox now. 

Tox already creates, based on the tox.ini, a virtualenv with `[tox]: envlist = py2.7` and installs

deps = [dep1,dep2]

So instead of recreating a virtualenv with defining a own requirements.txt we could just hook into the existing venv with `source .tox/py27/bin/activate` and put the new requirements the `deps = ` stanca.

Just a thought. don't know about the sideeffects that might arise.

jschmid1:
Oh, and it clashes with tox's env


`WC1 /home/jxs/projects/deepsea/venv/lib64/python2.7/site-packages/pylint/test/extensions/test_elseif_used.py cannot collect test class 'TestReporter' because it has a __init__ constructor`

theanalyst:
@jschmid1 I'm also in favor of using tox for everythign since it creates venvs by default and can support multiple pythons if we get there. it shouldn't be difficult to create a different tox task for pylint so that running say tox pylint would give us the same results

BlaineEXE:
We can un-do the merge, reopen this pull, and I can try to integrate this into tox if that sounds good.

theanalyst:
@BlaineEXE we don't need to undo the merge, the lint files and everything is still the same, we just drop bootstrap script once we have a working tox

BlaineEXE:
See #190

PR #188, opened 2017-04-05
Fixes rescind of storage for ceph.purge

swiftgist:
Commit bbf8df8 removed the storage role to prevent the creation of a directory that confused users; however, available_roles must still contain storage.  This fixes the ceph.purge functionality.


jan--f:
hmm hang on. disks are not pruged with this branch.

jan--f:
see #191 for further discussion

PR #190, opened 2017-04-05
Add linter to tox

BlaineEXE:
Remove the bootstrap script that created the 'venv' virtualenv since
having it activated while running tox caused tox's unit tests to fail.
Instead, the dev can activate one of tox's virtualenvs to use locally
if they so desire.

Add a 'lint' test environment to tox with all requirements from the
requirements.txt file (now deleted) for linting. The linter will search
through the directories srv/ and etc/. Though etc/ is unlikely to ever
have any python files, it is included just in case. Pylint returns
nonzero if it detects any warnings/errors, so tox reports a failure.

Set tox to run linting in 5 threads (ideal for 4-core workstations,
which seems a reasonable assumption) so tox will run the linting  more
quickly. `jobs=1` was left at 1 in the `.pylintrc` file for best
compatibility when running using local tools. Atom's pylint linter
encountered an error when setting `jobs=5` on some files, for example.

To run all tests, continue to run `tox`
To run only lint tests, run `tox -e lint`
To run only python 2.7 tests, run `tox -e py27`

Linter errors fail Travis CI, so also modify the travis config to allow
for linter errors.


BlaineEXE:
We may have to change how travis works so we can flag but not fail on linter errors.

jschmid1:
That'd involve changing the .travis.yml

to

```yml
language: python
python:
  - "2.7"
install: pip install tox
script: 
  - tox -e py27
  - tox -e pylint
cache: pip
```

-> separate the two checks and (I guess) we should be able to mark them as a hard-requirement individually.

and @swiftgist(or some SUSE adm) hast to do that:
https://help.github.com/articles/enabling-required-status-checks/

theanalyst:
We could use the matrix option in tox:

```yml
language: python
python:
  - 2.7
install: pip install tox
env: TOXENV=py27
matrix:
  include:
    - env: TOXENV=py27
      python: 2.7
    - env: TOXENV=lint
      python: 2.7
  allow_failures:
    - env: TOXENV=lint
script:
  - tox -e $TOXENV
cache: pip
```

BlaineEXE:
If you don't mind, @theanalyst and/or @jschmid1, take a look at the Travis run (now passing) and let me know if you think it looks good.

jschmid1:
It'd be nice to see the linter result as a "separate check". If it's hidden behind the unittests check, it will never catch our eyes again and it's basically useless.

Does anyone know if the matrix syntax allows to split the checks in github? @theanalyst

BlaineEXE:
I'll do a little digging this morning.

BlaineEXE:
It appears the Travis devs are open to the feature, but it's not implemented yet (and may not be ever): https://github.com/travis-ci/travis-tasks/pull/78

We could do it with this, but we have to host a web app to interface with: https://github.com/wizeline/travieso

We could also do it if we set up a Jenkins server for the builds as Ceph has done, but this is not a trivial amount of work. Unless SUSE has a Jenkins server we can use for our builds?

Side note: I added the `fast_finish: true` line to the matrix which will allow the build to be reported pass/fail on GitHub before linting has finished (since we allow the linter test to fail). The linter test takes a bit longer to run than the unit test.

BlaineEXE:
Any thoughts @theanalyst, @jschmid1, or @jan--f? I'd like to be sure to get a couple approvals now after I've updated Travis settings before I merge.

PR #193, opened 2017-04-11
Add minimal Debian build files

tserong:
This should be the minimum set of files necessary to build for
Debian and/or Ubuntu.  To use this on build.opensuse.org, the
changelog, compat, control and rules files need to be included
in the OBS project, renamed to debian.changelog, debian.compat,
debian.control and debian.rules.  OBS also requires a deepsea.dsc
file, which AIUI would ordinarily be generated by dpkg-source,
which is why I haven't included it in this commit.  For an
example build on OBS, see:

  https://build.opensuse.org/package/show/home:tserong:branches:home:swiftgist/deepsea

Note that this will build for Debian 8.0 and Ubuntu 16.x, but
not for Debian 7.0 or Ubuntu 14.x, as the latter two don't seem
to have salt-master available.

Also note that this is missing deepsea.postinst, which we
probably want to add at some point to run the same commands
as are in the %post section of the RPM spec file (for some
reason I couldn't get the postinst script to be included
when doing test builds on OBS, so I thought it best to leave
that for later).


tserong:
Note that this will not necessarily give a _runnable_ DeepSea on Debian/Ubuntu -- I gave it a quick try on an Ubuntu 16.04 VM, but stage 0 fails with:

````
----------
          ID: common packages
    Function: salt.state
      Result: False
     Comment: Run failed on minions: ubuntubox
              Failures:
                  ubuntubox:
                  ----------
                            ID: stage prep dependencies
                      Function: pkg.installed
                        Result: False
                       Comment: The following packages failed to install/update: lsscsi, gptfdisk The following packages were already installed: pciutils
                       Started: 08:28:40.470363
                      Duration: 5233.874 ms
                       Changes:   
                  
                  Summary for ubuntubox
                  ------------
                  Succeeded: 0
                  Failed:    1
                  ------------
                  Total states run:     1
     Started: 08:28:39.961022
    Duration: 5910.169 ms
     Changes:   
  Name: updates - Function: salt.state - Result: Changed
----------
          ID: restart
    Function: salt.state
      Result: False
     Comment: Run failed on minions: ubuntubox
              Failures:
                  ubuntubox:
                      Data failed to compile:
                  ----------
                      Rendering SLS 'base:ceph.updates.restart.default' failed: mapping values are not allowed here; line 11
                  
                  ---
                  [...]
                  
                  warning:
                    module.run:
                      - name: advise.reboot
                      - running: 4.4.0-22-generic
                      - installed: /bin/sh: 1: rpm: not found    <======================
                      - unless: "echo /bin/sh: 1: rpm: not found | grep -q 4.4.0-22-generic"
                  
                  
                  reboot:
                    cmd.run:
                  [...]
                  ---
     Started: 08:31:30.767291
    Duration: 533.927 ms
     Changes:   
  Name: mines - Function: salt.state - Result: Changed
  Name: complete - Function: salt.state - Result: Clean

Summary for ubuntubox_master
-------------
Succeeded: 13 (changed=5)
Failed:     2
-------------
Total states run:     15
````

jan--f:
thx @tserong, I've been meaning to fix the stage 0 suse-ism anyway. will merge this once they are out.

PR #194, opened 2017-04-11
pkg.install: add distribution conditional

jan--f:
This accounts for different package names in various distros. More to be
added later.

### DIFF
--- a/srv/salt/ceph/packages/common/default.sls
+++ b/srv/salt/ceph/packages/common/default.sls
@@ -1,7 +1,8 @@
+{% set os = salt['grains.get']('os') %}
 
+{% if os == 'SUSE' %}
 
-
-stage prep dependencies:
+stage prep dependencies suse:
   pkg.installed:
     - pkgs:
       - lsscsi
@@ -9,3 +10,19 @@ stage prep dependencies:
       - gptfdisk
     - fire_event: True
 
+{% elif os == 'Ubuntu' %}
+
+stage prep dependencies ubuntu:
+  pkg.installed:
+    - pkgs:
+      - lsscsi
+      - pciutils
+      - gdisk
+    - fire_event: True
+
+{% else %}
+
+nop:
+  test.nop
+
+{% endif %}


jan--f:
@tserong this should take care of the package install fail in #193. Could you give it a spin?

srv/salt/ceph/packages/common/default.sls
```diff
@@ -1,11 +1,28 @@
+{% os = salt['grains.get']('os') %}
```
tserong: Needs to be `set os = ...`

tserong: Also, we should probably do this as `os_family`, rather than `os`, because `os_family` is `Debian` for both Debian and Ubuntu, which'll save an if later.

srv/salt/ceph/packages/common/default.sls
```diff
@@ -1,11 +1,28 @@
...
 
+{% if os is 'SUSE' %}
```
tserong: Use `==` instead of `is`

srv/salt/ceph/packages/common/default.sls
```diff
@@ -1,11 +1,28 @@
...
+stage prep dependencies suse:
   pkg.installed:
     - pkgs:
       - lsscsi
       - pciutils
       - gptfdisk
     - fire_event: True
 
+{% elif os is 'Ubuntu' %}
+
```
tserong: use `== 'Debian'` (assuming change to os_family as suggested above)

srv/salt/ceph/packages/common/default.sls
```diff
@@ -1,11 +1,28 @@
...
+stage prep dependencies ubuntu:
+  pkg.installed:
+    - pkgs:
+      - lsscsi
+      - pciutils
+      - gdisk
```
tserong: Package set looks fine :-)

srv/salt/ceph/packages/common/default.sls
tserong: Oh, and if we change to `os_family` it needs to be `Suse` (different case)

srv/salt/ceph/packages/common/default.sls
jan--f: Do you know if the packages are good on Debian too? This would make the difference between using ```os``` or ```os_family```.

tserong: Googling suggests those names work for Debian too (but I haven't tested here, don't have a Debian box handy)

jan--f: k, then I'd propose to move to os_family once we can confirm the packages are the same.

tserong:
de9454f works fine for me

PR #198, opened 2017-04-11
Removed installing and enabling openattic-rpcd

LenzGr:
openATTIC 2.0.19 removed the XML-RPC API provided by the `openattic-rpcd` background service. Updated the Salt state file accordingly.

Signed-off-by: Lenz Grimmer <lgrimmer@suse.com>

### DIFF
--- a/srv/salt/ceph/openattic/default.sls
+++ b/srv/salt/ceph/openattic/default.sls
@@ -5,11 +5,6 @@ install openattic:
     - pkgs:
       - openattic
 
-enable openattic-rpcd:
-  service.running:
-    - name: openattic-rpcd
-    - enable: True
-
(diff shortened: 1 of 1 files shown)


jan--f:
Are we breaking older versions of Openattic with this PR? Say someone has a pre-2.0.19 version in their repos, will oA still work for them?

holgis:
> Are we breaking older versions of Openattic with this PR? Say someone has a pre-2.0.19 version in their repos, will oA still work for them?

If that's the case we may need to have separate brach for SES4 then.

LenzGr:
The `openattic-rpcd` is not actively used in the context of Ceph management, so this has no impact on older versions.

PR #202, opened 2017-04-13
Only run pytests for rpm target

swiftgist:


BlaineEXE:
It looks like I missed this ... I'm not sure this is the appropriate way to achieve what you're trying to do. All you needed to do was change the line `tox` to `tox -e py27`. Adding another file just adds noise to the repo.

PR #207, opened 2017-04-14
[DNM] Replace NTP state with ntp-formula

### LINKED ISSUE #91: ntp client configuration
DeepSea currently forces synchronization, but does not write an ntp.conf to the minions.  The request is to add the client configuration.

BlaineEXE:
Adhere to spec for https://github.com/SUSE/DeepSea/wiki/Time-Service

- Create global.yml config for new NTP management strategy. Remove
default/global.yml entirely.
- Remove existing 'ntp' state files. Replace with ntp-formula.
- Update stage to support new NTP config options.
- Update validate for new NTP strategy.

Resolves #91 


BlaineEXE:
I manually tested in the VMs by running `salt '*' state.apply ceph.time` and verified that editing/removing config file variables behaved as I expected.

Something I noticed is that it can take a few minutes for minions to sync to the master if they are far askew. I think it may be more appropriate to move ntp to stage 2 (configure) or to stage 0 (prep) so that stage 3 (deployment) will have synchronized time. Any thoughts as to which stage is more appropriate? Stage 0 might make more sense, but users may sometimes opt not to run stage 0. I could forsee this being a problem especially when adding nodes to a cluster; a user decides they don't need stage 0 or forgets to run it when adding a node, and then NTP isn't synchronized. If it's in stage 2, it's more likely to be run.

For my manual tests, I edited the `srv/pillar/ceph/stack/global.yml` file and verified that the `/etc/ntp.conf` files matched what I would expect based upon custom input or whether it should use defaults. Checking the pillar with `salt '*' pillar.items` was a good test too. I also verified that time was sync-ing properly with `ntpq -p`. These are obviously more integration/functional tests than unit tests. Do we have a way to run these?

BlaineEXE:
One of my still todo items is to validate that time sync is working properly. I've considered a few avenues so far:

1) Generically, check time on each server and verify that the difference between the min time and max time is less than X seconds. We would need to figure out what X should be. This helps to ensure servers are roughly in sync at least at present. This test can be done in addition to the other tests. We might have to write a special test to be able to collect all times and compare min/max between them all.
2) Check that time is being synchronized via at least one of the common time sync methods: timesyncd, ntpd, chronyd. This is broadly applicable, but then we have to support and test all the options.
3) If and only if DeepSea is configured to sync time, verify that time is being synchronized via ntp.

I am leaning toward option 3 as a definite. Option 2 seems like it is likely to be more work than we have time for and not worth the effort and risk of fragility. Option 1 seems useful and generally relevant, but it may be a non-trivial amount of effort. I haven't looked closely at the verify code yet to say for sure.

swiftgist:
I can see ceph.time called during Stage 0.  The time related validation would need to move from validate.pillar to validate.setup.  (Note: these names likely need to be fixed since we will have validation for Stage 1 and 2 soon.)  

As far as skipping Stage 0, new servers would still need those steps (mines, modules, etc.).  I think a Stage 1 validation may help here, but that can be done separately.

Creating a runtime test of time would be nice as well as using it as part of a validation.  Maybe use an initial range of 1 second, but only give a warning rather than an error with a failure.  (I'm thinking conservatively here for existing clusters where new servers are added but could be ignored for the moment.)  I think your Option 1 would cover the others.  A runner calling date +%s.%N (or the salt equivalent) and comparing the results should work and would signify that the time looks good regardless of the method or the time source.

BlaineEXE:
While working on validate today and investigating ways to verify NTP configs, I have been thinking that I think it best only modify validate in this pull request as much as it pertains to making it still work as it was with the new NTP setup. Moving NTP validation to a new stage and changing the validate test itself are both separate changes (from each other and from this pull request).

jan--f:
For checking if time is synced a call to ```timedatectl``` should suffice. I'd rather not see trying to ascertain a minimal offset with salt. That just screams corner cases.

BlaineEXE:
Testing on my end has all been successful:

Testing with `salt '*' state.apply ceph.time`:
- `ntp_server` not present or with no value, all nodes are clients
- `higher_stratum_servers` not present or no value, higher stratums are {0-3}.pool.ntp.org on nodes
- `higher_stratum_servers` set to custom values, custom values are propagated to ntp.conf on nodes

Testing with `salt-run state.orch ceph.stage.3`:
- All nodes configured with an ntp.conf when `manage` is `true`
- `manage` not present or with no value, NTP is not configured on any nodes
- entire `time_service` block removed, NTP is not configured on any nodes

BlaineEXE:
@jan--f, after thinking through trying to ascertain a minimal offset validation, I think it best not to go that route. It's not a trivial thing to measure when there are network latencies involved, especially if a Ceph cluster is distributed worldwide.

swiftgist:
I will say that I am uncertain about submodules.  We do not have any other components relying on submodules.  I think I would like the others to comment.  Also, I do not know if there's any concerns in relying on repos outside of github.com/SUSE (which may one day be github.com/Ceph).

With respect to the PR, I started to make changes to the spec file, but realized this will take more time than I have.  Would you update the spec file as well?

srv/modules/runners/populate.py
```diff
@@ -768,15 +768,15 @@ def _global(self):
...
-        self.writer.write(filename, contents)
+        # stack_dir = "{}/config/stack/default".format(self.root_dir)
+        # if not os.path.isdir(stack_dir):
+        #      _create_dirs(stack_dir, self.root_dir)
+        # filename = "{}/global.yml".format(stack_dir)
+        # contents = {}
+        # contents['time_server'] = '{{ pillar.get("master_minion") }}'
+        # contents['time_service'] = 'ntp'
+        #
+        # self.writer.write(filename, contents)
```
jan--f: I'm in favour of deleting old code. We can always get it back via git.

srv/pillar/ceph/stack/stack.cfg
```diff
@@ -1,6 +1,6 @@
 {### Default settings are generated - manual changes will be overwritten ###}
 {# global settings #}
-default/global.yml
+{# default/global.yml}
```
jan--f: This probably shouldn't be in here anymore?

BlaineEXE:
@swiftgist, I had intended to update the specfile myself once the code changes were approved from a high level. As for subrepos, we can do this pull with or without a subrepo. With allows us to pull the latest changes to ntp-formula more easily for bug fixes, etc. You'll see one of my TODO items is to change the hosting location of the subrepo to the SUSE org if we do decide to go through with this design. We definitely need to maintain our branch, and it doesn't belong in my personal space; I just did that for quick POC. We could use something called a subtree (https://www.atlassian.com/blog/git/alternatives-to-git-submodule-git-subtree) but I'm wary since it will merge the git histories together; I'm not sure we want that.

@jan--f, the commented-out lines aren't deleted in case for some reason we really want to have code to generate the default file, making it easy for me to revert the changes. I personally think it's non-transparent to the user, which is why I removed those lines to begin with. I'll set about removing those references entirely today.

BlaineEXE:
After poking at the spec file for a bit, I am actually leaning toward copying the files we want from https://github.com/saltstack-formulas/ntp-formula and discarding files we don't use. The `local_server` and `ng` directories, and `ntpdate.sls` and `server.sls` files are totally unused. We can also omit the `README.rst` and `pillar.example' files, retaining only the `LICENSE` file. Additionally, we should probably put a file stating where the files originate. I'm concerned about the licensing, however. Shot in the dark here, but perhaps @jecluis would have some insight?

BlaineEXE:
In order to reduce the verbosity of the `global.yml` file (the extensive comments make it a bit bloated), I have documented in the wiki requirements for and usage of the NTP-related sections of the config file here: https://github.com/SUSE/DeepSea/wiki/Time-Service

BlaineEXE:
I've now done all the cleanup I need for this feature relating to commit and pull messages. Before getting a final OK, I want to run through my test matrix again to make sure I didn't break anything while rebasing.

BlaineEXE:
Copying my test matrix here since it seems to have gotten lost (was in a previous comment):

Test proper NTP deployment - `salt '*' state.apply ceph.time
- `ntp_server`
  - Set to `{{ pillar.get('master_minion') }}`
  - Set to `mon1.ceph`
  - Blank
  - Key removed
- `higher_stratum_server`
  - Set to [0-3].pool.ntp.org
  - Set to [0-2].opensuse.pool.ntp.org
  - Blank
  - Key removed

Test staging - `salt-run state.orch ceph.stage[0-3]`
- `manage`
  - `true`
  - `false`
  - Blank
  - Key removed
- `time_service`
  - Key removed

BlaineEXE:
All tests pass on my end. Ready to merge.

BlaineEXE:
I updated `global.yml` to move the `time_service` section under the `deepsea` namespace. I think `deepsea` is less likely to confuse/collide with other instances of `ceph`; however, changing this from `deepsea` to `ceph` is a simple find/replace if there are objections.

BlaineEXE:
@swiftgist in an email had concerns about delivering `global.yml` in the RPM. In this pull, I have included the `noreplace` directive, which I think should be sufficient. I have opted to include `global.yml` in the source base and as part of initial installation as it is much more transparent to the user. Having config files generated by code during staging was very counter-intuitive to me when trying to figure all this out. Additionally, I found having defaults and user settings also counter-intuitive. I think it will be a lot easier for users to grok what's going on in the settings files when the settings aren't spread across multiple files and when they don't appear as if by magic somewhere between install and stage 3.

swiftgist:
It's not just about the noreplace.  What will happen when the user file is not replaced and we have added new values to global.yml.  Will DeepSea break?  

I agree using a runner to write these static values into the proposals area isn't intuitive, but treating static data one way and dynamic data another creates two different paths to maintain/debug.

Also, the stack.cfg has two sets of five parts.  Why would I want 4/5 to work one way and 1/5 to work another?  Will I be endlessly explaining the difference to users between bug reports, emails, etc.?

Salt was missing merge functionality and does not have a trivial way to support default structures with unique keys.  I included stack.py originally.  Saltstack has added it since.  From there, I tried to stick to one writer, one file.  That eliminates a slew of problems in debugging the field.  (There are a couple of exceptions currently.)  Salt minions have their own data, like it or not.  Not repeating common data seemed like a good idea.  At this point, the configuration has multiple files.

In the end, I will not blame the user for neglecting to change a dozen settings after an upgrade of DeepSea and subsequently damaging their Ceph cluster.  DeepSea will be better than that.

If you had questions, you could have asked me.

BlaineEXE:
I still think that, as a user, having configuration files appear by magic in a way that isn't documented or expected is a bit disconcerting. The data was before generated by the code with static (not calculated or otherwise processed) fields. I see no reason why this step of indirection should be necessary. It only leaves more things that can break for the devs and more confusion for the user. 

As an attempt to find a middle ground, I propose that I move the static config that currently exists in `global.yml` to the `default/global.yml` instead. This way, the default can be updated by the RPM, and the user's custom `global.yml` won't be touched.

It may also be good to provide some mechanism at RPM installation that copies the `default/global.yml` to `global.yml` if `global.yml` doesn't exist so (1) the user has a place to modify their config (if desired) and (2) updates to DeepSea defaults later won't affect the current install (current config file) if the user doesn't do the copy themselves.

Thoughts?

jecluis:
>  Additionally, we should probably put a file stating where the files originate. I'm concerned about the licensing, however. Shot in the dark here, but perhaps @jecluis would have some insight?

IANAL, and one should probably be consulted in this case, so take the following with a grain of salt.

As I understand from https://www.apache.org/licenses/GPL-compatibility.html and in specific from

> Apache 2 software can therefore be included in GPLv3 projects, because the GPLv3 license accepts our software into GPLv3 works. However, GPLv3 software cannot be included in Apache projects

Then I would assume **including** the files is fine. From the Apache License v2.0, you have to retain the copyright notice, license, etc, of the work you are including, modifying, distributing, etc. (see https://www.apache.org/licenses/LICENSE-2.0.html#redistribution , 4.d).

In the case of DeepSea, I'm guessing it would then become imperative to add a NOTICE file pointing which files came from `saltstack-formulas/ntp-formula`, and retaining their LICENSE in the NOTICE file (maybe?).

Next problem would be "what about if we modify those files?", to which, IIUC, the Apache License v2.0 is quite permissive about. From 2.,

> Subject to the terms and conditions of this License, each Contributor hereby grants to You a perpetual, worldwide, non-exclusive, no-charge, royalty-free, irrevocable copyright license to reproduce, prepare Derivative Works of, publicly display, publicly perform, **sublicense**, and distribute the Work and such Derivative Works in Source or Object form.

and from 4.d,

> You may add Your own copyright statement to Your modifications and may provide additional or different license terms and conditions for use, reproduction, or distribution of Your modifications, or for any such Derivative Works as a whole, provided Your use, reproduction, and distribution of the Work otherwise complies with the conditions stated in this License.

So I'm guessing all changes to those files could simply be GPLv3, as long as made clear and obvious.

Alternatively, if you want to avoid all this confusion, go with a submodule. All the changes you wish to do, do them on the submodule, while retaining the original license.

In any case, I will reinforce that my parents **DID NOT** raise a lawyer (that I know about, *yet*), and I totes probably have no idea what I'm talking about.

l-mb:
It would appear inclusion of this file should be fine given we're including an Apache 2.0 licensed file into GPLv3 and not vice-versa.

BlaineEXE:
Per what I've read at this page, I've added a section to our `COPYRIGHT` file informing of Apache 2.0 copyrighted materials.

BlaineEXE:
@jschmid1 brought up a good point in a chat earlier today: Since this PR (1) moves time service settings under the `ceph` namespace and (2) changes settings altogether, what does the upgrade from a previous DeepSea version to this one look like?

I have a few thoughts so far, but I'm still chewing on them:

1. We can create a runner similar to the `Validate` class to handle deprecations/upgrades.
  **a.** This could optionally open the user's `global.yml`, read the old config, interpret how the config must change for the new settings, and write the new, upgraded `global.yml` back. This is user-friendly but opaque to the user, as it happens 'as if by magic'.
  **b.** OR we could simply inform the user that they have a deprecated setting and give them a message telling how to update the config per the latest changes. This should tell what settings are deprecated and where to look to know how to configure the setting for themselves (in this case, the wiki)
2. We can support the old settings and the new settings together ... though I think this defeats the point of creating a ceph namespace in the first place.
3. ???

jan--f:
I finally got around to giving it a spin...sorry it took so long. First impression is good: installed it, ran ```salt '*' state.apply ceph.time``` and everythings dandy. I'd like to point out two things that will need to be addressed though:

- If the systemd timesynd is running, ntp is started along-side timesyncd. I think this is very much undesirable since they both try to manage the system time. Maybe it would be feasible to stop any other time service when ntp is started? At least systemd-timsyncd and maybe even chrony if present?

- Starting to use DeepSeas ntp formular clobbers an existing ntp.conf. When one wants to go back to not have their time managed by DeepSea, we leave everything behind...conf files, services running. Wondering if it is sensible to preserve any existing ntp.conf (if only the one coming with the package) to say ntp.conf.deepsea_save and copy it back once ceph.time is run with ```manage: false```? Just leaving everything behind seems a bit unclean.

jan--f:
One more question:

- There doesn't seem to be a way to only configure clients? I'd assume a lot of users would have an NTP server already running and it would be helpful to point all clients to the existing server instead of creating another level of indirection.

BlaineEXE:
@jan--f, per the documentation, though I may be able to make it clearer, if you don't set a server, all are configured as clients that point to the `higher_stratum` servers, which you can change to be your in-house NTP server.

I'm still looking into your previous comments.

BlaineEXE:
> - If the systemd timesynd is running, ntp is started along-side timesyncd. I think this is very much undesirable since they both try to manage the system time. Maybe it would be feasible to stop any other time service when ntp is started? At least systemd-timsyncd and maybe even chrony if present?

There is also `openntpd`. I don't feel great about adding this, as it seems like feature creep, but I think it should be a fairly easy add. I'll start poking at this.

> - Starting to use DeepSeas ntp formular clobbers an existing ntp.conf. When one wants to go back to not have their time managed by DeepSea, we leave everything behind...conf files, services running. Wondering if it is sensible to preserve any existing ntp.conf (if only the one coming with the package) to say ntp.conf.deepsea_save and copy it back once ceph.time is run with ```manage: false```? Just leaving everything behind seems a bit unclean.

There is this option baked into Salt that can provide backups; I'll try this out to see how it works and if it'll work for our case. I suspect it may create a backup each time `ceph.time` is applied, and we may see backup bloat over time. 
https://docs.saltstack.com/en/latest/ref/states/backup_mode.html#file-state-backups

BlaineEXE:
One of our primary users of this feature had some specs to share regarding this:


> Regarding the NTP configuration I believe we need to cover two scenarios:
> 
> 1. user does NOT have an NTP infrastructure
> 2. user does have an NTP infrastructure
> 
> In case of 1. we have to
> - setup three NTP servers (maybe on the three MONs)
> - make sure all three are peer to each other
> - make sure "burst and iburst" are used between them
> - set "tinker panic 0"
> - adjust restrictions are needed and useful (I always remove them all when I am internal)
> - make sure that all three NTP servers have three "external" NTP sources configured as "server" with iburst (no burst)
> - all three NTP servers must have DIFFERENT external NTP sources
> - configure all other SES servers to retreive their time from these three NTP servers that are hosted > on the MONs (server with burst and iburst)
> 
> In case of 2. we have to
> - configure the three existing NTP servers that the user already has in ntp.conf on all SES servers with burst and iburst
> 
> In 1&2 - configure /etc/sysconfig/ntp to
> NTPD_FORCE_SYNC_ON_STARTUP="yes"
> NTPD_FORCE_SYNC_HWCLOCK_ON_STARTUP="yes"

^ As a note, we already talked about these being SUSE-specific options

> (cont'd)
> 
> Oh - and we have to ensure that the timezone and "localtime / utc" settings are identical and correct on all servers.

jan--f:
Marking Do Not Merge until we pick this up again.

swiftgist:
We have gone with a simple ntp client setup.  When the demand for a sophisticated time setup arises, we can bring this forward then.

PR #217, opened 2017-04-21
Rescind: remove non-required gold and silver directory

supriti:


supriti:
#217

PR #230, opened 2017-05-04
Added docs to status, deploy, and undeploy functions

rjfd:

### DIFF
--- a/srv/modules/runners/ui_iscsi.py
+++ b/srv/modules/runners/ui_iscsi.py
@@ -247,6 +247,12 @@ def images(**kwargs):
 
 
 def status():
+    """
+    Status of iSCSI gateways deployment
+
+    Returns:
+        bool: True if all iSCSI gateways are deployed, False otherwise.
+    """
     local = salt.client.LocalClient()
     status = local.cmd('I@roles:igw', 'service.status', ['lrbd'], expr_form='compound')
     result = True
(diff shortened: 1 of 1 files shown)


swiftgist:
@rjfd Has service.status worked for you?  There's a condition in which service.status did not behave.  It's the reason I made the runner cephprocesses instead of cephservices.  With respect to the undeploy/service.stop, wouldn't you want to stop individual gateways? Also, undeploy makes me think the service will be removed.

rjfd:
@swiftgist yes `service.status` has been working for me. What is the condition in which service.status is not behaving for you?

This PR only adds documentation to functions that are already in the master branch. I'm going to close this PR and open it against master.

BlaineEXE:
I agree with @swiftgist that the word "undeploy" isn't particularly telling of what is intended to happen to the system? I think a bit more clarity should be added. What actually happens during a (un-)deployment, e.g., package install, service enable, service start?

rjfd:
@swiftgist @BlaineEXE I agree that `undeploy/deploy` was a bad choice of names, but this should have been caught in the previous PR #223 where I actually added these functions.
Also, I'm already using these function in openATTIC.

I can create a new PR to change these names, and also change in openATTIC PRs. Do you have a good suggestion?

Just a bit of context: the purpose of these functions is to start and stop the iSCSI targets that are configured from openATTIC UI.

swiftgist:
We can carry the conversation in another issue or list.  I did not look as closely at PR #223 since I knew you were adding functionality for oA.

PR #231, opened 2017-05-04
Adding status report generation runner

### LINKED ISSUE #220: upgrade summary report
The rolling upgrade updates the OS, Salt and Ceph.  A summary report (could work as a separate runner/utility) seems almost necessary to save the admin from doing the inspection.  A summary of the OS, Salt and Ceph versions would be enough with a list of minions associated with each.  It's probably fine to exclude the longest list of minions.  Something similar to 

Common versions:
OS: SLE12SP3
Salt: 2016.11.3
Ceph: 12.0

Minions client1, client2 have:
OS: SLE12SP2
Salt: 2015.8.12
Ceph: 10.2

Should there be any unexpected versions or minions in the list, the admin can investigate.

jschmid1:
resolves #220


BlaineEXE:
I think we should have this referenced in an sls file somewhere too, otherwise the code is dormant.

jschmid1:
the initial plan was, as described in the issue, to end the maintenance.upgrade process with a summary of versions. salt being salt (or me being not experienced enough with salt), when a runner is executed in either a state or orchestrate context, it'll not be rendered nicely. That means you won't be presented with that nice format of

```yaml
node:
  - os_version: x
  - cephversion: x
  - salt_version: x
```

but with a unformated stream of json(or py<dict>), that is pretty hard to handle without piping it into another runner that handles that.
For now, my plan was to drop a message in the docs, asking the user to call:

`salt-run status.report`

and compare the desired versions with the output.

swiftgist:
I gave it a try.  I will say that needing to run

```
salt-run status.summary
```
threw me for a bit.  I kept running `salt-run status.report` and tried to figure out what was wrong with the salt-master. :)

I will say I found the out of sync message distracting when empty and not helpful enough when it listed nodes.  Here's the examples:

```
statusreport:
    |_
      ----------
      out of sync:
    |_
      ----------
      ceph:
          ceph version 12.0.0-2445-g4e47e9f (4e47e9fe42b61d61e3ae13f5cfb75b17b9b32f7c)
      os:
          SUSE Linux Enterprise Server 12 SP3
      salt:
          2016.11.3
```

When I saw the out of sync like the above, I didn't know what it meant.

```
# salt-run status.summary
statusreport:
    |_
      ----------
      out of sync:
          - mds2.ceph
          - mds1.ceph
    |_
      ----------
      ceph:
          ceph version 12.0.0-2445-g4e47e9f (4e47e9fe42b61d61e3ae13f5cfb75b17b9b32f7c)
      os:
          SUSE Linux Enterprise Server 12 SP3
      salt:
          2016.11.3
```
and when I got this message on my other cluster, I still had to track down what was wrong.  I did it manually until I remembered to run the longreport.  

```
mds1.ceph:
    ----------
    ceph_version:
          File "/usr/bin/ceph", line 106, in <module>
            import rados
        ImportError: libceph-common.so.0: cannot map zero-fill pages
    os_version:
        SUSE Linux Enterprise Server 12 SP3
    salt_version:
        2016.11.3
mds2.ceph:
    ----------
    ceph_version:
          File "/usr/bin/ceph", line 106, in <module>
            import rados
        ImportError: libceph-common.so.0: cannot map zero-fill pages
    os_version:
        SUSE Linux Enterprise Server 12 SP3
    salt_version:
        2016.11.3
```
I think the report would have been more to the point if both were combined.  That is,

```
# salt-run status.summary
statusreport:
    |_
common:
      ----------
      ceph:
          ceph version 12.0.0-2445-g4e47e9f (4e47e9fe42b61d61e3ae13f5cfb75b17b9b32f7c)
      os:
          SUSE Linux Enterprise Server 12 SP3
      salt:
          2016.11.3

mds1.ceph:
    ----------
    ceph_version:
          File "/usr/bin/ceph", line 106, in <module>
            import rados
        ImportError: libceph-common.so.0: cannot map zero-fill pages
    os_version:
        SUSE Linux Enterprise Server 12 SP3
    salt_version:
        2016.11.3

mds2.ceph:
    ----------
    ceph_version:
          File "/usr/bin/ceph", line 106, in <module>
            import rados
        ImportError: libceph-common.so.0: cannot map zero-fill pages
    os_version:
        SUSE Linux Enterprise Server 12 SP3
    salt_version:
        2016.11.3
```
then I would realize that all minions are on the same version and that my mds servers are likely having memory issues all from one command.

jschmid1:
clearly, a better documentation would've been helpful here.

> and when I got this message on my other cluster, I still had to track down what was wrong. I did it manually until I remembered to run the longreport.

The workflow that I imagined works like this:

You want to see if your nodes are on the same version level.
(Thinking again, reminded me that I need to add a exclude filter, for e.g. client nodes that are not part of the checks because running on a different environment.)

`salt-run status.summary`

which will reach out to all the nodes in (for now I@cluster:ceph) and compute the most common versions. Nodes that don't match any of the 'Common Version' will be displayed in the 'out-of-sync' area. Which gives you an idea that something is probably wrong with those machines.

To get a more detailed report you will run:

(here again, adding targeting for e.g. roles would be helpful to strip down the results)
`salt-run status.longreport`

That will list all connected nodes with their versions. You will go through the list, find the correct node, fix it and rerun `salt-run status.summary`. This time it will be 'in-sync' and the error message will go away.


> I will say I found the out of sync message distracting when empty and not helpful enough when it listed nodes.

If all nodes are on the same patch-level, the 'out-of-sync' area will be empty. Which gives me the impression that everything is okay. _Naming can still be discussed_


> I think the report would have been more to the point if both were combined. 

Putting the summary on the bottom will safe you one execution.

So 2 options here:

1) Combine 'longreport' and 'summary'
2) Create a 3rd 'combined' method

swiftgist:
> clearly, a better documentation would've been helpful here.

I think less is more in this case.  There's not really a need to have the admin push two buttons when one will do.

The two options appear to be the same to each other unless you meant literally calling summary followed by longreport.  I think I would prefer

salt-run status.report

creates the summary followed by the specific minions that did not match and their current values.

swiftgist:
submitted under #281

PR #232, opened 2017-05-04
Improve openATTIC orchestration

votdev:
* Make openATTIC configurable via policy.cfg.
* Use module.run instead of cmd.run to let Salt choose which init system to use to restart services.
* Remove packages openattic openattic-base openattic-pgsql (stage 5)
* If configured, then purge openattic postgres database during uninstall (stage 5)
* Add support for custom pillar data via /src/pillar/openattic.sls to customize oA orchestration.


swiftgist:
Works fine for me.  Tested install, uninstall, reinstall on master node.  Tested install, uninstall on another node.

Created two issues #237 and #238 to be addressed later.

Issue #239, opened 2017-05-08
Should we s/ceph.cherrypy/ceph.salt-api/ ?


tserong:
I'm having a Tim-is-an-annoying-pedant moment. Why is ceph.cherrypy named ceph.cherrypy, rather than ceph.salt-api? I mean, it's there to install salt-api, which just happens to be using cherrypy internally...

rjfd:
@tserong I totally agree with your observation. Please open a PR to change that.

PR #241, opened 2017-05-08
Rename ceph.cherrypy to ceph.salt-api

tserong:
Fixes: https://github.com/SUSE/DeepSea/issues/239


LenzGr:
LGTM, thanks. Does this have any impact on applications that actually consume the REST API, or is this a pure cosmetic packaging change?

tserong:
Consumers of the API won't see any change. It's just that previously if you were setting this up by invoking `salt '<MYMASTER>' state.apply ceph.cherrypy`, you'd now run `salt '<MYMASTER>' state.apply ceph.salt-api`.

PR #246, opened 2017-05-09
stage 0: fix mines update regression

rjfd:
The commit 20b528c introduces a regression on the salt mines,
which prevents the mines from being populated because the packages
necessary for the mine functions have not been installed yet
at that phase.


swiftgist:
I was remembering that we moved mines before.  It needs to be after the packages but before any reboots (which this is).

PR #247, opened 2017-05-09
Wip ui rgw

swiftgist:
Credentials api for RadosGW

### DIFF
--- a/srv/modules/runners/ui_rgw.py
+++ b/srv/modules/runners/ui_rgw.py
@@ -29,12 +29,14 @@ class Radosgw(object):
         Initialize and call routines
         """
         if canned:
-            self._canned(canned)
+            self._canned(int(canned))
         else:
             self.cluster = cluster
             self.credentials = {'access_key': None,
                                 'secret_key': None,
-                                'urls': []}
+                                'urls': [],
+                                'success': False}
+
             self.pathname = pathname
             self._admin()
             self._urls()
@@ -76,6 +78,7 @@ class Radosgw(object):
                 return
         self.credentials['access_key'] = user['keys'][0]['access_key']
         self.credentials['secret_key'] = user['keys'][0]['secret_key']
+        self.credentials['success'] = True
 
     def _urls(self):
         """
@@ -94,29 +97,31 @@ class Radosgw(object):
                 return
 
         for client_file in glob.glob("{}/client.*".format(self.pathname)):
-            port = None
+            port = '7480' # civetweb default port
+            ssl = ''
             with open(client_file, 'r') as rgw_file:
                 for line in rgw_file:
                     if 'port=' in line:
-                        match = re.search(r'port=(\d+)', line)
+                        match = re.search(r'port=(\d+)(s?)', line)
                         port = match.group(1)
+                        ssl = match.group(2)
             parts = client_file.split('.')
-            resource = None
+            resource = ''
             # dedicated keys - use host part
             if len(parts) == 4:
                 resource = parts[2]
             # shared keys - use role part
             if len(parts) == 3:
                 resource = parts[1]
-            if port:
+            if resource and port:
                 resource += ":{}".format(port)
             if resource:
-                self.credentials['urls'].append("http://{}".format(resource))
+                self.credentials['urls'].append("http{}://{}".format(ssl, resource))
 
 
 def credentials(canned=None):
     """
     Return the administrative credentials for the RadosGW
     """
-    radosgw = Radosgw(int(canned))
+    radosgw = Radosgw(canned)
     return radosgw.credentials
--- a/srv/modules/runners/ui_rgw.py
+++ b/srv/modules/runners/ui_rgw.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+# -*- coding: utf-8 -*-
 
 """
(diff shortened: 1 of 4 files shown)


p-na:
Will try that out

srv/modules/runners/ui_rgw.py
```diff
@@ -0,0 +1,122 @@
...
+            user = None
+            for user_file in glob.glob("{}/user.*".format(self.pathname)):
+                user = json.loads(open(user_file).read())
+                if 'system' in user and user['system'] == "true":
+                    break
+                user = None
+            if not user:
+                # No system user
+                log.error("No system user for radosgw found")
+                return
```
rjfd: I think we should add another entry to the `credentials` dictionary to represent the success or failure of the operation. Because currently if the execution takes this path, oA will not be able to distinguish from a success or failure.
What do you think of adding `self.credentials['success'] = False` before the return? and also add `self.credentials['success'] = True in the `__init__` function.

theanalyst: can't we return a http status of 5XX (ideally 503) instead?

srv/modules/runners/ui_rgw.py
```diff
@@ -0,0 +1,122 @@
+# -*- coding: utf-8 -*-
+
+"""
+Consolidate any user interface rgw calls for Wolffish and openATTIC.
```
theanalyst: fix the description while we're at it

srv/modules/runners/ui_rgw.py
```diff
@@ -0,0 +1,122 @@
...
+            if 'rgw_endpoint' in cached[minion]:
+                self.credentials['urls'].append(cached[minion]['rgw_endpoint'])
+                return
+
+        for client_file in glob.glob("{}/client.*".format(self.pathname)):
+            port = None
+            with open(client_file, 'r') as rgw_file:
+                for line in rgw_file:
+                    if 'port=' in line:
+                        match = re.search(r'port=(\d+)', line)
```
theanalyst: this should cover the common use case for http, for https this will be say `443s` or some decimal succeeding s , also there is the case of port not appearing in the config, which defaults to civetweb port=7480 by ref http://docs.ceph.com/docs/master/install/install-ceph-gateway/#using-ssl-with-civetweb

srv/modules/runners/ui_rgw.py
```diff
@@ -0,0 +1,122 @@
...
+            # dedicated keys - use host part
+            if len(parts) == 4:
+                resource = parts[2]
+            # shared keys - use role part
+            if len(parts) == 3:
+                resource = parts[1]
+            if port:
+                resource += ":{}".format(port)
+            if resource:
+                self.credentials['urls'].append("http://{}".format(resource))
```
theanalyst: maybe https depending on the port containing the `s` string

theanalyst:
There is one usecase which we can't purely determine from Deepsea, and this would be when there are multiple RGWs fronted by a LB/reverse proxy of some sort in which case only the visible endpoint for an RGW would be whatever is configured in the LB and this settings do not normally appear in ceph.conf

srv/modules/runners/ui_rgw.py
swiftgist: Hmm, I don't know how temporary a failure would be.  If no system user exists at this point, it's unlikely it will exist.

Adding 'success' is easy enough but what did you want to base it on?  That a system user is found with an access and secret keys?  Or that the urls aren't empty?

srv/modules/runners/ui_rgw.py
swiftgist: Do you mean delete that line?

srv/modules/runners/ui_rgw.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def credentials(canned=None):
+    """
+    Return the administrative credentials for the RadosGW
+    """
+    radosgw = Radosgw(int(canned))
```
kmroz: Heads up that int() will throw a TypeError if canned==None.

theanalyst: good catch, just pass canned through, do a try catch in the init maybe?

srv/modules/runners/ui_rgw.py
rjfd: I don't know much about RGW credentials, so just return `success = True` if the contents of the credentials variable are OK to be used, and `success = False` otherwise.

@theanalyst I don't think it's possible from the runner to influence the HTTP return code of the salt-api.

theanalyst: ack

srv/modules/runners/ui_rgw.py
kmroz: A try/catch would be good idea... here or in Radosgw:__init__().  What about changing the default of 'canned' from None to 0 across the patch as well?

theanalyst: yeah, but that'll still not handle say when someone passes canned(3) which will just return without setting anything. and I guess we could do a more stricter validation in the init

kmroz: Good point. The eyes didnt register that canned is a selection mechanism, not an on/off switch. :facepalm:

swiftgist:
I believe I addressed all the changes requested.  Rebased as well.

PR #252, opened 2017-05-12
[wip] Single node update 

### LINKED ISSUE #224: Update a single node
Can we create an option so that a single node is selected for full upgrade? Instead of rolling the entire cluster, we'd like to specify a single node to verify the SP upgrade will do well.

jschmid1:
closes #224

I'm still puzzled on salt's behavior.

when 'single_node_upgrade' is set:

`salt '*' state.show_sls ceph.maintenance.upgrade.minion`

I get the expected rendered output with only one node.(I also get all nodes when 'single_node_upgrade' is unset).

When actually executed, though, it iterates through _all_ nodes.. 

any thoughts?

### DIFF
--- a/srv/modules/runners/orderednodes.py
+++ b/srv/modules/runners/orderednodes.py
@@ -16,7 +16,7 @@ def _preserve_order_sorted(seq):
     return [x for x in seq if x not in seen and not seen.add(x)]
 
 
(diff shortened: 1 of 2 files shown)


jschmid1:
Closed as the issue was closed

Issue #205, opened 2017-04-13
Easier inclusion of custom statements into ceph.conf


l-mb:
The default ceph.conf.j2 only includes files for configured RGW instances.

A naïve user might modify the file directly to adjust/include additional settings, but that'd be overwritten by the next package update. (And also confuse rpm -V.)

A more savvy user might instead add a custom file to configuration_init: in global.yml, using our ceph.conf.j2 as a template. However, that'd mean they'd have to manually merge any further updates we provide to that file.

A really savvy user would instead set up a custom file, but include ours via a j2 macro directive, which is not part of our default template though.

I don't think this is user-friendly enough and most likely, users will get this wrong. Users also should not have to any macro code to just include a single "a = b" statement.

We need to make it easier to include custom snippets - simply including all files under ceph.conf.d/ in numbered sequence (similar to udev.d etc) would do it. The ceph.conf.j2 file should include a comment to document that it is read-only and should not be changed by customers. (Maybe even ship it 0400 only.)

swiftgist:
I've been wanting to do this for a while, but wasn't sure how to handle customizations to subsections.  But something is definitely needed.

Having a /srv/salt/ceph/configuration/files/ceph.conf.d should work well and seem obvious enough.  I think we can rely on a Jinja include here since it's more forgiving about absent files.  

Not sure about extrapolating this to all other configuration files, but I think I favor this strategy over creating more entries in the Salt pillar.

PR #256, opened 2017-05-16
configuration: add include for [global,osd,mon,mds,client].conf

jan--f:
This should provide an easy way to to add custom configuration options
to ceph.conf. Each file can contain global configuration (for the cluster
of a daemon type) and per daemon configuration. Note that per daemon
configuration must be preceded by the daemon name, e.g. [osd.1].
This addresses #205.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,7 @@ copy-files:
 	install -m 644 srv/salt/ceph/configuration/check/*.sls $(DESTDIR)/srv/salt/ceph/configuration/check/
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/configuration/files
 	install -m 644 srv/salt/ceph/configuration/files/ceph.conf* $(DESTDIR)/srv/salt/ceph/configuration/files/
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/configuration/files/ceph.conf.d
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/events
 	install -m 644 srv/salt/ceph/events/*.sls $(DESTDIR)/srv/salt/ceph/events/
 	# state files - diagnose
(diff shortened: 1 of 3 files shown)


jan--f:
This can even overwrite the default config since ceph seems to respect the last occurence of a particular config line. So cephx auth could be turned off by add the respective lines to ceph.conf.d/global.conf.

jan--f:
Uncertain where docs for this should go. Could be in otherwise empty files in ```srv/salt/ceph/configuration/files/ceph.conf.d``` or somewhere else.

swiftgist:
What do you think of 

{% for file in [ salt['cmd.run']('ls /srv/salt/ceph/configuration/files/ceph.conf.d/*.conf') ] %}
{% include file ignore missing %}
{% endfor %}

My thought is that the admin could pick any naming convention.  As far as docs, I think a README in .../ceph.conf.d would work.

jan--f:
> What do you think of
> 
> {% for file in [ salt['cmd.run']('ls /srv/salt/ceph/configuration/files/ceph.conf.d/*.conf') ] %}
> {% include file ignore missing %}
> {% endfor %} 

I thought of this too. The problem I found with this though is that the config lines need to go into the correct stanza. E.g. auth settings need to be in the [global] stanza. Also (at least from my limited experiments) there can only be exactly one, say [global], stanza. If one adds another one its simply ignored.

All this would make looping over and including all conf files rather messy I think. Users had to pay attention to the order of their configuration fragments or deal with strange errors where ceph ignores parts of their fragments.

The current approach would accept global options in global.conf. The other files are for daemon-global configurations and the user could even add daemon instance specific config by adding, say [osd.1] and then listing some config options. DeepSea can then assemble the config file in the correct order. This seems less flexible but is imho clearer and has less potential for surprising results.

jschmid1:
I think I'm in favor of following the daemon-global subsection approach you initially came up with.
Giving naming freedom is generally a nice thing, but *will* lead to unintentionally overwriting/shadowing of configuration and end up in a debug hell.. 

I really think that having one file for each daemon is comprehensible and easy to understand.

swiftgist:
I'm fine with it.  I talked to Jan and think the loop I created could be added to the README for any admin that needs an example on how to split out the configuration file if that's really desired.  That should be enough.

PR #260, opened 2017-05-16
ntp client

swiftgist:
Configure the ntp client.


swiftgist:
@jan--f That's one directory up.  .../ceph/time/init.sls is time_init, .../ceph/time/ntp/init.sls is time_ntp.

jan--f:
right...my bad

PR #264, opened 2017-05-18
ntp: detect if ntp is running and if so do nothing

jan--f:
When ntp is running, do not overwrite ntp.conf since we can assume that
ntp is all setup. This should be a better default then blindly blowing
away a running ntp setup. The old behaviour can still be enabled by
setting 'time_init: ntp'.


jan--f:
> ("somewhere?") 

That is the deciding factor here. Yes we should and I will.

Issue #243, opened 2017-05-08
Deploy and configure Prometheus node exporters on every node


LenzGr:
It would be nice if DeepSea would allow me to set up a functional monitoring infrastructure, by setting up a central monitoring node running [Prometheus](https://prometheus.io/) (and [Grafana](http://grafana.org/)?) and then deploy and configure the Prometheus [node_exporter](https://github.com/prometheus/node_exporter) on every Ceph cluster node, to gather some basic system statistics.

There's already a Salt Formula related to Prometheus - [saltstack-prometheus-formula](https://github.com/bechtoldt/saltstack-prometheus-formula) - maybe it can be used here?

tserong:
I had a quick look at that salt formula, and most of it seems to be extracting prometheus binaries from tarballs and creating systemd unit files...

swiftgist:
On Tuesday, May 09, 2017 03:19:13 AM Tim Serong wrote:
> I had a quick look at that salt formula, and most of it seems to be
> extracting prometheus binaries from tarballs and creating systemd unit
> files...

I haven't looked at your packages, but I would guess that extracting unit files 
isn't really needed.

tserong:
> I haven't looked at your packages, but I would guess that extracting unit files isn't really needed.

Yeah, I reckon not. My latest golang-github-prometheus-prometheus, golang-github-prometheus-node_exporter and golang-github-digitalocean-ceph_exporter and grafana RPMs can all be found in https://build.opensuse.org/project/show/home:tserong for the moment if anyone wants to play.

jan--f:
For now we will use the ceph_node_exporter. This doesn't need to be installed on all nodes afaik. It runs as a client to ceph (and needs a keyring) so any node will suffice.
Any opposition to simply putting it on the salt master? This node has an admin keyring anyway and I can't think of a reason against it of the top of my head.

LenzGr:
@jan--f no objections against running the ceph_node_exporter on the Salt master by default. Yes, the regular node exporter needs to be installed on every node :)

swiftgist:
PR Merged for exporters

PR #265, opened 2017-05-18
Install and deploy prometheus exporters

jan--f:
This adds support for installation and deployment of the prometheus exporters. Assumption is that the repos offer the requested packages. The ceph_exporter is installed on the salt_master (since it needs an admin keyring), the node_exporter is installed on all nodes that are targeted.

This is not yet integrated into any particular stage, but it's easy enough to do.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,6 +155,11 @@ copy-files:
 	install -m 644 srv/salt/ceph/mon/files/*.j2 $(DESTDIR)/srv/salt/ceph/mon/files/
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/mon/restart
 	install -m 644 srv/salt/ceph/mon/restart/*.sls $(DESTDIR)/srv/salt/ceph/mon/restart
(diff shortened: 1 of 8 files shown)


tserong:
This probably also wants `service.enabled` somehow for both exporters so they'll be running after a reboot (unless that's already handled by magic). Otherwise LGTM.

l-mb:
Should we install ceph_exporter on the ceph-mgr node(s) instead?

jan--f:
@tserong of course...thx...fixed.

jan--f:
@l-mb the current ceph exporter is this one here: https://github.com/digitalocean/ceph_exporter
So atm it doesn't have anything to do with mgr. That being said this will probably get more complex once we install and configure prometheus and grafana too.

jan--f:
Fixes #243

tserong:
I don't think installing ceph_exporter on ceph-mgr nodes will buy us anything. The exporter is just a client, same as any other client.

PR #267, opened 2017-05-18
Fix ordering of Storage node removal

swiftgist:
I thought I had cherry-picked this from the SES4 branch in the past, but the logic is still broken in main.  I verified that removal works properly against master with these changes.


swiftgist:
Works for me with the latest changes.

jan--f:
@swiftgist yes, same here. Ready to merge from my perspective.

PR #271, opened 2017-05-19
check human editable yaml files for syntax

jschmid1:
We had a user issue where the syntax of a osd file was slightly messed up. The parsing exception was swallowed by salt and as a result, the user got irritating results.

avoid that by preparsing

* any file that the policy.cfg points to
* any file under /srv/pillar/stack

as an additional check I added

* check if you have any files under the profiles directory (or even a directory)

and fails at stage 2


swiftgist:
Works for me... added #272

PR #275, opened 2017-05-22
Wip osd

swiftgist:
Backward compatible osd functionality, supports ceph namespace data structure, allows creation/deletion (i.e. Stage 5) of bluestore.  Initial support for bluestore + dmcrypt.

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -452,7 +452,7 @@ class OSDConfig(object):
         if self.journal:
             disks = __salt__['mine.get'](tgt=__grains__['id'], fun='cephdisks.list')
             if disks:
-                for disk in self.disks[__grains__['id']]:
+                for disk in disks[__grains__['id']]:
                     # Check size of journal disk
                     if disk['Device File'] == self.journal:
                         if int(disk['Bytes']) < 10000000000000: # 10GB
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -8,6 +8,8 @@ import logging
 import time
 import re
 import pprint
+import yaml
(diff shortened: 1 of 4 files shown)


srv/salt/_modules/osd.py
```diff
@@ -253,3 +292,902 @@ def restore_weight(id, **kwargs):
...
+        if self._config_version() == OSDConfig.V2:
+            return self._check_existence('journal_size', self.tli, self.device, default=self._journal_default())
+
+    def _journal_default(self):
+        """
+        """
+        if self.journal:
+            disks = __salt__['mine.get'](tgt=__grains__['id'], fun='cephdisks.list')
+            if disks:
+                for disk in self.disks[__grains__['id']]:
```
jan--f: This line contains an error. Should be
```disk in disks[__grains__['id']]:```
Want me to push a fix?

jan--f:
minor nit for usablilty: if one runs the osd module on a mon node like so:
```salt mon1 osd.is_prepared /dev/vdb``` a backtrace is returned.

Not sure if we can do something sensible about this...just cam across it.

rjfd:
@swiftgist is it possible to provide a few instructions on how to use this new PR?

swiftgist:
@rjfd I will, but currently this is backward compatible.  Since I need to package for the next milestone, I planned to send the instructions to a few mailing lists.  The ultimate goal is to use Jan's custom profile runner.

Take a look at the end of #62 to see the latest examples that I used.  Jan verified that these worked for him.

I have another PR incoming to deal with the removal.  That will reduce the instructions necessary for Stage 5 using dmcrypt + bluestore.

rjfd:
@swiftgist great! thanks

PR #277, opened 2017-05-23
Fix missing ganesha.service file in installation and RPM generation

rjfd:


supriti:
LGTM

PR #278, opened 2017-05-24
ui_rgw: fixed bugs in ui_rgw runner

rjfd:
The use of the string `'__opts__'` was triggering an exception inside salt code, and `**kwargs` was missing in the `credentials` function, which is required to call the runner from salt-api.

### DIFF
--- a/srv/modules/runners/ui_rgw.py
+++ b/srv/modules/runners/ui_rgw.py
@@ -86,10 +86,11 @@ class Radosgw(object):
         urls.
         """
         search = "I@cluster:{}".format(self.cluster)
+        __opts__ = salt.config.client_config('/etc/salt/master')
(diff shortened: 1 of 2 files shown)


srv/modules/runners/ui_rgw.py
```diff
@@ -119,7 +120,7 @@ def _urls(self):
                 self.credentials['urls'].append("http{}://{}".format(ssl, resource))
 
 
-def credentials(canned=None):
+def credentials(canned=None, **kwargs):
```
theanalyst: how are we using kwargs here?

theanalyst: It shouldn't affect anything though

rjfd: When calling the runner from salt-api rest interface, salt-api adds an additional parameter. See #255

PR #290, opened 2017-05-29
Internal vs external rep of disks

jschmid1:
resolves deployment issues on bare-metal

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -86,16 +86,8 @@ def configured(**kwargs):
         if 'format' in kwargs and kwargs['format'] != 'filestore':
             return []
     log.debug("devices: {}".format(devices))
-    for device in devices:
-        # find real device
-        cmd = "readlink -f {}".format(device)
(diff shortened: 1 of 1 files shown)


srv/salt/_modules/osd.py
```diff
@@ -322,7 +314,8 @@ def __init__(self, device, **kwargs):
         filters = kwargs.get('filters', None)
         # top_level_identifiier
         self.tli = self._set_tli()
-        self.device = device
+        self.device = self.set_device(device)
+        self.by_id_path = device
```
jan--f: nit: remove trailing whitespace here

srv/salt/_modules/osd.py
```diff
@@ -741,7 +746,9 @@ def osd_partition(self):
         TODO: dmcrypt
         """
         device = self.osd.device
-        if 'osds' in self.settings and device in self.settings['osds']:
+        internal_ident = self.osd.by_id_path
```
jan--f: more trailing whitespaces.

PR #292, opened 2017-05-30
is_prepared and is_activated needs correct device identifier

jschmid1:
bare-metal returns by-id disk path while virtualized env return only the short path.

while the previous fixes for the disk identifier solved the initial problem, this is a followup fix for is_prepared() and is_activated()

fix: don't use the raw device as it comes from osd.configured() but rather the converted/sanitized attribute

part2)

data+journals struct wasn't correctly accessed hence not parsed.

### DIFF
--- a/srv/salt/_modules/cephdisks.py
+++ b/srv/salt/_modules/cephdisks.py
@@ -319,12 +319,19 @@ class HardwareDetections(object):
         # TODO: Search for all possible codes:
         data = "Partition GUID code: 45B0969E-9B03-4F30-B4C6-B4B80CEFF106"
         journal = "Partition GUID code: 4FBD7E29-9D25-41B8-AFD0-062C0CEFF05D"
+        db = "Partition GUID code: 30CD0809-C2B2-499C-8879-2D6B78529876"
+        wal = "Partition GUID code: 5CE17FCE-4087-4169-B7FF-056CC58473F9"
+        osd_lockbox = "Partition GUID code: FB3AABF9-D25F-47CC-BF5E-721D1816496B"
         sgdisk_path = self._which('sgdisk')
         for partition_id in ids:
             cmd = "{} -i {} {}".format(sgdisk_path, partition_id, device)
             proc = Popen(cmd, stdout=PIPE, stderr=PIPE, shell=True)
             for line in proc.stdout:
-                if (line.startswith(data) or line.startswith(journal)):
+                if line.startswith(data) or \
+                   line.startswith(journal) or \
+                   line.startswith(db) or \
+                   line.startswith(wal) or \
+                   line.startswith(osd_lockbox):
                     return True
             for line in proc.stderr:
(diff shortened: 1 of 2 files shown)


jschmid1:
WIP: Do not merge yet

srv/salt/_modules/cephdisks.py
```diff
@@ -319,12 +319,19 @@ def _osd(self, device, ids):
         # TODO: Search for all possible codes:
         data = "Partition GUID code: 45B0969E-9B03-4F30-B4C6-B4B80CEFF106"
         journal = "Partition GUID code: 4FBD7E29-9D25-41B8-AFD0-062C0CEFF05D"
+        db = "Partition GUID code: 30CD0809-C2B2-499C-8879-2D6B78529876"
+        wal = "Partition GUID code: 5CE17FCE-4087-4169-B7FF-056CC58473F9"
+        osd_lockbox = "Partition GUID code: FB3AABF9-D25F-47CC-BF5E-721D1816496B"
         sgdisk_path = self._which('sgdisk')
```
jschmid1: verify with:
https://github.com/SUSE/ceph/blob/master/src/ceph-disk/ceph_disk/main.py#L49

srv/salt/_modules/osd.py
```diff
@@ -76,15 +76,21 @@ def configured(**kwargs):
     """
     osds = []
     devices = []
+    # That doesn't allow mixed configurations
+    # storage[osds] OR storage[data+journals]
```
jan--f: This can now be removed?

srv/salt/_modules/osd.py
```diff
@@ -76,15 +76,21 @@ def configured(**kwargs):
...
         and 'osds' in __pillar__['ceph']['storage']):
         devices = __pillar__['ceph']['storage']['osds']
         devices = _filter_devices(devices, **kwargs)
+        devices = devices.keys()
     if 'storage' in __pillar__ and 'osds' in __pillar__['storage']:
         devices = __pillar__['storage']['osds']
         log.debug("devices: {}".format(devices))
         if 'format' in kwargs and kwargs['format'] != 'filestore':
             return []
+    if 'storage' in __pillar__ and 'data+journals' in __pillar__['storage']:
```
jan--f: I think this if clause would be better as a two level if. Like:
```
if 'storage' in __pillar__ :
    if 'osds' in __pillar__['storage']:
        ...
    if 'data+journals' in __pillar__['storage']:
        ...
```

srv/salt/_modules/osd.py
jschmid1: I'd like to leave it here as a marker for a future issue, 
will assign it an issue though

srv/salt/_modules/osd.py
jschmid1: will be fixed in a followup PR

PR #301, opened 2017-05-31
Terminate step is separate now, call directly

swiftgist:


jschmid1:
The CI failed for internal reasons, not a blocker.

jan--f:
@jschmid1 Yup they definitely stay behind. ```ceph-osd@${removed_id}.service```
```ceph-disk deactivate PATH``` is taking care of that outside of DeepSea.

swiftgist:
In wip-osd-remove, the remove logic calls "systemctl disable ceph-osd@#" which removes the symlink.  

If I call terminate directly, I may not want the OSD disabled.

Issue #294, opened 2017-05-30
validate.discovery should respect DEV_ENV=true


jan--f:
want to deploy a cluster without OSDs and can't get past stage.2.

swiftgist:
Is this related to #268?  Guaranteeing at least one storage node is present with DEV_ENV?

PR #307, opened 2017-06-02
Added the DEV_ENV check to the monitor validation

capri1989:
Now it's possible to deploy less than 3 MON setups for testing/development

Signed-off-by: Kai Wagner <kwagner@suse.com>

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -260,7 +260,7 @@ class Validate(object):
                 'mon' in self.data[node]['roles']):
                 monitors.append(node)
 
-        if len(monitors) < 3:
+        if (not self.in_dev_env and len(monitors) < 3) or (self.in_dev_env and len(monitors) < 1):
             msg = "Too few monitors {}".format(",".join(monitors))
             self.errors['monitors'] = [ msg ]
         else:
@@ -413,8 +413,11 @@ class Validate(object):
             self.errors.setdefault(name, []).append(msg)
         elif same_hosts:
             count = len(same_hosts.keys()[0].split(","))
-            if count < 3:
-                msg = "Must have at least three entries"
+            if (not self.in_dev_env and count < 3) or (self.in_dev_env and count < 1):
+                if self.in_dev_env:
+                    msg = "Must have at least one monitor"
+                else:
+                    msg = "Must have at least three monitors"
                 self.errors[name] = [ msg ]
         else:
             msg = "Missing {}".format(name)


swiftgist:
This is inline with #268.  I think we can visit #294 when we all have a moment.

capri1989:
I can change my PR and set the requirements for OSD to "zero" to fix both issues at the same time. What do you think?

srv/modules/runners/validate.py
```diff
@@ -260,7 +260,7 @@ def monitors(self):
                 'mon' in self.data[node]['roles']):
                 monitors.append(node)
 
-        if len(monitors) < 3:
+	if (not self.in_dev_env and len(monitors) < 3) or (self.in_dev_env and len(monitors) < 1):
```
jan--f: Please convert the leading tab to spaces. Python does not like mixing the too.

srv/modules/runners/validate.py
```diff
@@ -413,7 +413,7 @@ def _monitor_check(self, name):
             self.errors.setdefault(name, []).append(msg)
         elif same_hosts:
             count = len(same_hosts.keys()[0].split(","))
-            if count < 3:
+	    if (not self.in_dev_env and count < 3) or (self.in_dev_env and count < 1):
```
jan--f: Again /s/tab/space/ and please remove trailing whitespaces.

srv/modules/runners/validate.py
```diff
@@ -413,7 +413,7 @@ def _monitor_check(self, name):
             self.errors.setdefault(name, []).append(msg)
         elif same_hosts:
             count = len(same_hosts.keys()[0].split(","))
-            if count < 3:
+	    if (not self.in_dev_env and count < 3) or (self.in_dev_env and count < 1):        
                 msg = "Must have at least three entries"
```
jan--f: The error message should probably change accordingly. Maybe
```
if self.in_dev_env:
    msg = "Must have at least one monitor"
else:
    msg = "Must have at least three monitors"
```

srv/modules/runners/validate.py
```diff
@@ -413,9 +413,13 @@ def _monitor_check(self, name):
             self.errors.setdefault(name, []).append(msg)
         elif same_hosts:
             count = len(same_hosts.keys()[0].split(","))
-            if count < 3:
-                msg = "Must have at least three entries"
-                self.errors[name] = [ msg ]
+            if (not self.in_dev_env and count < 3) or (self.in_dev_env and count < 1):
+                if self.in_dev_env:
+                    msg = "Must have at least one monitor"
+                    self.errors[name] = [ msg ]
```
rjfd: Move this line to after the "if" block, and remove the same line in the "else" branch.

rjfd:
@capri1989 after the addressing the comments, squash all commits into a single one.

srv/modules/runners/validate.py
```diff
@@ -413,9 +413,12 @@ def _monitor_check(self, name):
...
             count = len(same_hosts.keys()[0].split(","))
-            if count < 3:
-                msg = "Must have at least three entries"
-                self.errors[name] = [ msg ]
+            if (not self.in_dev_env and count < 3) or (self.in_dev_env and count < 1):
+                if self.in_dev_env:
+                    msg = "Must have at least one monitor"
+                else:
+                    msg = "Must have at least three monitors"
+            self.errors[name] = [ msg ]
```
rjfd: I think this statement must be inside the first conditional block, otherwise you'll be adding an error message even if everything is ok

PR #319, opened 2017-06-09
runners: cephprocess avoid empty list as a default arg.

theanalyst:
Since this would bring a whole world of pain that is hard to debug as
lists tend to be mutable, unless this is the expected behaviour. We are
doing a `if not roles` which should cover the case for an empty list.
Otherwise subsequent calls would have the value of roles filled from
previous cache, which may not be the expected behaviour.


jan--f:
Closing. Feel free to reopen if needed.

PR #320, opened 2017-06-09
runners: orderednodes avoid empty list as a default argument

theanalyst:
In this case the change would be less behavioural, but still useful,
since exclude state would be maintained between the calls, and most
likely based on salt runnner's implementation this might only spawn a
single process per call so we might be in the green, but let's be
explicit and only use exclude if the caller supplies it.


jan--f:
Since we don't use append (or any other mutator for that matter) on the default parameter I don' think we will have this problem. So I'd be in favour of keep the old (simpler) code.

PR #326, opened 2017-06-12
Changed the formatting of the rolebased policy.cfg + oA role

capri1989:
Changed the formatting of the rolebased policy.cfg + oA role

Signed-off-by: Kai Wagner <kwagner@suse.com>


capri1989:
I added the original values again

PR #328, opened 2017-06-12
qa: make 1- and 2-node clusters reach HEALTH_OK

smithfarm:
For the time being we're introducing a `qa/.config` (which must be populated for each run) to tell the integration testing script how many storage nodes there are. If the number is less than three, then the necessary ceph.conf knobs are twiddled to ensure the cluster reaches HEALTH_OK.

No longer tolerate HEALTH_WARN.

Use "wait.until status=HEALTH_OK" to give the cluster some time to reach HEALTH_OK. The timeout seems to be five minutes. Don't know could that be extended, but it seems to reach HEALTH_OK pretty quick, so perhaps 5 minutes is enough.

### DIFF
--- a/qa/README
+++ b/qa/README
@@ -2,7 +2,18 @@ DeepSea integration test suite
 ==============================
 
 The DeepSea integration suite residing in qa/ consists of a number of
-executable files (bash scripts), all of which make the following assumptions:
+executable files (bash scripts), which can be classified as follows:
+
+qa/common/ -> scripts that are sourced by the test scripts
+qa/suites/ -> test scripts
+
+The test scripts are designed to be run from the qa/ directory. Please
+study carefully this README and the comments at the top of the test script
+before trying to run any of the test scripts.
+
+
+Assumptions made by all test scripts
+------------------------------------
 
 1. there are at least two test nodes (VMs or physical machines) that are
    running the same OS (e.g. Leap 42.3) and can see eachother over the network. 
@@ -16,5 +27,15 @@ executable files (bash scripts), all of which make the following assumptions:
 5. the DeepSea code under test has already been installed (git clone ; make
    install) on the master node
 6. the integration test script (this script) is run from the qa/ subdirectory
-   of the DeepSea repo clone
+   of the DeepSea repo clone.
+
+
+The following section is WIP - do not use!
+------------------------------------------
+
+7. a file called 'deepsea-qa-cluster-layout' (working name) has optionally been
+   created inside the same directory where the DeepSea repo was cloned. If it
+   exists, this file is expected to contain data points describing the cluster
+   layout (e.g. which nodes are storage nodes, which are client nodes, etc.).
+   If it is not present, the test script will try to do without it.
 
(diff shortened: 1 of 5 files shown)


smithfarm:
I have a gnawing feeling that this `qa/.config` thing is ill-conceived, or at least overkill for determining the number of storage nodes. Each script already knows whether it calls for a client node. The number of storage nodes is simply: total_number_of_nodes minus number_of_client_nodes (where the latter can be either 0 or 1).

But I have a feeling that `qa/.config` may turn out to be useful down the road, so we might as well have the infrastructure. What does anyone think?

What is the most elegant way of determining the total number of nodes/minions?

smithfarm:
How about this proposal. Some tests may be able to take advantage of multiple client nodes, yet not *require* any particular number of client nodes. That would be a use case for `qa/.config` with the following structure:

```
TOTALNODES=5
STORAGENODES=3
CLIENTNODES=2
```

smithfarm:
This passes `deepsea/basic` in teuthology, which currently consists of four tests:

* minimal HEALTH_OK test on 1 node, 2 nodes, and 3 nodes
* minimal MDS health test on 3 nodes

rjfd:
@smithfarm why do we need the the contents of the `.config` file? Couldn't we find out automatically the resources available?

smithfarm:
@rjfd I'm not sure we need it. The immediate problem it is solving is this: if there are less than three storage nodes, I have to apply special settings to `ceph.conf`. Assuming that the custom configuration settings are only applied in Stage 3, after Stage 2 I could query the roles to find out how many storage nodes there are. However, I would also like to set DEV_ENV to true if there are enough storage nodes and MONs. And I'm assuming DEV_ENV needs to get set at the very beginning of the script, i.e. before Stage 0.

rjfd:
@smithfarm DEV_ENV is only used in stage 3

smithfarm:
OK, on the assumption that we don't need any of this information until Stage 3, I'm implementing functions that use salt to count the total number of nodes, total number of nodes with storage role, etc. No need for a configuration file, at least for now.

jan--f:
I think its  a good idea to pass the aimed-for environment like this. Its portable, as-in work in teuthology and when running the script manually. It ultimately allows us to run tests with customized cluster layouts, i.e. a tester can influence policy.cfg (or in Teuthology we can pass what is specified in suite's config).
We could adjust the script to set defaults should the .config file not be present to allow simple runs without setting an explicit config.

I'd be in favour of making it a non-hodden file though. If the file would somehwo stay behind and influence a test run, a hidden file might not be very obvious, making it a bit of a magic file.

rjfd:
> It ultimately allows us to run tests with customized cluster layouts, i.e. a tester can influence policy.cfg (or in Teuthology we can pass what is specified in suite's config).

That's a good point, although in this case I would prefer a different format for the `.config` file.
Instead of just specifying the number of storage/client nodes, I think it would be preferable to explicitly enumerate the hostnames for storage/client nodes.

smithfarm:
The config file is turning into a bit of a rabbit hole. Since I've already implemented a workaround, I'm dropping it from this PR and we can revisit it later when it's really needed.

rjfd:
@smithfarm sounds good

smithfarm:
OK, this latest version passes `deepsea/basic`.

PR #331, opened 2017-06-14
qa: implement basic NFS-Ganesha integration test

smithfarm:
This PR adds a minimalistic `suites/basic/health-nfs-ganesha.sh` script to the integration test suite. It:

* adds a ganesha role
* configures NFS-ganesha for "No_root_squash" and "NIV_DEBUG"
* runs Stages 0-4
* works around buggy behavior (first attempt to communicate with the NFS server times out about half the time)
* mounts the NFS export as root
* touches a file
* unmounts the export

### DIFF
--- a/qa/common/nfs-ganesha.sh
+++ b/qa/common/nfs-ganesha.sh
@@ -1,0 +1,105 @@
+#
+# This file is part of the DeepSea integration test suite
+#
+
+function _nfs_ganesha_node {
+  _first_x_node ganesha
+}
+
+function nfs_ganesha_no_root_squash {
+  local GANESHAJ2=/srv/salt/ceph/ganesha/files/ganesha.conf.j2
+  sed -i '/Access_Type = RW;/a \\tSquash = No_root_squash;' $GANESHAJ2
+}
+
+#
+# Since we don't seem to be using NFSv4, the effect of this option is unclear
+#
+function nfs_ganesha_no_grace_period {
+  local GANESHAJ2=/srv/salt/ceph/ganesha/files/ganesha.conf.j2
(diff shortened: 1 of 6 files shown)


qa/common/nfs-ganesha.sh
```diff
@@ -0,0 +1,105 @@
...
+  _run_test_script_on_node $TESTSCRIPT $GANESHANODE
+}
+
+function nfs_ganesha_cat_config_file {
+  salt -C 'I@roles:ganesha' cmd.run 'cat /etc/ganesha/ganesha.conf'
+}
+
+function nfs_ganesha_showmount_loop {
+  local TESTSCRIPT=/tmp/test-nfs-ganesha.sh
+  salt -C 'I@roles:ganesha' cmd.run "while true ; do showmount -e $GANESHANODE | tee /tmp/showmount.log || true ; grep -q 'Timed out' /tmp/showmount.log || break ; done"
```
rjfd: I have exactly the same issue. Everytime I restart the nfs-ganesha service when a client issues the first command, either showmounts or mount, it will return a timeout. I also find out that if I just kill the command before getting the timeout and re-issue the command again, it works. So maybe this is actually a bug of the nfs-ganesha service.

smithfarm: For me, the first command sometimes works. The hang is not 100% reproducible, but easy to reproduce (roughly 50% of the time).

smithfarm:
Passed a `deepsea/basic` run at http://167.114.236.223:8081/ubuntu-2017-06-14_12:25:33-deepsea:basic-ses5---basic-openstack/

PR #333, opened 2017-06-14
NFS-Ganesha: Add "Protocols" and "Transport" in ganesha.conf

supriti:

### DIFF
--- a/srv/salt/ceph/ganesha/files/ganesha.conf.j2
+++ b/srv/salt/ceph/ganesha/files/ganesha.conf.j2
@@ -9,6 +9,10 @@ EXPORT
 
 	Access_Type = RW;
 
+	Protocols = 3, 4;
+
+	Transports = UDP, TCP;
+
     FSAL {
             Name = CEPH;
             User_Id = "{{user_id}}";
@@ -31,6 +35,10 @@ EXPORT
 	Pseudo = "/{{ user }}";
 
 	Access_Type = RW;
+
+	Protocols = 3, 4;
+
+	Transports = UDP, TCP;
 
 	FSAL {
 		Name = RGW;


srv/salt/ceph/ganesha/files/ganesha.conf.j2
```diff
@@ -8,6 +8,10 @@ EXPORT
 	Pseudo = /cephfs/;
 
 	Access_Type = RW;
+        
+        Protocols = 4;
```
rjfd: Shouldn't we specify the default, which is `3, 4`?

srv/salt/ceph/ganesha/files/ganesha.conf.j2
```diff
@@ -8,6 +8,10 @@ EXPORT
...
+
+        Transports = TCP;
```
rjfd: The same here
Shouldn't we specify the default, which is `TCP, UDP`?

smithfarm:
@supriti Good idea to force NFSv4 by default.

nit: Could the white space be lined up? Maybe the original is using a tab character to indent, and your patch is using 8 spaces?

srv/salt/ceph/ganesha/files/ganesha.conf.j2
smithfarm: Is there any scenario in which NFSv4 would use UDP? (Not sure if it has a "compatibility" mode for communicating with NFSv2/3 UDP clients.)

smithfarm: in any case, `TCP, UDP` is fine with me

srv/salt/ceph/ganesha/files/ganesha.conf.j2
smithfarm: fine with me

supriti: Agreed. Updated.

PR #338, opened 2017-06-16
DNM: refactored fsid detection, also tries to get disk id's for data and journal partitions

kmroz:
I rebased the original PR so we'd have less squashing to do later on... though it may require a `git reset --hard wip-kmroz-import-running-cluster-rebase` on `wip-import-running-cluster` to merge?  Otherwise I'll just push the whole history.

@tserong / @swiftgist - thoughts?


kmroz:
Closing this in favour of #339.

PR #339, opened 2017-06-19
populate: get existing cluster's fsid from ceph.conf

tserong:
This removes the need to talk via librados to a running cluster on
the master_minion node, which might not yet be set up to talk to
the existing cluster at the time of import.

### DIFF
--- a/srv/modules/runners/populate.py
+++ b/srv/modules/runners/populate.py
@@ -26,6 +26,12 @@ import logging
 
 import sys
 
+try:
+    import configparser
+except ImportError:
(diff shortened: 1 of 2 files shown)


kmroz:
@tserong - Per your ConfigParser concerns, I wrote up https://github.com/SUSE/DeepSea/pull/338 to cover extracting fsid's from a 'liberally' formatted ceph.conf :smile:  Seems you've gone with a little pre-processing before handing ceph.conf over to ConfigParser.  Let's merge this and I'll close out the other PR.

:+1:

Issue #341, opened 2017-06-19
After RGW deployment, curl $(hostname --fqdn) fails


smithfarm:
This bug was found by the DeepSea integration test suite. After deploying a cluster with the rgw role, I determine the FQDN of the rgw node and run `curl $RGW_FQDN`. This is supposed to produce the following XML:

```
<?xml version="1.0" encoding="UTF-8"?>
<ListAllMyBucketsResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Owner><ID>anonymous</ID><DisplayName></DisplayName></Owner><Buckets></Buckets></ListAllMyBucketsResult>
```

But instead I get:

```
2017-06-19T13:52:52.966 INFO:teuthology.orchestra.run.target167114254163.stdout:    + curl target167114254163.teuthology
2017-06-19T13:52:52.966 INFO:teuthology.orchestra.run.target167114254163.stdout:      % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
2017-06-19T13:52:52.967 INFO:teuthology.orchestra.run.target167114254163.stdout:                                     Dload  Upload   Total   Spent    Left  Speed
2017-06-19T13:52:52.967 INFO:teuthology.orchestra.run.target167114254163.stdout:
2017-06-19T13:52:52.967 INFO:teuthology.orchestra.run.target167114254163.stdout:      0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
2017-06-19T13:52:52.968 INFO:teuthology.orchestra.run.target167114254163.stdout:    100   239  100   239    0     0  32677      0 --:--:-- --:--:-- --:--:-- 34142
2017-06-19T13:52:52.968 INFO:teuthology.orchestra.run.target167114254163.stdout:    + test -f /tmp/rgw_test.xml
2017-06-19T13:52:52.968 INFO:teuthology.orchestra.run.target167114254163.stdout:    + xmllint /tmp/rgw_test.xml
2017-06-19T13:52:52.968 INFO:teuthology.orchestra.run.target167114254163.stdout:    <?xml version="1.0" encoding="UTF-8"?>
2017-06-19T13:52:52.969 INFO:teuthology.orchestra.run.target167114254163.stdout:    <Error><Code>NoSuchBucket</Code><BucketName>target167114254163.teuthology</BucketName><RequestId>tx000000000000000000001-005947d734-1016-default</RequestId><HostId>1016-default-default</HostId></Error>
```

[...]

PR #342, opened 2017-06-19
qa: Add rgw curl test to basic suite

smithfarm:
This currently gets all the way to the end before failing due to #341.


smithfarm:
With this PR and #343 applied to the `wip-qa-suite` branch (which avoids the Prometheus errors that otherwise prevent us from getting past Stage 2), this test passes.

PR #345, opened 2017-06-19
cephinspector: try to obtain disk id of journal and data partitions

kmroz:


kmroz:
Need to cover nvme formats... dnm for now.

kmroz:
@tserong / @swiftgist - added the nvme check, should be good.

srv/salt/_modules/cephinspector.py
```diff
@@ -6,17 +6,41 @@
 from subprocess import Popen, PIPE
 import json
 
+def _get_disk_id(partition):
+    """
+    Return the disk id of a partition/device, or an empty string if not available.
+    """
+    disk_id_cmd = Popen("find -L /dev/disk/by-id -samefile " + partition + " \( -name ata* -o -name nvme* \)", stdout=PIPE, stderr=PIPE, shell=True)
```
tserong: Do we care about disks whose names start with "scsi" (full disclosure: I don't really know how this by-id generation stuff works)

srv/salt/_modules/cephinspector.py
```diff
@@ -6,17 +6,41 @@
...
 def _append_to_ceph_disk(ceph_disks, partition, journal_dev):
     """
     Populate ceph_disks dictionary with data and journal partitions.
     """
+
+    # We don't care about the trailing number on journal_dev.
+    journal_dev = journal_dev.rstrip("1234567890")
+    # For nvme journal, strip the trailing 'p' as well.
+    if "nvme" in journal_dev:
+        journal_dev = journal_dev[:-1]
```
tserong: Why don't we care about the trailing numbers and 'p'? Does ceph only care about the base device, not the specific journal partition if the device is shared between multiple OSDs?

srv/salt/_modules/cephinspector.py
kmroz: Talked with @swiftgist about that.  Apparently we only care about ata/nvme.

srv/salt/_modules/cephinspector.py
kmroz: As far as I understand, our (DeepSea's) storage role yml files operate on base device for data and journal paritions.

PR #349, opened 2017-06-20
Fixed handling of FQDN hostnames

rjfd:


swiftgist:
@rjfd Small nitpick... both grains.get and grains.item... should we settle on using one?

rjfd:
@swiftgist I user `grains.item` when I need to get more than one key, and `grains.get` when I need to get a single key.

swiftgist:
Now why did Saltstack go and do that? :)  It's not like grains.item doesn't work with one argument.

PR #350, opened 2017-06-21
ganesha: RGW exports should export a user bucket instead of "/"

rjfd:
Currently NFS-Ganesha exports using RGW backend are being configured to export the `/` path. This poses some limitations to what the user can do when mounting such export. For instance, it cannot create a file, only directories.

This PR changes the RGW based exports configuration to export a bucket `demo` that is created for every RGW user configured by DeepSea.

### DIFF
--- a/srv/salt/_modules/rgw.py
+++ b/srv/salt/_modules/rgw.py
@@ -5,6 +5,10 @@ import logging
 from subprocess import call, Popen, PIPE
 import os
 import json
+import boto
+import boto.s3.connection
+import boto.exception
+
 
 log = logging.getLogger(__name__)
 
@@ -104,9 +108,24 @@ def add_users(pathname="/srv/salt/ceph/rgw/cache"):
                 log.info("stderr: {}".format(line))
 
             proc.wait()
-        
-        
-        
+
+
+def create_bucket(**kwargs):
+    s3conn = boto.connect_s3(
+        aws_access_key_id=kwargs['access_key'],
+        aws_secret_access_key=kwargs['secret_key'],
+        host=kwargs['host'],
+        is_secure=kwargs['ssl'],
+        port=kwargs['port'],
+        calling_format=boto.s3.connection.OrdinaryCallingFormat(),
+    )
+    try:
+        s3conn.create_bucket(kwargs['bucket_name'])
+    except boto.exception.S3CreateError:
+        return False
+    return True
+
+
 def _key(user, field, pathname):
     """
     Read the filename and return the key value.  
--- a/srv/salt/ceph/ganesha/files/silver-common.conf.j2
+++ b/srv/salt/ceph/ganesha/files/silver-common.conf.j2
@@ -6,9 +6,13 @@ EXPORT
 {
 	Export_ID={{ loop.index }};
 
-	Path = "/";
+	# exporting {{ user }}-demo bucket
+	Path = "{{ user }}-demo";
+	
+	# You can also export the root path / inplace of bucket name
+	# Path = "/"
 
-	Pseudo = "/{{ user }}";
+	Pseudo = "/{{ user }}/{{ user }}-demo";
 
 	Access_Type = RO;
 
--- a/srv/salt/ceph/stage/radosgw/default.sls
+++ b/srv/salt/ceph/stage/radosgw/default.sls
@@ -21,3 +21,15 @@ rgw users:
 
 {% endfor %}
 {% endif %}
+
+{% set endpoint = salt.saltutil.runner('ui_rgw.endpoints')[0] %}
+rgw demo buckets:
+  salt.state:
+    - tgt: {{ salt['pillar.get']('master_minion') }}
+    - tgt_type: compound
+    - sls: ceph.rgw.buckets
+    - pillar:
+        'rgw_host': {{ endpoint['host'] }}
+        'rgw_port': {{ endpoint['port'] }}
+        'rgw_ssl': {{ endpoint['ssl'] }}
+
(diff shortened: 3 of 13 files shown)


supriti:
LGTM. Just one point. "nfs-ganesha-rgw" package installs both the sample config files "rgw.conf" and "rgw_bucket.conf". We should document somwhere that both are valid options for nfs-ganesha. May be in the /srv/salt/ceph/ganesha/files/ganesha.conf.j2 add comment "You can also export path inplace of bucket".

rjfd:
@swiftgist do not merge this one yet, I found a bug

srv/salt/ceph/ganesha/files/silver-common.conf.j2
```diff
@@ -6,9 +6,9 @@ EXPORT
 {
 	Export_ID={{ loop.index }};
 
-	Path = "/";
+	Path = "demo";
```
theanalyst: I'd suggest creating a bucket as `{{user}}-demo` or something, this is because bucket names are unique in S3 by default and this would lead to clashes for the second user

srv/salt/_modules/rgw.py
```diff
@@ -104,9 +107,19 @@ def add_users(pathname="/srv/salt/ceph/rgw/cache"):
...
-        
-        
+
+
+def create_bucket(**kwargs):
+    s3conn = boto.connect_s3(
+        aws_access_key_id=kwargs['access_key'],
+        aws_secret_access_key=kwargs['secret_key'],
+        host=kwargs['host'],
+        is_secure=False,
```
theanalyst: can you get the is_secure value from the way we derive rgw ui credentials, if the radosgw is https terminated setting is_secure as false would fail

rjfd:
@swiftgist fixed the bug, and everything seems to be working in my test cluster

srv/salt/_modules/rgw.py
```diff
@@ -104,9 +107,19 @@ def add_users(pathname="/srv/salt/ceph/rgw/cache"):
...
+        calling_format=boto.s3.connection.OrdinaryCallingFormat(),
+    )
+    s3conn.create_bucket(kwargs['bucket_name'])
```
theanalyst: I'd wrap this in a try catch block. create_bucket would raise a boto.exception.S3CreateError or so... 
http://boto.cloudhackers.com/en/latest/ref/boto.html#boto.exception.S3CreateError

srv/salt/ceph/stage/radosgw/default.sls
```diff
@@ -21,3 +21,14 @@ rgw users:
...
+{% set minion_id = salt.saltutil.runner('select.minions', cluster='ceph', roles='rgw')[0] %}
+{% set host = salt.saltutil.cmd(minion_id, 'grains.get', ['fqdn'])[minion_id]['ret'] %}
+rgw demo buckets:
+  salt.state:
+    - tgt: {{ salt['pillar.get']('master_minion') }}
+    - tgt_type: compound
+    - sls: ceph.rgw.buckets
+    - pillar:
+        'rgw_host': {{ host }}
+
```
theanalyst: add a newline at end of the file

srv/salt/ceph/ganesha/files/silver-common.conf.j2
swiftgist: How is that even possible?  Do the users not have their own namespace?

swiftgist:
My only concern is that the default export point is the demo bucket.  When root is exported for a given user, then all buckets are accessible from a single mount.  Otherwise, the entire list of users and buckets must be managed in NFS.  That's not exactly attractive to why an admin would use NFS for rgw in the first place.

I understand having a demo bucket to show additional operations and a example is always good.

srv/salt/ceph/ganesha/files/silver-common.conf.j2
theanalyst: nope, s3 uses a global bucket namespace instead of a tenant namespace by default

rjfd:
@swiftgist 

> How is that even possible? Do the users not have their own namespace?

This is one of the reasons I'm exporting a bucket as the default instead of the root path. If you create an RGW export of "/", and then try to create a directory that clashes with the bucket name created by another user, you'll get an "-2 no file or directory exists" error, which is completely unrelated with what is actually happening.

The bottom line is, the user will try to use an RGW export in the same way it uses a CephFS export, and will find strange behaviors that he'll report as bugs, but are actually limitations of the RGW backend implementation.
But if we only export buckets, none  of the strange behavior will occur, and will be equivalent to a CephFS export. That is why I'm changing the default export to use a bucket.

srv/salt/ceph/ganesha/files/silver-common.conf.j2
rjfd: good catch

rjfd:
@supriti added a comment to the ganesha.conf* files saying that users can also export the root path "/"

rjfd:
@theanalyst addressed your comments and pushed the changes

srv/salt/_modules/rgw.py
```diff
@@ -104,9 +108,24 @@ def add_users(pathname="/srv/salt/ceph/rgw/cache"):
...
+        aws_access_key_id=kwargs['access_key'],
+        aws_secret_access_key=kwargs['secret_key'],
+        host=kwargs['host'],
+        is_secure=kwargs['ssl'],
+        port=kwargs['port'],
+        calling_format=boto.s3.connection.OrdinaryCallingFormat(),
+    )
+    try:
+        s3conn.create_bucket(kwargs['bucket_name'])
+        return True
```
theanalyst: minor nit, can you move the return true after the try except block

rjfd: done

rjfd:
@theanalyst I'm doing a test run to check if everything is working. Will update this PR with the result

rjfd:
@theanalyst test passed :)

PR #352, opened 2017-06-21
cephinspector: skip "disks" with no partitions

tserong:
When ceph-disk encounters /dev/sr0, that's included in its output, but the
dict for that device has no "partitions" key, which results in a traceback
(KeyError: 'partitions') a couple of lines later.  That traceback then lands
in the generated YAML file (yuck).

In general, we need a good way of dealing with failures like this from
cephinspector so that we don't accidentally do Stupid Things[TM] later :-/


tserong:
Actually, I'm just going to quietly merge this myself, because it's trivial, and I've got something much more worthy of review coming up in a moment ;-)

PR #353, opened 2017-06-21
ui_rgw: Use FQDN for RGW endpoints url

rjfd:
Since PR #343 the RGW endpoint URL is using the FQDN. The ui_rgw runner must also be updated accordingly.


rjfd:
@theanalyst tests fixed

PR #356, opened 2017-06-21
missing else block in no-partition branch

jschmid1:
Not sure if it was left out intentionally, but it looks like it's just missing.


swiftgist:
I believe you are correct about the else just missing.

PR #357, opened 2017-06-21
populate: import admin, mon and osd bootstrap keys

tserong:

### DIFF
--- a/srv/modules/runners/populate.py
+++ b/srv/modules/runners/populate.py
@@ -1013,6 +1013,7 @@ def engulf_existing_cluster(**kwargs):
 
     ceph_conf = None
     previous_minion = None
+    admin_minion = None
 
     for minion, info in local.cmd("*", "cephinspector.inspect").items():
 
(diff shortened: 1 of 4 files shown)


srv/modules/runners/populate.py
```diff
@@ -925,6 +925,18 @@ def engulf_existing_cluster(**kwargs):
     local.cmd("*", "saltutil.sync_all")
     local.cmd("*", "state.apply", ["ceph.mines"], expr_form="compound")
 
+    # TODO: is there a saner way to get the master_minion at this point (when
+    # there might not yet be a master role assigned?)
```
kmroz: I think this is all we have to work with at this point.

srv/modules/runners/populate.py
```diff
@@ -925,6 +925,18 @@ def engulf_existing_cluster(**kwargs):
...
+    master_minion = ""
+    for minion, master in local.cmd("*", "pillar.get", ["master_minion"]).items():
+        if master:
```
kmroz: What do you think - `if master and master != "_REPLACE_ME_":` ?  Might save a developer, that copied over their wip branch, some debug time :smile:

tserong:
OK, this should be much less clever now ;-)

PR #364, opened 2017-06-21
grafana: insist on /var/lib/grafana

smithfarm:
The grafana-server.service will fail to start if this directory doesn't exist.
It's supposed to be created by the grafana RPM, but we had a case where it
wasn't getting created and we were failing farther down, in the "curl" command.

### DIFF
--- a/srv/salt/ceph/monitoring/grafana/init.sls
+++ b/srv/salt/ceph/monitoring/grafana/init.sls
@@ -9,6 +9,9 @@ grafana-server:
       - pkg: grafana
     - watch:
       - file: /etc/grafana/grafana.ini
+  file.exists:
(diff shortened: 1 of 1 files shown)


smithfarm:
The argument for doing this is: "fail early, fail often"

I.e. if we know for sure that grafana will not work without this directory, fail as soon as we can ascertain that it's not there, with an error message the points to the real problem.

jan--f:
I got rid of the conditional creation of that directory in the grafana package so it'll be there if the installation was successful. So I'd be inclined to not check this here. Also not exactly an early fail, since the package installation should have failed.

jan--f:
not needed anymore.

PR #369, opened 2017-06-22
qa: install jq dependency at beginning of each test

smithfarm:
It came to my attention that not everyone has jq installed :-)


smithfarm:
I'll give it a spin

PR #376, opened 2017-06-26
Add condition to avoid uninitialized var most_common_items

### LINKED ISSUE #372: status.report: most_common_item can be uninitialized and trigger traceback
```
def report(cluster_name='ceph', stdout=True, return_data=False ):
...
        if counter_obj.most_common():
            most_common_item = counter_obj.most_common()[0][0]
--->    common_keys.update({key_ident: most_common_item})
...
```

jschmid1:
closes #372

### DIFF
--- a/srv/modules/runners/status.py
+++ b/srv/modules/runners/status.py
@@ -33,6 +33,7 @@ def report(cluster_name='ceph', stdout=True, return_data=False ):
         key_ident = minion_data[0]
         minion_data_dct = minion_data[1]
         counter_obj = Counter(minion_data_dct.values())
+        most_common_item = None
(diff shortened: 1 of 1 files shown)


srv/modules/runners/status.py
```diff
@@ -35,12 +35,13 @@ def _organize(minion_data):
         counter_obj = Counter(minion_data_dct.values())
         if counter_obj.most_common():
```
kmroz: I think you need to set `most_common_item = None` outside the `if` statement?

jschmid1:
``` python
Counter()
```
can't be None as it's always returning a Counter object. setting most_common_item to None is sufficient.

PR #379, opened 2017-06-26
Add osd/partition to spec and makefile

supriti:


jschmid1:
Shouldn't the osd/partition (and possibly more) be deprecated in favor of the new osd.py module?

supriti:
I was trying to setup a "default-small" cluster, and it complained about missing ceph.osd.partition. In file, https://github.com/SUSE/DeepSea/blob/master/srv/salt/ceph/osd/default-small.sls we are including partition. I am not aware of osd.py module, so could be we just missed out updating it in default-small.sls.

swiftgist:
We need to remove default-small.sls.  I missed that while cleaning up.  The default.sls now calls osd.deploy which handles drives under 10G too.

supriti:
@swiftgist Okay. I was not aware of osd.deploy handling the case.  I will close this PR then. Wiki also need to updated accordingly(https://github.com/SUSE/DeepSea/wiki/override)

PR #381, opened 2017-06-26
packagemanager: fixup updates needed check for zypper dup

### LINKED ISSUE #380: packagemanager.dup: fails to run `zypper dup`
Issue is with `Zypper::_updates_check()`.  `zypper lu` will not determine if upgrades (via `zypper dup`) are needed.

kmroz:
Fixes #380

Signed-off-by: Karol Mroz <kmroz@suse.de>

### DIFF
--- a/srv/salt/_modules/packagemanager.py
+++ b/srv/salt/_modules/packagemanager.py
@@ -176,6 +176,14 @@ class Zypper(PackageManager):
             log.debug('Executing {}'.format(cmd))
             return False
 
+    def _upgrades_needed(self):
+        """
+        In the upgrade path, always perform a `zypper dup`.  `zypper dup` has no equivalent
+        `zypper lu` check, and performing a `zypper dup --dry-run` followed by a `zypper dup`
+        just adds to wait time for the admin.
+        """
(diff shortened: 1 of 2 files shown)


kmroz:
This breaks some unit tests... will need to fix.

kmroz:
Re-factored to avoid breaking unit tests.

kmroz:
@jschmid1 - I _think_ you replaced all invocations of ceph.update with ceph.upgrade during our debug session... I've pushed those changes here as well.  Could you take a look please?  We're not relying on maintenance/upgrade/* to perform `zypper up`, correct?  Or would this change modify the expected behaviour?

kmroz:
Works for me as it is now.

jschmid1:
> @jschmid1 - I think you replaced all invocations of ceph.update with ceph.upgrade during our debug session...

Correct

> I've pushed those changes here as well. Could you take a look please? We're not relying on maintenance/upgrade/* to perform zypper up, correct? Or would this change modify the expected behaviour?

No, we only perform a zypper dup but ask the user to update the cluster beforehand.

PR #385, opened 2017-06-28
stage 1: network discovery revisited

rjfd:
This new discovery algorithm uses two new heuristics to find the
main public network and the cluster network.

The heuristic to find the public networks works like this:
If a network matches one of the "hostname -i" of the minions, then
the network is considered as a public network.

The heuristic to find the cluster networks works like this:
If a network that has more than one minion does not match any
of the salt-master interface addresses, then the network is
considered as a cluster network.

Otherwise, the discovery fails to the already existing behavior.

### DIFF
--- a/srv/modules/runners/populate.py
+++ b/srv/modules/runners/populate.py
@@ -26,6 +26,7 @@ import ipaddress
 import logging
 import deepsea_minions
 import validate
+import operator
 
 import sys
 
@@ -477,7 +478,7 @@ class CephRoles(object):
         self.search = target.deepsea_minions
 
         self.networks = self._networks(self.servers)
-        self.public_networks, self.cluster_networks = self.public_cluster(self.networks)
+        self.public_networks, self.cluster_networks = self.public_cluster(self.networks.copy())
 
         self.available_roles = [ 'storage' ]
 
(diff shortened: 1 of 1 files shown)


swiftgist:
I'll give this a try soon.

Martin-Weiss:
I do not believe that we can rely on "hostname -i" to be the proper answer for the public network.
Any idea why we rely on the IP address at all instead of using the DNS name?

A cluster node can have many IP addresses and it depends on the customers naming conventions and network rules which IP matches to which hostname and to /etc/HOSTNAME.

Think about hosting iSCSI, public, cluster, admin, RGW networks on the same node. The only rule that should exist is that a separate name for each IP needs to be existing in DNS.

Oh - and it might even be the case that the "admin" server can not reach the public nor cluster network at all as both are storage networks that might be isolated. In this scenario the admin server or SUSE Manager server will only be able to connect to the cluster nodes via the admin network..

I believe we need to have some sort of a configuration file where an admin can specify what the admin, public, cluster, iscsi, rgw, nfs, cifs etc. networks are as any sort of auto-detection will be false in one or the other network.

An other option might be to specify "DeepSea only supports public and cluster network and the IP of the public network MUST match the /etc/HOSTNAME and must be in the public network" - but this would limit the scenarios in which we could deploy...

So 
1. could we change the usage of IP addresses to using DNS names?
2. can we add some sort of configuration file where we specify which subnets to be used for what?
Hint: for 2. we have /srv/pillar/ceph/proposals/config/stack/default/ceph/cluster.yml already - so why don´t we add the required information here BEFORE executing stage.0/1/2?

Martin-Weiss:
Have tried this with SES 5 milestone 9 but I believe that the populate.py is not the right version as I get this output which does not include the new proposal creation for the disks:

admin-p:/srv/pillar/ceph # salt-run populate.proposals
[WARNING ] Although 'dmidecode' was found in path, the current user cannot execute it. Grains output might not be accurate.
[WARNING ] No proposal for Intel372GB as journal on 2INTEL372GB-2Intel372GB-12HGST3726GB
[WARNING ] No proposal for Intel372GB as journal on 2INTEL372GB-2Intel372GB-12HGST3726GB
[WARNING ] No proposal for Intel372GB as journal on 2INTEL372GB-2Intel372GB-12HGST3726GB
[WARNING ] No proposal for Intel372GB as journal on 2INTEL372GB-2Intel372GB-12HGST3726GB

However - the guessed IP addresses look much better:

admin-p:/srv/pillar/ceph/proposals # grep -ir 172 *
config/stack/default/ceph/cluster.yml:cluster_network: 172.16.2.0/24
config/stack/default/ceph/cluster.yml:public_network: 172.16.1.0/24, 10.2.3.0/24, 10.2.5.0/24
role-igw/stack/default/ceph/minions/osd01-p.ses.intern.thomas-krenn.com.yml:public_address: 172.16.1.51
role-igw/stack/default/ceph/minions/admin-p.ses.intern.thomas-krenn.com.yml:public_address: 172.16.1.50
role-igw/stack/default/ceph/minions/osd04-p.ses.intern.thomas-krenn.com.yml:public_address: 172.16.1.54
role-igw/stack/default/ceph/minions/osd03-p.ses.intern.thomas-krenn.com.yml:public_address: 172.16.1.53
role-igw/stack/default/ceph/minions/osd02-p.ses.intern.thomas-krenn.com.yml:public_address: 172.16.1.52
role-mon/stack/default/ceph/minions/osd01-p.ses.intern.thomas-krenn.com.yml:public_address: 172.16.1.51
role-mon/stack/default/ceph/minions/admin-p.ses.intern.thomas-krenn.com.yml:public_address: 172.16.1.50
role-mon/stack/default/ceph/minions/osd04-p.ses.intern.thomas-krenn.com.yml:public_address: 172.16.1.54
role-mon/stack/default/ceph/minions/osd03-p.ses.intern.thomas-krenn.com.yml:public_address: 172.16.1.53
role-mon/stack/default/ceph/minions/osd02-p.ses.intern.thomas-krenn.com.yml:public_address: 172.16.1.52

--> no 10.x (admin network) used for the public_address of the minions - VERY GOOD!

but in the config/stack/default/ceph/cluster.yml I see still all networks the Admin has:

public_network: 172.16.1.0/24, 10.2.3.0/24, 10.2.5.0/24

But that might not cause the major issues we have seen with the wrong addresses for the MON and IGW.

So yes - I believe your fix for IP address guessing makes our implementation better! 
Could you get it checked in, please?

rjfd:
> Have tried this with SES 5 milestone 9 but I believe that the populate.py is not the right version as I get this output which does not include the new proposal creation for the disks:

Correct, this PR needs to be rebased on top of the latest master to get the new proposal scheme.

> but in the config/stack/default/ceph/cluster.yml I see still all networks the Admin has:
> public_network: 172.16.1.0/24, 10.2.3.0/24, 10.2.5.0/24

I thought of this problem and think I have a solution but I need your opinion. Do you think that is admissible to just keep the first public network discovered by the algorithm if the that network includes all minions and the salt-master?

I don't see a reason to add two public network entries if one of them already includes all the nodes of your cluster, does this make sense?

Martin-Weiss:
>>    but in the config/stack/default/ceph/cluster.yml I see still all networks the Admin has:
   public_network: 172.16.1.0/24, 10.2.3.0/24, 10.2.5.0/24

>I thought of this problem and think I have a solution but I need your opinion. Do you think that is  admissible to just keep the first public network discovered by the algorithm if the that network includes all  minions and the salt-master?

It depends on "what is the first one"! My best practice is that /etc/HOSTNAME of the servers ALWAYS uses the public network name and refers to the public IP - but that might not be "the first one found".

> I don't see a reason to add two public network entries if one of them already includes all the nodes of your cluster, does this make sense?

The requirement for "public_network" is that ALL networks where we have MONs are part of the public_network configuration. So in an scenario where I have two MON in 10.1.1.0/24 (10.1.1.3 and 10.1.1.4) and one MON in 10.2.2.0/24 (10.2.2.5) - the public_network needs to have exactly these two networks and none of the other ones.

abonillasuse:
I like this approach with the heuristic method. One thing as well could be to guess the public network by looking at the routing table and using the interface being used for the default route. I just deployed an M10 and the discovery was exact.

Martin-Weiss:
> One thing as well could be to guess the public network by looking at the routing table and using the interface being used for the default route. 

I have several customers that want their public network to be isolated for storage communication and they use the admin interface for NTP, DNS, patching, SSH, salt etc. In these scenarios the default route goes to the admin interface which is NOT the public interface one..

rjfd:
@Martin-Weiss I rebased the PR on top of the latest master and also improved the construction of the final public network list. Can you test this patch again?

smithfarm:
@rjfd Please rebase again.

rjfd:
@smithfarm done, will you run the teuhology suites with this PR? that would be great

smithfarm:
**running** http://149.202.174.158:8081/ubuntu-2017-09-01_17:23:24-deepsea:basic:health-ok-ses5---basic-openstack/

smithfarm:
@rjfd Running health-ok first. If it passes, next step would be the full basic suite.

smithfarm:
**running** http://149.202.174.158:8081/ubuntu-2017-09-01_17:41:13-deepsea:basic-ses5---basic-openstack/

rjfd:
@smithfarm pushed a fix, please re-run test suite.

smithfarm:
@rjfd That change seems to have done the trick - all tests are green now: http://149.202.174.158:8081/ubuntu-2017-09-04_07:00:01-deepsea:basic-ses5---basic-openstack/

rjfd:
@swiftgist @jan--f could you test this PR by deploying DS in any cluster you have available?

jschmid1:
Correctly detects cluster/pub networks on my setup. LGTM

rjfd:
@jschmid1 thanks for trying it.

@swiftgist do you think this can get in for the next milestone?

Issue #368, opened 2017-06-22
[qa] Create a stripped down version of stage.0


jschmid1:
After trying the newly added qa/suites/basic/* suites on my vagrant setup, I think we might need a stripped down/small version of stage.0

### Reason
(Under the premise you haven't updated your base image in a while)
1) If you quickly want to bootstrap a cluster and programmatically execute i.e. 'health-ok.sh', the master will be updated/rebooted, hence 'heath-ok.sh' will be interrupted. 

2) The update process, ofc depending on the state of your base image, will take up most of the time and will hinder you from testing more frequently during development.

3) Updating from the already added SUSE repos will also update 'DeepSea', which makes testing an older version a pain.

### What should be removed
 * `ceph.updates.restart`
* `ceph.updates`

Thoughts?

smithfarm:
I thought about not running stage 0 in the test, since it breaks every time a kernel update is released (Salt reboots the VMs and that's the end of the test).

But after some thought I decided against it, because running a different version of one of the stages reduces test coverage. I deploy my OVH images automatically, so redeploying them for each kernel update is not a problem.

smithfarm:
I don't mind having this in DeepSea, but I would object to its use in the DeepSea integration suite. Better to put our efforts into making it easy to have always up-to-date images, vagrant boxes, etc.

jschmid1:
> But after some thought I decided against it, because running a different version of one of the stages reduces test coverage.

If you always have an up-to-date base image you will only test a certain amount of the code anyways, i.e. never execute the underlying module. We wouldn't loose too much there.

jschmid1:
> I don't mind having this in DeepSea, but I would object to its use in the DeepSea integration suite. Better to put our efforts into making it easy to have always up-to-date images, vagrant boxes, etc.

Would be the cleaner solution. That's something we should pursue regardless.

jan--f:
Another way to deal with this is to make sure the kernel packages are in a good state before starting the test. For vagrant this can be part of the provisioning.

jschmid1:
> Another way to deal with this is to make sure the kernel packages are in a good state before starting the test. For vagrant this can be part of the provisioning.

good point. 

Without making a debate on principles of this, I'd like to raise the general need for a switch to disable updates and reboots ( i.e. in a SUMA controlled env )

smithfarm:
> If you always have an up-to-date base image you will only test a certain amount of the code anyways, i.e. never execute the underlying module.

[...]

PR #387, opened 2017-06-28
Conditional reboots and updates

jschmid1:
We had this request multiple times in the past.

Followup to issue #368

### DIFF
--- a/srv/salt/ceph/stage/prep/master/default.sls
+++ b/srv/salt/ceph/stage/prep/master/default.sls
@@ -20,10 +20,16 @@ repo master:
     - tgt: {{ salt['pillar.get']('master_minion') }}
     - sls: ceph.repo
 
+{% if salt['pillar.get']('no_update') != True %}
+
 prepare master:
   salt.state:
     - tgt: {{ salt['pillar.get']('master_minion') }}
     - sls: ceph.updates
+
+{% endif %}
+
+{% if salt['pillar.get']('no_reboot') != True %}
 
 {% set kernel= grains['kernelrelease'] | replace('-default', '')  %}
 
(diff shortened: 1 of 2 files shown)


swiftgist:
Would it be easier to have default-noupdate.sls or something?  With two variables, we have other situations to consider.  If I set updates, but not reboots, do we really want to debug why things aren't working because the wrong kernel is running?  Or processes are still using the previous version of a library?

The opposite seems benign.  I am willing to reboot, but I don't want updates. That seems okay, but the workflow is strange.  I want to manually do updates myself, but then I'm willing for Stage 0 to reboot my nodes if the kernel changed.

jan--f:
I also think we should stay consistent with our pillar usage and provide a default-noreboot.sls or so to use with stage_prep_*.

Though imho the default should really be something minimal, i.e. without the updates. But I suppose that is a different issue.

jschmid1:
> Would it be easier to have default-noupdate.sls or something? With two variables, we have other situations to consider. If I set updates, but not reboots, do we really want to debug why things aren't working because the wrong kernel is running? Or processes are still using the previous version of a library?
> I also think we should stay consistent with our pillar usage and provide a default-noreboot.sls or so to use with stage_prep_*.

Understood.

> Though imho the default should really be something minimal, i.e. without the updates. But I suppose that is a different issue.

I fully support that. Even if we tell our users to run a 'Non-Stage' command in order to get updates and reboots. There _definitely_ will be the request to avoid automatic reboots and updates. And imho, telling them to use an alternative stage is not suuper nice.
_\<nit>_ In addition we are duplicating code, which needs to be maintained. _\</nit>_

jschmid1:
figured the resulting amount of files is pretty messy.

1) Updates but no reboots
2) Reboots but no updates
3) Reboots and updates
4) No reboots no updates

smithfarm:
@jschmid1 Which of those 4 will be the default?

jschmid1:
That's what we should agree on. What I hear from i.e. Martin is - Updates are ok, please no reboots.

I personally also vote for:

1) Updates but no reboots

swiftgist:
@jschmid1 if you allow updates, but do not do reboots, what do you expect from the admin?

jschmid1:
Only kernel (and certain other things like glibc) updates do require reboots to take effect. What I read and heard, it's common practice to reboot machines in a predefined maintenance window.

jschmid1:
closed in favor of #419

PR #388, opened 2017-06-28
Wip osd

swiftgist:
Support migrating one node at a time.


srv/salt/ceph/migrate/osds/default.sls
```diff
@@ -8,19 +8,33 @@ safety is engaged:
 
 {% endif %}
 
+wait on healthy cluster:
+  salt.state:
+    - tgt: {{ salt['pillar.get']('master_minion') }}
+    - tgt_type: compound
+    - sls: ceph.wait.until
```
jschmid1: ceph/wait/until needs to be added to deepsea.spec and Makefile

swiftgist: Will rework with the slightly different pathnames we discussed in email.

PR #389, opened 2017-06-28
Only warn about a missing profiles

jschmid1:
Without having to use DEV_ENV, one can now deploy Monitors without being forced to have 'profiles*/' defined.

Follow-up as a request from Martin.

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -648,7 +648,7 @@ class Validate(object):
 	profiles = [ prf  for prf in accum_files if 'profile' in prf ]
 	if not profiles:
 	    message = "There are no files under the profiles directory. Probably an issue with the discovery stage."
-	    self.errors.setdefault('profiles_populated', []).append(message)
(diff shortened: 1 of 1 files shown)


swiftgist:
The reason for the request is that Martin wanted to do an incremental configure and step-by-step of the process.  I am not in favor of doing incremental configurations.  That creates a lot of additional checks beyond the normal validations.

For the step-by-step, I updated the bug to list out all of the steps in Stage 3.  If an admin wants to do each step, then manually following the /srv/salt/ceph/stage/3/default.sls is the way to do it.

I think changing this to a warning will bite other users that do not understand the impact of the warning, will proceed with Stage 3 and then be unhappy.  For the normal user running the orchestrations, missing hardware profiles really is an error.

jan--f:
hmm I agree with @swiftgist. If they want to do step by step they can either use DEV_ENV or apply states manually or write their own orchestration.

swiftgist:
Prefer current behavior of returning Error

PR #390, opened 2017-06-29
configuration: add RBD specific configuration template

rjfd:


swiftgist:
Is rbd default features = 3 still necessary?

rjfd:
@swiftgist Nothing has changed in kRBD and therefore if we remove the ` rbd default features = 3` the kernel client will be unable to mount the images

swiftgist:
With ceph.conf.d, I think the intention is that these would always be user files and DeepSea wouldn't own anything there other than the README.  

Would it be better to move the rbd.conf up one level and add it to the Makefile/spec file?  In the long term, should the global section be segregated into several files and ceph.conf.j2 is simply includes?

swiftgist:
Never mind me.  Apparently, that eye checkup yesterday did no good. :)

Issue #183, opened 2017-03-31
DeepSea doesn't know how to deploy ceph-mgr


tserong:
Since Kraken, ceph-mgr will be automatically started on the mon nodes if it's not otherwise configured to run elsewhere. We should teach DeepSea how to deploy ceph-mgr for the case where you don't want to run it on the mon nodes (the service will need to be masked on the mon nodes and enabled and started on whatever other nodes the user chooses).

LenzGr:
Good suggestion - I wonder if that would become another possible node role?

swiftgist:
I think that would make the most sense.  Volker is doing something similar for an openattic role.  https://github.com/votdev/DeepSea/commit/7557af6d7535b673ad5b536b0db2e1d3d0cecfcb

Same idea, but for ceph-mgr.

tserong:
Note to self: also add an mgr section for custom config (see #256)

tserong:
The client admin keyring will also need `caps mgr  = "allow *"`

tserong:
Working on this now (the systemd magic to automatically create mgrs is going away soon -- see https://github.com/ceph/ceph/pull/16023)

tserong:
Don't close this yet. Once #398 is merged, we still need to add:

- remove
- rescind
- mgr support when importing existing clusters
- mgr section for custom config as mentioned above

(`caps mgr = "allow *" has already been taken care of)

PR #398, opened 2017-06-30
Add role-mgr, to deploy ceph-mgr instances

tserong:
PROBLEMS:
- Error EINVAL: entity mgr.ses4-2 exists but key does not match
(when run over an existing cluster with auto-deployed mgrs)

TODO:
- remove
- rescind
- only mandatory on Luminous+

ALSO:
- add mgr support to engulfer

Fixes: https://github.com/SUSE/DeepSea/issues/183

### DIFF
--- a/srv/salt/ceph/stage/deploy/default.sls
+++ b/srv/salt/ceph/stage/deploy/default.sls
@@ -61,11 +61,30 @@ admin:
     - tgt_type: compound
     - sls: ceph.admin
 
+mgr keyrings:
+  salt.state:
+    - tgt: 'I@roles:mgr and I@cluster:ceph'
+    - tgt_type: compound
+    - sls: ceph.mgr.keyring
+    - failhard: True
+
 monitors:
   salt.state:
     - tgt: 'I@roles:mon and I@cluster:ceph'
     - tgt_type: compound
     - sls: ceph.mon
+    - failhard: True
+
+mgr auth:
+  salt.state:
+    - tgt: {{ salt['pillar.get']('master_minion') }}
+    - sls: ceph.mgr.auth
+
+mgrs:
+  salt.state:
+    - tgt: 'I@roles:mgr and I@cluster:ceph'
+    - tgt_type: compound
+    - sls: ceph.mgr
     - failhard: True
 
 setup ceph exporter:
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,19 @@ copy-files:
 	install -m 644 srv/salt/ceph/mds/files/*.j2 $(DESTDIR)/srv/salt/ceph/mds/files/
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/mds/restart
 	install -m 644 srv/salt/ceph/mds/restart/*.sls $(DESTDIR)/srv/salt/ceph/mds/restart
+	# state files - mgr
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/mgr
+	install -m 644 srv/salt/ceph/mgr/*.sls $(DESTDIR)/srv/salt/ceph/mgr/
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/mgr/key
+	install -m 644 srv/salt/ceph/mgr/key/*.sls $(DESTDIR)/srv/salt/ceph/mgr/key/
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/mgr/auth
+	install -m 644 srv/salt/ceph/mgr/auth/*.sls $(DESTDIR)/srv/salt/ceph/mgr/auth/
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/mgr/keyring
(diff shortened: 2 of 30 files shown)


tserong:
This is not ready, but I could use some feedback on a couple of things, notably:

1) Deploying mgrs is mandatory for Luminous+, and optional for Kraken. Does DeepSea know what version of ceph it's trying to deploy? Think: making the validate runner not require role-mgr if Jewel is being deployed. Or is that not a priority?

2) The auth key generation and everything here follows pretty much the same pattern as for MDS. This means DeepSea first generates keys in /srv/salt/ceph/mgr/cache, then later runs `ceph auth add`. If you already have a Kraken or Luminous cluster running (which right now will have the systemd magic auto-created ceph-mgr instances), *then* you add role-mgr to your policy.cfg, reflecting where the existing mgrs are, when you run stage 3 it will fail, because it's trying to replace existing mgr keys with newly created ones. Do we need to solve this?

tserong:
Actually, I can have it forcibly clobber any existing mgr keys with the ones DeepSea generates, which I think solves question 2 (see previous commit). Is this in any way bad?

To clarify the problem: in Kraken, and in current builds of Luminous, there's some systemd magic which causes mgr to start on all the MON nodes. If mgr keyrings don't exist, they will be generated automatically when the service starts, thanks to the magic. This means that if you add role-mgr to policy.cfg, whether or not you're running stage 3 to deploy a new cluster, or re-running it on an existing cluster, the ceph.mgr.auth piece will fail, because the keys DeepSea generates don't match the ones that already exist. Forcing replacement of the existing keys fixes this.

Once https://github.com/ceph/ceph/pull/16023 goes in, that problem evaporates for Luminous, so it won't cause trouble on new deployments, i.e. we don't need a forced overwrite in this case. But it still will be a problem for any existing Kraken or pre-release Luminous clusters managed by DeepSea.

I figure we can either go with my clobbering, or document that you should manually replace /srv/salt/ceph/mgr/cache/*keyring with keys from your existing cluster if you hit the error "Error EINVAL: entity mgr.HOSTNAME exists but key does not match" when running stage 3. Which of these two is least worst?

swiftgist:
Would it be possible to install a generated key for mgr prior to starting the mons?  That is, put the file in the right place during mon creation?

tserong:
Should be. That would fix it for new deploys, regardless of whether the systemd magic key generation is present or not (the magic only activates if there's not already a keyring). The problem would then only appear for existing Kraken or prerelease Luminous clusters.

tserong:
Yeah, seems to work, although for complete assurance I should really do a proper redeploy from scratch, rather than what I just did, which was manually rip the running mgrs out of an existing cluster then re-run stage 3.

Anyway, this means the problem will only remain for existing Kraken and prerelease Luminous clusters. In those cases, possibly documenting some manual screwing around is preferable to weird hacks?

swiftgist:
The current side effect is Stage 3 complaining that the keyring file on disk doesn't match the keyring in `ceph auth list`.  Asking the admin to choose between overwriting the cache file with the contents of `ceph auth get mgr.nodeX` or running `ceph auth del mgr.nodeX` and rerunning a Stage doesn't feel too terrible.

smithfarm:
> the problem will only remain for existing Kraken and prerelease Luminous clusters. In those cases, possibly documenting some manual screwing around is preferable to weird hacks

+1 

Supporting kraken and pre-release Luminous should be very low on the priority list IMO.

jschmid1:
@smithfarm, @tserong, @swiftgist 
Just noticed that https://github.com/ceph/ceph/pull/16023 already found it's way into our codebase, which means that the default deployment is broken. 
We should merge the basic functionality asap and hand in remaining features as remove, rescind later.

smithfarm:
Yeah, sorry, I'm merging upstream master into the product on a regular basis - fail what may.

jschmid1:
Not your fault @smithfarm :)
Just raising awareness that we should ship the corresponding fix asap

tserong:
@jschmid1, I've squashed that back to one commit so it can (hopefully) be merged immediately. I've also added mgr to the list in `srv/modules/runners/orderednodes.py:29`. I'm not sure about `srv/modules/runners/populate.py:519` -- there's a list on line 526, but I already added mgr to that earlier.

jschmid1:
yikes, for the time being we need a `test.nop` in '/srv/salt/ceph/rescind/mgr' to avoid breakage

srv/salt/ceph/stage/deploy/default.sls
```diff
@@ -61,13 +61,32 @@ admin:
     - tgt_type: compound
     - sls: ceph.admin
 
+mgr auth:
```
jschmid1: Just had a timeout during deployment due to a, supposedly wrong ordering.

We try to auth against a monitor that is not yet deployed, see line 76.
This should be moved below the mon creation

tserong:
Please don't merge this yet, I might have missed something, need to test again.

tserong:
@jschmid1, the latest version implements rescind, and fixes the mgr auth ordering WRT mon creation.

jschmid1:
giving it a spin @tserong

jschmid1:
works like a charm! lgtm from my side

PR #399, opened 2017-06-30
osd.py: always find highest partition for filestore

tserong:
Currently, for filestore devices, if the journal is on the same
device, osd_partition() is hardcoded to return 2.  This means that
is_prepared() will later return False when run against a disk
where the OSD paritition is actually partition 1.  Instead, let's
just have it always return the highest OSD-type partition number.

Question: can we remove the note that says "Do not remove the logic
with the hardcoded values until the evaluation of activate command
is resolved"?

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -902,21 +902,10 @@ class OSDCommands(object):
     def osd_partition(self):
         """
         Find the OSD partition
-
-        Note: Do not remove the logic with the hardcoded values until
-        the evaluation of activate command is resolved
         """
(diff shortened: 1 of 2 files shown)


swiftgist:
Yeah, we can get rid of the note.  I believe I wrote that before _highest_partition existed.

tserong:
OK, cool. I just need to fix/update/do something sensible with the failing tests. Will take care of that on Monday, or thereabouts.

swiftgist:
Created #422.  I may be overthinking the activate command, but with little time to test all cases, I went with just changes to is_prepared and is_activated.

PR #401, opened 2017-06-30
Refrain from creating a demo RBD image

smithfarm:
The rbd pool is longer created as of https://github.com/ceph/ceph/pull/15894 - so don't create a demo image in it, either.

### DIFF
--- a/srv/salt/ceph/pool/default.sls
+++ b/srv/salt/ceph/pool/default.sls
@@ -8,9 +8,3 @@ wait:
         'status': "HEALTH_ERR"
     - fire_event: True
 
-demo image:
-  cmd.run:
-    - name: "rbd -p rbd create demo --size=1024"
-    - unless: "rbd -p rbd ls | grep -q demo$"
-    - fire_event: True
-


smithfarm:
Just out of curiosity, why is DeepSea creating this image?

I tried running the `rbd/cli-generic.sh` workunit, but it bails out right at the beginning with the following error:

```
2017-06-30T16:53:27.632 INFO:tasks.workunit.client.0.target137074026213.stderr:+ rbd ls
2017-06-30T16:53:27.632 INFO:tasks.workunit.client.0.target137074026213.stderr:+ wc -l
2017-06-30T16:53:27.632 INFO:tasks.workunit.client.0.target137074026213.stderr:+ grep -v '^0$'
2017-06-30T16:53:27.852 INFO:tasks.workunit.client.0.target137074026213.stdout:1
2017-06-30T16:53:27.852 INFO:tasks.workunit.client.0.target137074026213.stdout:nonempty rbd pool, aborting!  run this script on an empty test cluster only.
2017-06-30T16:53:27.853 INFO:tasks.workunit.client.0.target137074026213.stderr:+ echo 'nonempty rbd pool, aborting!  run this script on an empty test cluster only.'
2017-06-30T16:53:27.853 INFO:tasks.workunit.client.0.target137074026213.stderr:+ exit 1
2017-06-30T16:53:27.853 INFO:tasks.workunit:Stopping ['rbd/cli_generic.sh'] on client.0...
```

Obviously I could delete the pool myself, but better for DS to not create it at all, I figure, unless it's really needed for something.

swiftgist:
The demo image is serving as an example for iSCSI configuration.  I think it's okay to move this to an alternate default.  If the rbd pool is created by default, what pool is?  Or is ceph supposed to remain in HEALTH_WARN/HEALTH_ERR?

smithfarm:
Did the absence of any pools cause HEALTH_WARN/HEALTH_ERR before? I don't think it does anymore (in latest master). Will investigate further.

How can I move this code to an alternative default instead of outright dropping it?

smithfarm:
FWIW the new rbd-cli-generic test case in teuthology DeepSea suite passes when this PR is included.

smithfarm:
Also, it's probably worth repeating that the upstream PR that stops Ceph from automatically creating the rbd pool is https://github.com/ceph/ceph/pull/15894

jschmid1:
> Did the absence of any pools cause HEALTH_WARN/HEALTH_ERR before? I don't think it does anymore (in latest master). Will investigate further.

Not sure if it ever did, but `ceph version 12.1.0-671-g2c11b88d14 (2c11b88d14e64bf60c0556c6a4ec8c9eda36ff6a) luminous (rc)` it doesn't

PR #403, opened 2017-07-03
add -n, --replacefiles flags to zypper migration

jschmid1:
Fixes bug in the upgrade_init: migration path.

When upgrading more than one base product, which we do, there is a selection to make, hence passing `-n`

Only difference is the order in which you have to provide the `--non-interactive` switch. in `zypper migration` you have to pass it to the subcommand(migration) instead of zypper.

Intentionally _NOT_ passing the `--allow-vendor-change`.

### DIFF
--- a/srv/salt/_modules/packagemanager.py
+++ b/srv/salt/_modules/packagemanager.py
@@ -282,7 +282,7 @@ class Zypper(PackageManager):
             log.info("No updates available.")
 
     def _migrate(self):
-        # There is no dryrun or replacefiles or non-interactive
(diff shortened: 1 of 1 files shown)


srv/salt/_modules/packagemanager.py
```diff
@@ -283,10 +283,13 @@ def _handle(self, strat='up'):
 
     def _migrate(self):
         # There is no dryrun or replacefiles or non-interactive
```
jan--f: Update comment here.

Issue #386, opened 2017-06-28
Remove osd/default_small.sls


supriti:
osd.py handles the small configuration now. We can get rid of default_small.sls Wiki should be updated accordingly. https://github.com/SUSE/DeepSea/wiki/override 

Refer: https://github.com/SUSE/DeepSea/pull/379 .

PR #409, opened 2017-07-04
Get rid of osd debris

jschmid1:
partly addresses #386 

there should be no more invocation of `osd.partition`


jschmid1:
Wiki also updated, see https://github.com/SUSE/DeepSea/wiki/override/3948717bf757d637cc10f7db524a3826bbdf22e1

PR #412, opened 2017-07-05
Add policy migration

swiftgist:

### DIFF
--- a/srv/modules/runners/push.py
+++ b/srv/modules/runners/push.py
@@ -75,6 +75,26 @@ def proposal(filename = "/srv/pillar/ceph/proposals/policy.cfg", dryrun = False)
     pillar_data.output(common)
     return True
 
+def convert(filename = "/srv/pillar/ceph/proposals/policy.cfg"):
+    """
+    Convert the hardware profiles that policy.cfg is using and update
+    the policy.cfg.
+
+    Note: While the call "push.convert" does not make much sense, this
+    runner contained most of the necessary logic
+    """
+    if not os.path.isfile(filename):
+        log.warning("{} is missing - nothing to migrate".format(filename))
+        return ""
+    if os.path.isfile("{}-original".format(filename)):
+        log.error("Already migrated - remove {}-original before rerunning".format(filename))
+        return ""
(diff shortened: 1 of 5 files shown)


srv/modules/runners/push.py
```diff
@@ -132,6 +152,56 @@ def output(self, common):
...
+        """
+        Migrate the original data structure to the ceph namespace data
+        structure
+        """
+        if 'storage' in yaml and 'osds' in yaml['storage']:
+            yaml['ceph'] = {}
+            yaml['ceph']['storage'] = {}
+            yaml['ceph']['storage']['osds'] = {}
+            for osd in yaml['storage']['osds']:
+                yaml['ceph']['storage']['osds'][osd] = { 'format': 'bluestore' }
```
jschmid1: Although most of the time correct, we change the format to bluestore regardless of the user's choice. Same with the wal/db distrubution to the old journal.

srv/modules/runners/push.py
```diff
@@ -132,6 +152,56 @@ def output(self, common):
...
+    def rename(self, filename):
+        """
+        Copy the policy.cfg and update the existing file
+        """
+        shutil.copyfile(filename, "{}-original".format(filename))
+        lines = []
+        with open(filename, "r") as policy:
+            for line in policy:
+                if line.startswith('profile-'):
+                    line = "migrated-{}".format(line)
```
jschmid1: That would end up as:

`migrated-profile-default/` right?

A bit confusion when you were always used to scan for profile-*. Maybe change it to `profile-XXXX-migrated` ?

jschmid1:
code works as intended with storage:osds and storage:osd+journals, lgtm!

srv/modules/runners/push.py
swiftgist: I thought this would make it stand out a bit.  Also, the logic is simple with prepending a string.

jschmid1: Don't have a strong opinion on it. I just thought I'd share my initial thoughts. We can leave it as is.

swiftgist:
Updated Makefile, spec file and added a log message.

PR #415, opened 2017-07-06
disable openattic-rpcd

shundezhang:
service openattic-rpcd doesn't exist


swiftgist:
I do not recall exactly the version of openATTIC where the rpcd has been removed but I believe it's still necessary in the 2.X series.  

Normally, I would suggest just deleting the incorrect state, but since users may be on different versions of openATTIC for this branch of DeepSea, the commenting is fine.

PR #416, opened 2017-07-06
fixed check method of disk activation

shundezhang:
should use name sdX (like the data-only type) to check if a disk is already built and mounted for data+journal type


swiftgist:
I believe this can behave differently on fresh installations vs. existing installations.  It's the second time I've seen this request, so I will merge it.

PR #419, opened 2017-07-06
Alternative stage 0 in speparate files

jschmid1:
An alternative solution to https://github.com/SUSE/DeepSea/pull/387 based off of https://github.com/SUSE/DeepSea/issues/368


smithfarm:
@jschmid1 @swiftgist Does this change the behavior of `salt-run state.orch ceph.stage.0` ?

jschmid1:
@smithfarm no, those are just alternatives - the default is untouched

PR #421, opened 2017-07-06
iscsi: handle iscsi configure when no demo rbd image exists

rjfd:
This PR addresses the issue of configuring an iSCSI target when no RBD demo image was created.

This PR is related to #401 


swiftgist:
I'll worry about the cat to grep later. :)

PR #430, opened 2017-07-10
Moving from "auto" to "sharedsecret" salt-api eauth system

rjfd:
This PR changes the default salt-api authentication system (eauth) to use the "sharedsecret" eauth.

Besides the eauth change, this PR also changes DeepSea to install and configure salt-api upon DeepSea installation.


swiftgist:
There's 3 different installations here, right?  One for the Makefile for the local developer, one for the rpm installation and one for Salt.  

Originally, I left the salt-api out of the rpm since 'auto' was not something that should be the default.  With sharedsecret as default, the admin would only use the Salt state to change authentication methods.  

If the admin did run the salt state, does changing the sharedsecret require a restart of the master and api?

rjfd:
@swiftgist 

> With sharedsecret as default, the admin would only use the Salt state to change authentication methods.

Yes, maybe you're right, the salt-api state can be used as a way to change authentication methods. Currently it is always deploying the "sharedsecret" only. I can change this.

> If the admin did run the salt state, does changing the sharedsecret require a restart of the master and api?

In the  current salt-api state implementation, if a shared secret is already defined, the state does not change the secret. But yes, if the secret changes, then the salt-api daemon needs to be restarted.

swiftgist:
That's fine.  Just checking that I read things right. :)

PR #434, opened 2017-07-11
Added automatic configuration of openATTIC

rjfd:
This PRs adds to the openattic installation the automatic configuration of the salt-api credentials and the configuration of the grafana hostname.

### DIFF
--- a/srv/modules/runners/sharedsecret.py
+++ b/srv/modules/runners/sharedsecret.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+import os
+
+
+def show():
+    '''
+    This function reads the contents of the sharedsecret.conf file and parses the secret key.
+    This file has the following structure:
+
+    ``sharedsecret: <secret_key>``
+
+    Returns:
+        str: the secret key used by salt-api sharedsecret eauth
+
+    '''
+    filename = '/etc/salt/master.d/sharedsecret.conf'
+    if not os.path.exists(filename):
+        return None
+    with open(filename, 'r') as fhandler:
+        line = fhandler.readline()
+        idx = line.find(':')
+        if idx == -1:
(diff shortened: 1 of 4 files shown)


LenzGr:
LGTM, thanks!

srv/modules/runners/sharedsecret.py
```diff
@@ -0,0 +1,14 @@
...
+import os
+
+
+def show():
+    filename = '/etc/salt/master.d/sharedsecret.conf'
+    if not os.path.exists(filename):
+        return None
+    with open(filename, 'r') as fhandler:
+        line = fhandler.readline()
+        idx = line.find(':')
```
votdev: Maybe you should add a comment why it is necessary to parse and split the line. I don't know how the file  looks like, so an example would be helpful to understand the code.

srv/modules/runners/sharedsecret.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+import os
+
+
+def show():
+    filename = '/etc/salt/master.d/sharedsecret.conf'
```
votdev: I would like to suggest to name the file salt_api_secret.conf to make it clear for what this file is used.

rjfd: That I cannot do. This is an "internal" file expected by the sharedsecret eauth module.

srv/modules/runners/sharedsecret.py
rjfd: Yes, that would great. Thanks for the suggestion.

srv/modules/runners/sharedsecret.py
votdev: OK, i thought it is a user generated file.

PR #435, opened 2017-07-11
Change some debug levels from WARNING to INFO

LenzGr:
Reducing the noise as requested in bsc#1048157

Signed-off-by: Lenz Grimmer <lgrimmer@suse.com>

### DIFF
--- a/srv/modules/runners/minions.py
+++ b/srv/modules/runners/minions.py
@@ -57,9 +57,9 @@ def ready(**kwargs):
             expected = set(key.list_keys()['minions'])
 
         if actual == expected:
-            log.warn("All minions are ready")
+            log.info("All minions are ready")
             break
-        log.warn("Waiting on {}".format(",".join(list(expected - actual))))
+        log.info("Waiting on {}".format(",".join(list(expected - actual))))
         if end_time:
             if end_time < time.time():
                 log.warn("Timeout reached")


swiftgist:
The issue, though, is that Salt is set to WARNING by default.  This means the message will disappear for all users.  

Should we be changing the default level to INFO?  The lack of feedback for Salt commands is one of the largest criticisms I hear.  I would argue that I have put in debug messages that should really be info.

jan--f:
I'd also be in favour to keep this as is. Otherwise the default logging level would have to be INFO which adds quite a bit of noise on the console and in log files.

LenzGr:
My concern with keeping this "as is": these messages are prepended with the word "WARNING", which could be confusing for inexperienced users.

A message like "All minions are ready" should not be seen as a warning, in my opinion. I therefore agree with the suggestion made by Eric in the downstream bug report in that DeepSea should set the default log_level to INFO and we should balance the feedback by setting debug messages to the appropriate log level, e.d. debug/info. I'd be happy to take a stab at this, if you generally have no objections against this approach.

jan--f:
I'm not sure that will do the trick. Salt itself is quite noisy on the INFO level. That is why @swiftgist chose to go with this approach the is admittedly not the cleanest. There is just no other way of reporting status to the user.

LenzGr:
Oh, good to know. Thanks for the insight. In that case, feel free to decline the PR (and we should close the downstream bug report accordingly)

PR #438, opened 2017-07-11
ui_rgw runner: Add "admin entry path" to the RGW urls info

rjfd:


swiftgist:
I'm not attached to the 'canned' name, but I do like having a runtime debug for resolving issues.  Otherwise, debugging whether the breakdown is in the salt-api configuration vs. the real backend becomes a lot more involved.

rjfd:
@swiftgist ok, no problem. I've removed the commit that deleted the canned stuff.

Issue #436, opened 2017-07-11
Ceph doesn't support mixed configuration - but we do.


jschmid1:
The code in [ceph-disk](https://github.com/ceph/ceph/blob/master/src/ceph-disk/ceph_disk/main.py#L1474 ) only checks for `-i 0 --allow/depends/wants-journal` so the answer of `allows_journal` depends on the format of OSD #0. We are not strictly forbidding the deployment of mixed configuration. 

There is one place where it could bite us though.. Bluestore -> Filestore migration (whoever that wants)
If you have now a #0 with bluestore, the next creation of a filestore with dedicated journal will fail.
`ceph-disk -v prepare --fs-type xfs --filestore  --data-dev --journal-dev /dev/vdb1 /dev/vdf2`

will be stopped by 

```
raise Error('journal specified but not allowed by osd backend')                            
ceph_disk.main.Error: Error: journal specified but not allowed by osd backend 
```

swiftgist:
Is this about somebody mixing journals and wal/db on the same device?  Or is this trying to place a journal on a device with a bluestore OSD?  

I think the failure is okay since the exception should stop the show and DeepSea is not constraining the functionality of ceph-disk.  

As far as real deployments, I could see one minion configured as entirely bluestore or entirely filestore.  Some site may wish to do a slow migration and only change one rack at a time.  Still, that won't trigger this.  

If somebody says they need this to work, I think asking for the use case is reasonable. Any fixes would naturally happen in ceph-disk itself.

jschmid1:
> Is this about somebody mixing journals and wal/db on the same device? Or is this trying to place a journal on a device with a bluestore OSD?

This is about mixing bluestore and filestore OSDs in general. ceph-disk prevents you from creating a dedicated-journal-filestore-osd if your OSD#0 happens to be a bluestore OSD.

see https://github.com/ceph/ceph/blob/master/src/ceph-disk/ceph_disk/main.py#L1474

jschmid1:
> I think the failure is okay since the exception should stop the show and DeepSea is not constraining the functionality of ceph-disk.

Unfortunately the error is not passed all the way up to the stage.3 orchestrate runner. It will 'just not deploy' and be silent. You'll only be able to track down the issue if you execute the ceph-disk command manually on the OSD node

swiftgist:
Then, how is my VM setup working?  I have the first four nodes running bluestore, the fifth running filestore with the new format and the last running filestore in the old format.

[...]

PR #440, opened 2017-07-11
Add validation for disk_format

jschmid1:
closes https://github.com/SUSE/DeepSea/issues/436

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -283,6 +283,29 @@ class Validate(object):
         else:
             self.passed['mgrs'] = "valid"
 
+    def formats(self):
+	"""
+	As ceph-disk does indirectly not support two types of disk formats
+	we intervene as early as possible
+        See https://github.com/SUSE/DeepSea/issues/436
+	"""
+        formats = []
+        for node in self.data.keys():
+            # we can only detect formats in the new format
+            # the old will always be homogenous
+            if ('ceph' in self.data[node] and
+                'storage' in self.data[node]['ceph']):
+                hosts = self.data[node]['ceph']['storage']['osds']
+                for device in hosts.keys():
+                    if 'format' in hosts[device]:
+                       formats.append(hosts[device]['format'])
+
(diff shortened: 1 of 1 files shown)


swiftgist:
Won't this prevent different minions from having different formats?

jschmid1:
True, we should allow this.
Updating..

srv/modules/runners/validate.py
```diff
@@ -283,6 +283,30 @@ def mgrs(self):
...
+	"""
+        formats = {}
+        for node in self.data.keys():
+            # we can detect disk-formats only in the new yaml format
+            if ('ceph' in self.data[node] and
+                'storage' in self.data[node]['ceph']):
+                devices = self.data[node]['ceph']['storage']['osds']
+                formats[node] = []
+                for dev_name in devices.keys():
+                    if 'format' in devices[dev_name]:
```
jan--f: if ```format``` is not a key in the profile _I think_ we default to bluestore. Maybe we should add an else branch here?

srv/modules/runners/validate.py
```diff
@@ -283,6 +283,30 @@ def mgrs(self):
...
+                       formats[node].append(devices[dev_name]['format'])
+        for node, df in formats.iteritems():
+	    if len(set(formats)) > 1:
```
jan--f: Is ```formats``` correct here? Shouldn't ```len(set(df))``` be checked?

srv/modules/runners/validate.py
```diff
@@ -283,6 +283,30 @@ def mgrs(self):
...
+		msg = "Please use uniform Disk Formats across nodes. Either Bluestore or Filestore"
```
jan--f: To me this sounds like "One format across all nodes". How about ```Ceph only supports on OSD backend per host. Please use either Filestore or Bluestore for OSDs on any one host.```

jan--f: @smithfarm Could you lend your professional/nativ-speaker eye please?

PR #452, opened 2017-07-13
Disable upgrade if MON/OSD are on the same node (temporarily)

jschmid1:
Due to the current situation you have to upgrade    
    all monitos before ceph allows you to start any OSD                 
    Our current implementation of maintenance upgrades                  
    triggers this behavior if you happen to have                        
    Monitors and Storage roles assigned on the same node                
    (And more then one monitor)  
    To avoid this, before actually providing a proper solution,         
    we stop users to execute the upgade in the first place.

### DIFF
--- a/srv/modules/runners/upgrade.py
+++ b/srv/modules/runners/upgrade.py
@@ -1,0 +1,42 @@
+import salt.client
+import salt.utils.error
+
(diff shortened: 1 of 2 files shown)


jschmid1:
Closed in favor of a upcomming PR

PR #456, opened 2017-07-14
build/ops: require python-netaddr unconditionally

smithfarm:
python-netaddr is part of the POOL repo in both SLE-12-SP2 and SLE-12-SP3, so
there's no danger.

### DIFF
--- a/deepsea.spec
+++ b/deepsea.spec
@@ -33,9 +33,7 @@ Requires:       salt-master
 Requires:       salt-minion
 Requires:       salt-api
 Requires:       python-ipaddress
-%if 0%{?sle_version} == 120200 && 0%{?is_opensuse} == 1
 Requires:       python-netaddr
-%endif
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildArch:      noarch
 


smithfarm:
See https://scc.suse.com/packages?package=python-netaddr

smithfarm:
Testing now that this works with SP3 as well (since SP3 is not officially in SCC yet)

smithfarm:
Why extend the conditional, though? Which version of SLE/openSUSE is missing this package?

jan--f:
iirc opensuse does not need it as a dependency.

smithfarm:
Test build without the conditional: https://build.opensuse.org/package/show/home:smithfarm:branches:filesystems:ceph/deepsea

smithfarm:
Build succeeds for all targets.

smithfarm:
ah, but this is a runtime dependency :-( need to actually fire up a Leap 42.2 system

smithfarm:
Nah, it's there in 42.2: https://build.opensuse.org/package/show/openSUSE:Leap:42.2/python-netaddr

smithfarm:
@jan--f Do you know *why* openSUSE doesn't need it as a dependency? 

Also, what damage would be done if an unneeded dependency were installed in some cases?

smithfarm:
If you prefer, I can change the conditional to:

```
%if ! 0%{?is_opensuse}
Requires: python-netaddr
%endif
```

That will work for me.

smithfarm:
see https://github.com/SUSE/DeepSea/issues/458

jan--f:
ok I dug out the history. The dependency way only needed for 42.2, but since the ```?leap_version``` macro didn't work reliably I changed it to what it is now.
So I guess that means we 42.2 isn't the only system we need this dependency anymore. I guess we can just require it everywhere.

Issue #431, opened 2017-07-10
add an rbd text exporter to admin node


jan--f:

jan--f:
related openattic issue: https://tracker.openattic.org/browse/OP-2435

jan--f:
PR #460 implements this.

PR #460, opened 2017-07-14
add smartmon and rbd monitoring

jan--f:
Furthermore this PR restructures the dashboard deployment. Dashboards are now added via the grafana http api instead of pointing grafana at a json file. This means that the url of all dashboards will change.


jschmid1:
This time I'd actually like to see the commits squashed to keep things a bit more organized.

At a first glance I noticed longer files like `files/ceph-cluster.json` which I don't quite understand. Are these configration files for grafana? I also wonder likely they are to change/be customized?

jan--f:
hmm ok I'll how I can make it a bit more organized.
The json file are the dashboard specifications for grafana. So far they are not meant to be customized.

jan--f:
Fixes: #431

Issue #450, opened 2017-07-13
there are no ceph-mgrs after an upgrade


jschmid1:
Due to the latest changes in ceph we don't get a ceph-mgr alongside the mons automatically.

We need to adapt (mostly documentation)
* after the upgrade run stage1/2
* Add ceph-mgr role 
* run stage 3
* make sure you have ceph-mgrs running
* set `ceph osd require-osd-release luminous`

PR #465, opened 2017-07-17
Adjust upgrade strategy to latest upstream changes

jschmid1:
These commits respect some recent changes also:

  * ceph-mgr is now mandatory (adapt keyrings)
  * only restart mons first
  * set luminous osds
  * add a wait to allow ceph to catch up after a reboot
  * add preflight checks
  * make zypper-migration default
  * fix a sync bug

In addition we have keyring changes that needs to be reviewed please.
keyring files I haven't touched yet are: ganesha, benchmark. Could you please confirm that mgr capabilities have to be adjusted for them aswell?


srv/modules/runners/orderednodes.py
```diff
@@ -15,6 +15,16 @@ def _preserve_order_sorted(seq):
     seen = set()
     return [x for x in seq if x not in seen and not seen.add(x)]
 
+def only(cluster='ceph', role=None):
```
jan--f: Is the purpose of this function to get a list of all minions in the cluster?
If so we already have the select runner. A list of minions can be aquired using the ```select.minions``` runner call.

srv/salt/ceph/mds/files/keyring.j2
```diff
@@ -1,6 +1,7 @@
 [{{ client }}]
         key = {{ secret }}
-        caps mds = "allow"
-        caps mon = "allow profile mds"
-        caps osd = "allow rwx"
+        caps mds = "allow *"
```
jan--f: Please don't re-add the * for the mds caps.

srv/salt/ceph/wait/until/expired/30sec/init.sls
```diff
@@ -0,0 +1,3 @@
+include:
+  - .{{ salt['pillar.get']('wait_until_ok', 'default') }}
```
jan--f: This clashes with ```srv/salt/ceph/wait/until/OK/init.sls```. Is that intentional ?

srv/modules/runners/orderednodes.py
jan--f: Sorry of course the function can return a list of a certain role. Though I believe the select runner can do this as well via ```select.attr```?

srv/salt/ceph/mds/files/keyring.j2
jschmid1: Mh, we had the "allow *" caps in [SES4](https://github.com/SUSE/DeepSea/blob/SES4/srv/salt/ceph/mds/files/keyring.j2), which will not be refreshed if you upgrade to SES5.

```
ceph auth get mds.<mdsnode>:
        caps mds = "allow *"
        caps mgr = "allow profile mds"
        caps mon = "allow profile mds"
        caps osd = "allow rwx"
```
If you do a `ceph auth add mds.data1 -i /srv/salt/ceph/mds/cache/data1.keyring` if will fail with:

` Error EINVAL: entity mds.data1 exists but cap mds does not match`

Do we have to remove/reauth in order to get it right?

srv/salt/ceph/wait/until/expired/30sec/init.sls
jschmid1: good catch, c&p mistakes

srv/modules/runners/orderednodes.py
jschmid1: Right, I've missed that.

`salt-run select.minions roles=mon` would've worked too.

Updating

srv/salt/ceph/mds/files/keyring.j2
jan--f: Hmm true. And that issue doesn't exist when adding the mgr caps?

jschmid1: Well, it doesn't exist if the provided keyring *exactly* matches the previous keyring(including the *)

jschmid1:
closes #450

Issue #397, opened 2017-06-30
NFS-Ganesha RGW: check in ganesha.conf.j2 if bucket creation succeded. 


supriti:
In ganesha.conf.j2 (https://github.com/SUSE/DeepSea/blob/master/srv/salt/ceph/ganesha/files/ganesha.conf.j2) we are adding "Path = "{{ user }}-demo";" But it may happen that for some reason bucket creation failed. Deploying ganesha with non existent bucket will result in nfs-ganesha failure:

"init_export_root :EXPORT :CRIT :Lookup failed on path, ExportId=2 Path=demo-demo FSAL_ERROR=(No such file or directory,2)"

We should add a check that bucket is actually present.

swiftgist:
Fixed in #682

PR #470, opened 2017-07-18
NFS-Ganesha: Check if bucket exists

supriti:

### DIFF
--- a/srv/salt/_modules/rgw.py
+++ b/srv/salt/_modules/rgw.py
@@ -8,7 +8,8 @@ import json
 import boto
 import boto.s3.connection
 import boto.exception
-
+import glob
+import re
 
 log = logging.getLogger(__name__)
 
@@ -43,7 +44,7 @@ def configuration(role):
             for rgw_config in  __pillar__['rgw_configurations'].keys():
                 if rgw_config in role:
                     return rgw_config
-    return 
+    return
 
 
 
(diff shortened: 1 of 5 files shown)


supriti:
Fix for https://github.com/SUSE/DeepSea/issues/397

srv/salt/_modules/rgw.py
```diff
@@ -125,6 +125,16 @@ def create_bucket(**kwargs):
         return False
     return True
 
+def lookup_bucket(user, bucket):
+    s3conn = boto.connect_s3(
+        aws_access_key_id=access_key(user),
+        aws_access_key_id=secret_key(user),
+    )
+
```
theanalyst: we could make S3Conn as a class, and use the above create and lookup as methods?

srv/salt/_modules/rgw.py
```diff
@@ -125,6 +125,16 @@ def create_bucket(**kwargs):
         return False
     return True
 
+def lookup_bucket(user, bucket):
+    s3conn = boto.connect_s3(
```
rjfd: I think this connection will fail because you are not passing the host/port properties and this function will be executing in the master_minion where no rgw daemon is running.
You should connect the same way as in the above `create_bucket` function.

supriti: I would clean up as suggested by @theanalyst and make one single connection class and pass all the required parameters. Will update PR accordingly.

supriti:
Test is failing. Investing it. 

/home/travis/build/SUSE/DeepSea/srv/modules/runners/ui_rgw.py:91: TypeError: string indices must be integers, not str

supriti:
As the logic is moved from runner to module, tests are failing. One way to fix this would be to move the test for urls in test for rgw.py

supriti:
@swiftgist I guess you already worked on this PR. Should I close this one?

swiftgist:
@supriti I wanted to make sure you saw that your commits were in #682, but we can close this one now.

PR #471, opened 2017-07-18
Exception handling and None checks in osd+journal scenarios

jschmid1:
1)Don't use old style exceptions
2)Catch a bug with only osds+journal deployments

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -107,14 +107,16 @@ def configured(**kwargs):
         devices = __pillar__['ceph']['storage']['osds']
         devices = _filter_devices(devices, **kwargs)
         devices = devices.keys()
-    if 'storage' in __pillar__ and 'osds' in __pillar__['storage'] and 
-        type(__pillar__['storage']['osds']) is list:
+    if ('storage' in __pillar__ and
+        'osds' in __pillar__['storage'] and
+        type(__pillar__['storage']['osds']) is list):
         devices = __pillar__['storage']['osds']
         log.debug("devices: {}".format(devices))
         if 'format' in kwargs and kwargs['format'] != 'filestore':
             return []
-    if 'storage' in __pillar__ and 'data+journals' in __pillar__['storage'] and 
-        type(__pillar__['storage']['data+journals']) is list:
+    if ('storage' in __pillar__ and
+        'data+journals' in __pillar__['storage'] and
+        type(__pillar__['storage']['data+journals']) is list):
         [devices.append(x.keys()[0]) for x in __pillar__['storage']['data+journals']]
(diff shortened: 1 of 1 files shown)


srv/salt/_modules/osd.py
```diff
@@ -107,7 +107,7 @@ def configured(**kwargs):
         devices = __pillar__['ceph']['storage']['osds']
         devices = _filter_devices(devices, **kwargs)
         devices = devices.keys()
-    if 'storage' in __pillar__ and 'osds' in __pillar__['storage']:
+    if 'storage' in __pillar__ and 'osds' in __pillar__['storage'] and __pillar__['storage']['osds'] is not None:
```
jan--f: how about
```
if 'storage' in __pillar__ and 'osds' in __pillar__['storage'] and type(__pillar__['storage']['osds']) is list:
```
here. This way we catch other config errors too...say a string.

jan--f: And we should also add this test in line 115.

srv/salt/_modules/osd.py
```diff
@@ -107,12 +107,14 @@ def configured(**kwargs):
         devices = __pillar__['ceph']['storage']['osds']
         devices = _filter_devices(devices, **kwargs)
         devices = devices.keys()
-    if 'storage' in __pillar__ and 'osds' in __pillar__['storage'] and __pillar__['storage']['osds'] is not None:
+    if 'storage' in __pillar__ and 'osds' in __pillar__['storage'] and
```
jan--f: multi-line conditions need brackets, otherwise this is a syntax error. See https://stackoverflow.com/questions/181530/styling-multi-line-conditions-in-if-statements

srv/salt/_modules/osd.py
```diff
@@ -107,12 +107,14 @@ def configured(**kwargs):
...
         devices = devices.keys()
-    if 'storage' in __pillar__ and 'osds' in __pillar__['storage'] and __pillar__['storage']['osds'] is not None:
+    if 'storage' in __pillar__ and 'osds' in __pillar__['storage'] and 
+        type(__pillar__['storage']['osds']) is list:
         devices = __pillar__['storage']['osds']
         log.debug("devices: {}".format(devices))
         if 'format' in kwargs and kwargs['format'] != 'filestore':
             return []
-    if 'storage' in __pillar__ and 'data+journals' in __pillar__['storage']:
+    if 'storage' in __pillar__ and 'data+journals' in __pillar__['storage'] and
```
jan--f: See commet on line 110

srv/salt/_modules/osd.py
jschmid1: yikes, thanks

Issue #472, opened 2017-07-18
We do not handle machines with more than 26 disks


jschmid1:
On slightly bigger machines the disk shortname identifier pattern change after /dev/sdz is reached.
It continues by starting over again and and enumerating on the second digit.

>  /dev/sdz is followed by /dev/sdaa followed by /dev/sdab  

I know of at least *one* place in the code where we don't take the different pattern into account.

1) https://github.com/SUSE/DeepSea/blob/c81b3826869aa26ad2463cc866721eb4dbf2294d/srv/salt/_modules/osd.py#L889

Extend the list as we find more.

edit: formatting

jschmid1:
unittests for that would be great

jschmid1:
closed via #473

PR #473, opened 2017-07-18
dont wildly glob in order to respect longer device names

jschmid1:
tackles #472 

The failing test is there to prove our mistake. Need to investigate further to see if there are more kinks.

### DIFF
--- a/tests/unit/_modules/test_osd.py
+++ b/tests/unit/_modules/test_osd.py
@@ -1008,7 +1008,7 @@ class TestOSDPartitions():
         osd_config = OSDConfig()
         obj = osdp_o(osd_config)
         ret = obj._last_partition(osd_config.device)
-        glob_mock.glob.assert_called_with('/dev/sdx?*')
+        glob_mock.glob.assert_called_with('/dev/sdx[0-9]')
         assert type(ret) is int
 
     @mock.patch('srv.salt._modules.osd.glob')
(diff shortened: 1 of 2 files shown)


tests/unit/_modules/test_osd.py
```diff
@@ -1116,6 +1116,62 @@ def test_highest_partition(self, glob_mock, part_mock, osdc_o):
...
+        """
+        Given there is a device
+        And the device has partitions
+        And is_partition() returns True
+        And the device is not a nvme
+        And the device is the 27th so will be /dev/sdaa
+        Expect to return 2
+        """
+        kwargs = {'device': '/dev/sdaa'}
+        glob_mock.glob.return_value = ['/dev/sdaa']
```
jan--f: This is wrong here ``` glob.glob("{}[0-9]".format('/dev/sdaa'))``` will return ```[]``` not just the device.

jan--f:
hmm I'm not convinced this actually changes the behaviour of the code. I think the glob was ok...even better since it also catches partition numbers > 9

jschmid1:
> hmm I'm not convinced this actually changes the behaviour of the code. I think the glob was ok...even better since it also catches partition numbers > 9

discussed offline. Bottom line is -> handle NVME and standard devices separately, due to the 'p' suffix in the NVME partitioning scheme. Also adjust the regex to `[0-9]*`  rather than `[0-9]+` or even `[0-9]{1,10}` as `glob` isn't extremely rich in processing regex.

Issue #476, opened 2017-07-19
After stage 3, admin keyring has insufficient mgr caps


smithfarm:
I noticed that, after deploying latest upstream master with latest DeepSea master, I cannot do `ceph pg dump`:

<pre>
target149202175080:/home/ubuntu # ceph pg dump
Error EACCES: access denied' does your client key have mgr caps? See http://docs.ceph.com/docs/master/mgr/administrator/#client-authentication
</pre>

Pretty odd, given that the admin keyring appears to have the mgr caps:

<pre>
target149202175080:/home/ubuntu # cat /etc/ceph/ceph.client.admin.keyring
[client.admin]
        key = AQDwfW9ZAAAAABAATcA/JrAy7PAz7k+biXIQ3A==
        caps mds = "allow *"
        caps mon = "allow *"
        caps osd = "allow *"
        caps mgr = "allow *"
</pre>

Yet, it doesn't *really* have the caps:

<pre>
target149202175089:/home/ubuntu # ceph auth get client.admin
exported keyring for client.admin
[client.admin]
        key = AQDwfW9ZAAAAABAATcA/JrAy7PAz7k+biXIQ3A==
        caps mds = "allow *"
        caps mgr = "allow profile mds"
        caps mon = "allow *"
        caps osd = "allow *"
</pre>

Somehow, DeepSea is creating the `/etc/ceph/ceph.client.admin.keyring` correctly, but not importing it. After manual import, the caps are fine and `ceph pg dump` works.

@jschmid1 @jan--f @theanalyst

smithfarm:
This state is observable after stage 3 - for log, see http://147.135.133.225:8081/ubuntu-2017-07-19_19:58:08-deepsea:basic:health-ok-ses5---basic-openstack/

smithfarm:
```
2017-07-19T20:06:48.868 INFO:teuthology.orchestra.run.target217182142067.stderr:+ ceph auth get client.admin
...
2017-07-19T20:06:49.075 INFO:teuthology.orchestra.run.target217182142067.stderr:exported keyring for client.admin
2017-07-19T20:06:49.076 INFO:teuthology.orchestra.run.target217182142067.stdout:[client.admin]
2017-07-19T20:06:49.076 INFO:teuthology.orchestra.run.target217182142067.stdout:	key = AQBMu29ZAAAAABAA25R0hmhhleV2XXt3jIxm1A==
[...]

Issue #468, opened 2017-07-17
"mount error 2 = No such file or directory" after running Stage 4 with mds roles


smithfarm:
`qa/suites/basic/health-mds.sh` fails. Stage 4 finishes successfully, HEALTH_OK, all OSDs present, but the fs fails to mount:

```
2017-07-17T17:21:09.079 INFO:teuthology.orchestra.run.target213032077223.stdout:    + trap 'echo "Result: NOT_OK"' ERR
2017-07-17T17:21:09.079 INFO:teuthology.orchestra.run.target213032077223.stdout:    ++ whoami
2017-07-17T17:21:09.079 INFO:teuthology.orchestra.run.target213032077223.stdout:    ++ hostname --fqdn
2017-07-17T17:21:09.080 INFO:teuthology.orchestra.run.target213032077223.stdout:    + echo 'cephfs mount test script running as root on target213032077232.teuthology'
2017-07-17T17:21:09.080 INFO:teuthology.orchestra.run.target213032077223.stdout:    cephfs mount test script running as root on target213032077232.teuthology
2017-07-17T17:21:09.080 INFO:teuthology.orchestra.run.target213032077223.stdout:    ++ ceph-conf --lookup mon_initial_members
2017-07-17T17:21:09.080 INFO:teuthology.orchestra.run.target213032077223.stdout:    ++ tr -d '[:space:]'
2017-07-17T17:21:09.080 INFO:teuthology.orchestra.run.target213032077223.stdout:    + TESTMONS=target213032077223
2017-07-17T17:21:09.081 INFO:teuthology.orchestra.run.target213032077223.stdout:    ++ grep 'key =' /etc/ceph/ceph.client.admin.keyring
2017-07-17T17:21:09.081 INFO:teuthology.orchestra.run.target213032077223.stdout:    ++ awk '{print $NF}'
2017-07-17T17:21:09.081 INFO:teuthology.orchestra.run.target213032077223.stdout:    + TESTSECR=AQBm8WxZAAAAABAAnF5bxDvoTOlw1T1lD6oADA==
2017-07-17T17:21:09.081 INFO:teuthology.orchestra.run.target213032077223.stdout:    + echo 'MONs: target213032077223'
2017-07-17T17:21:09.081 INFO:teuthology.orchestra.run.target213032077223.stdout:    MONs: target213032077223
[...]

PR #477, opened 2017-07-19
Ensure /etc/ceph/ceph.client.admin.keyring matches the actual admin keyring caps

smithfarm:
Fixes: #468
Fixes: #476

### DIFF
--- a/qa/common/common.sh
+++ b/qa/common/common.sh
@@ -199,7 +199,7 @@ function cat_ceph_conf {
 function ceph_cluster_status {
   ceph osd tree
   ceph osd pool ls detail
-  ceph -s
+  ceph health detail -f json-pretty
 }
 
 
--- a/qa/common/common.sh
+++ b/qa/common/common.sh
@@ -204,9 +204,15 @@ function admin_auth_status {
 }
 
 function ceph_cluster_status {
-  ceph osd tree
-  ceph osd pool ls detail
+  ceph pg stat -f json-pretty
+  _grace_period 1
   ceph health detail -f json-pretty
+  _grace_period 1
+  ceph osd tree
+  _grace_period 1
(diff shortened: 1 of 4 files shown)


smithfarm:
Can you help me here? I'm trying to get it to rewrite `/etc/ceph/ceph.client.admin.keyring` with the output of `ceph auth export client.admin` . . . but something is missing.

smithfarm:
The problem I'm trying to solve is #476 - i.e., the *actual* `client.admin` (admin keyring) caps come from `srv/salt/ceph/mon/files/keyring.j2` and up until this PR those were different from what we are putting in `/etc/ceph/ceph.client.admin.keyring`.

tserong:
I assume we got into this problem in the first place because the mgr caps were updated correctly in srv/salt/ceph/admin/files/keyring.j2 by b39872d, but not in srv/salt/ceph/mon/files/keyring.j2, and the latter will have been what got pulled in by the mons during initial deployment?

The simplest fix for this is probably just to fix the caps in srv/salt/ceph/mon/files/keyring.j2, rather than also later exporting the keyring -- I WTF'd a couple of times at that convolution ;-) -- OTOH, doing that does mean there's only one authoritative place the admin key caps are set now (srv/salt/ceph/mon/files/keyring.j2)...

smithfarm:
@tserong Got it - the idea to export the keyring at the end of stage 3 has been voted down.

smithfarm:
Passes DS suite:

* http://147.135.133.225:8081/ubuntu-2017-07-20_09:17:00-deepsea:basic-ses5---basic-openstack
* http://147.135.133.225:8081/ubuntu-2017-07-20_09:50:58-deepsea-ses5---basic-openstack

PR #469, opened 2017-07-17
qa: add iSCSI gateway smoke test

smithfarm:
Note: merge https://github.com/SUSE/ceph/pull/136 when this is merged.

PR #478, opened 2017-07-20
mon keyring: append admin key

jan--f:
Instead of maintaining the admin caps twice, rather append the admin key
to the mon keyring from its source.

### DIFF
--- a/srv/salt/ceph/mon/key/default.sls
+++ b/srv/salt/ceph/mon/key/default.sls
@@ -1,10 +1,8 @@
 
-{% set admin_keyring = "/srv/salt/ceph/admin/cache/ceph.client.admin.keyring" %}
-
 {% set keyring_file = "/srv/salt/ceph/mon/cache/mon.keyring" %}
-{{ keyring_file}}:
+{{ keyring_file }}:
   file.managed:
-    - source: 
+    - source:
       - salt://ceph/mon/files/keyring.j2
     - template: jinja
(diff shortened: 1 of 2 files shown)


jan--f:
basic/health-ok succeeds in a vagrant run.

srv/salt/ceph/mon/key/default.sls
```diff
@@ -13,8 +11,9 @@
     - makedirs: True
     - context:
       mon_secret: {{ salt['keyring.secret'](keyring_file) }}
-      admin_secret: {{ salt['keyring.secret'](admin_keyring) }}
     - fire_event: True
 
-
-
+{{ keyring_file }} append admin keyring:
+  file.append:
```
theanalyst: does salt ensure that reruns will not append contents again?

jan--f: yes salt takes care of that.
https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html#salt.states.file.append

smithfarm:
@jan--f If you agree, I'll cherry-pick this into one or both of my outstanding PRs #469 and #477 - for testing purposes - in the end, we should have a single PR to merge.

jan--f:
@smithfarm sounds good

smithfarm:
The commit in this PR has been cherry-picked to #477 - let's merge it there.

PR #480, opened 2017-07-20
Failhard if a minion is missing

jschmid1:
During an upgrade we do clusterwide health checks:

* Are all minions up
* Are all services on each minion up
 
checks for service sanity lives in `ceph.processes` which calls the module `cephprocesses.wait`. It's supposed to get the role-minion relationship from the pillar. If a minion is down tough, it's timeout will be handles as a non-blocking(not part of the cluster), hence skipped.
To circumvent this behavior, and overall improve the reliability, wait for minions to be up(for 300seconds).
If all minions fail to reach back to the master -> FAIL(hard).

### DIFF
--- a/srv/salt/ceph/maintenance/upgrade/master/default.sls
+++ b/srv/salt/ceph/maintenance/upgrade/master/default.sls
@@ -5,6 +5,7 @@ pre master readycheck:
     - name: minions.ready
     - timeout: {{ salt['pillar.get']('ready_timeout', 300) }}
     - hardfail: True
+    - failhard: True
 
 {% set notice = salt['saltutil.runner']('advise.salt_upgrade') %}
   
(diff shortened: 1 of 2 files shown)


jschmid1:
I initially, still under the assumption that `ceph.processes` is aware of offline-hosts, put it there because I thought the failhard option wouldn't be sufficient so I added a complement `'hardfail'` which triggers a Exception when the `minion.ready` runner times out.
I'll retest without raising an Exception and only failhard and update the PR accordingly

jschmid1:
just tested this again.

Returning false while failhard is set is not sufficient to abort the execution. An exception needs to be raised to actually fail-hard.

Sidenote:
Raising an exception is not enough when failhard is not set.

swiftgist:
I think the failhard and hardfail is confusing just from the looks.  Would it be possible to check that failhard is there to raise the exception?

jschmid1:
it does not get passed to the module, unfortunately.
``` python
(Pdb) dir()
['actual', 'client', 'end_time', 'expected', 'key', 'kwargs', 'pdb', 'results', 'ret', 'settings']
(Pdb) kwargs
{'timeout': 300, 'hardfail': True}
(Pdb) 
```

swiftgist:
Changes in #503

PR #481, opened 2017-07-20
remove leading and trailing whitespace

jschmid1:
strip: 

> Return a copy of the string with the leading and trailing characters removed. The chars argument is a string specifying the set of characters to be removed. If omitted or None, the chars argument defaults to removing whitespace. The chars argument is not a prefix or suffix; rather, all combinations of its values are stripped:

as opposed to rstrip:

> Return a copy of the string with trailing characters removed. The chars argument is a string specifying the set of characters to be removed. If omitted or None, the chars argument defaults to removing whitespace. The chars argument is not a suffix; rather, all combinations of its values are stripped:

### DIFF
--- a/srv/modules/runners/push.py
+++ b/srv/modules/runners/push.py
@@ -284,7 +284,7 @@ class PillarData(object):
             for line in policy:
                 # strip comments from the end of the line
                 line = re.sub('\s+#.*$', '', line)
-                line = line.rstrip()
+                line = line.strip()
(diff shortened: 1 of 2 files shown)


jschmid1:
> Did someone actually have leading ws in their policy.cfg?
> 
> In any case would you mind adding test cases for trainling and leading withespaces with comments?

Discussed offline. The answer is yes.

PR #483, opened 2017-07-21
fix apparmor checks

kmroz:


srv/modules/runners/ready.py
```diff
@@ -72,7 +72,7 @@ def apparmor(self):
         """
         contents = self.local.cmd(self.search , 'cmd.shell', [ '/usr/sbin/apparmor_status --enabled 2>/dev/null; echo $?' ], expr_form="compound")
         for minion in contents:
-            if contents[minion] == 0:
+            if contents[minion] == '0':
```
jschmid1: what about casting `content[minion]` to str() or int()? this would increase durability when salt decides to change it's return values..

kmroz: I get the feeling you've been bitten before, Joshua :)

Sure, I'll update the PR.  Thanks for the input.

PR #489, opened 2017-07-27
Wip validate

swiftgist:
finish refactor, add openattic check
    

swiftgist:
The extraneous 'True' lines have always annoyed me.  Forgot why they are there.  It would be nice to not print them if we knew the admin ran the command directly.  That can wait though.

PR #491, opened 2017-07-27
Icinga is no longer a dependency for openATTIC

swiftgist:


swiftgist:
I am thinking /srv/salt/ceph/rescind/openattic/default.sls needs to split into two default files.  I did not immediately see that openattic.sls-example is driving the rescind/openattic logic.

I think the rgw.sls-example needs to go away as well.  Originally, the rgw.sls seemed like a natural idea, but then the endless includes make docs more confusing compared to just adding it commented (or enabled) in the global.yml or cluster.yml.

Think I need another PR.

swiftgist:
added #500 to address the icinga entry in openattic.sls-example

PR #494, opened 2017-07-27
small prototype to help users obtain rgw tokens

kmroz:


kmroz:
@capri1989

screen shot 2017-08-03 at 4 07 09 pm

PR #496, opened 2017-07-31
Add Prometheus exporter for Ceph RGW user bandwidth stats

votdev:
* Implement a Prometheus exporter for RGW statistics. The exporter is running on port 9156 by default. Additional RGW stats can be retrieved by this exporter in future, e.g. bucket stats.
* Add systemd service unit to manage Prometheus Ceph RGW exporter. The command 'systemctl daemon-reload' is executed via service.systemctl_reload after the service unit has been installed.
* Add Grafana 'Ceph - Object Gateway users' dashboard
* The Prometheus exporter is based on Python. The required Python package prometheus_client is installed via [python-prometheus-client](https://build.suse.de/package/show/SUSE:SLE-12-SP3:Update:Products:SES5/python-prometheus-client).
* The systemd unit and Prometheus exporter will be deinstaller on the minions when the RGW role is removed.
* Restart prometheus if /etc/prometheus/prometheus.yml has been changed

Open issues:
* Should we always install the Grafana 'Ceph - Object Gateway users' dashboard, or only if there is at least one RGW role?

This PR is related to https://tracker.openattic.org/browse/OP-2214.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -333,6 +333,8 @@ copy-files:
 	install -m 644 srv/salt/ceph/rescind/rgw/*.sls $(DESTDIR)/srv/salt/ceph/rescind/rgw/
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/rescind/rgw/keyring
 	install -m 644 srv/salt/ceph/rescind/rgw/keyring/*.sls $(DESTDIR)/srv/salt/ceph/rescind/rgw/keyring/
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/rescind/rgw/monitoring
+	install -m 644 srv/salt/ceph/rescind/rgw/monitoring/*.sls $(DESTDIR)/srv/salt/ceph/rescind/rgw/monitoring/
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/rescind/storage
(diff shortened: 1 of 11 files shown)


swiftgist:
@votdev Does installing the dashboard cause it to appear empty in Grafana when there's not an rgw role?  Also, will the sls file parse successfully with no targets if no rgw role is assigned?  See https://github.com/SUSE/DeepSea/pull/496/files#diff-971b95ba6355428345ae8e633db9a9c3R25

If Grafana and Salt fail gracefully, I think that's fine.

votdev:
@swiftgist Grafana shows an empty graph if no RGW role is assigned. Prometheus is running fine, either there are no ceph-rgw-exporter targets. The exporter is not listed at htpp://xyz:9090/targets. Finally i can not see any errors in Salt nor Grafana/Prometheus in case of no RGW role.

votdev:
Currently python-prometheus.client is installed via PIP but it will be replaced with a 'real' RPM package after the [OBS package request](https://build.opensuse.org/request/show/513687) has been approved.

votdev:
The Python package [python-prometheus-client](https://build.suse.de/package/show/SUSE:SLE-12-SP3:Update:Products:SES5/python-prometheus-client) is now in the IBS repo.

Issue #504, opened 2017-08-02
Error in stage 4: ceph osd pool application enable iscsi-images rbd


Exotelis:
After #499 was merged I get the following error when I run stage 4:

```
dev-thanatos-1.oa.suse.de_master:
----------
          ID: igw config
    Function: salt.state
      Result: False
     Comment: Run failed on minions: dev-thanatos-1.oa.suse.de
              Failures:
                  dev-thanatos-1.oa.suse.de:
                    Name: igw nop - Function: test.nop - Result: Clean Started: - 10:41:09.610280 Duration: 1.132 ms
                    Name: ceph osd pool create iscsi-images 128 - Function: cmd.run - Result: Clean Started: - 10:41:09.630292 Duration: 605.34 ms
                  ----------
                            ID: demo pool enable application
                      Function: cmd.run
                          Name: ceph osd pool application enable iscsi-images rbd
                        Result: False
                       Comment: Command "ceph osd pool application enable iscsi-images rbd" run
                       Started: 10:41:10.350337
                      Duration: 278.296 ms
                       Changes:
                                ----------
                                pid:
                                    27051
                                retcode:
                                    22
                                stderr:
                                    no valid command found; 10 closest matches:
                                    osd pool stats {<name>}
                                    osd pool set <poolname> size|min_size|crash_replay_interval|pg_num|pgp_num|crush_rule|hashpspool|nodelete|nopgchange|nosizechange|write_fadvise_dontneed|noscrub       |nodeep-scrub|hit_set_type|hit_set_period|hit_set_count|hit_set_fpp|use_gmt_hitset|target_max_bytes|target_max_objects|cache_target_dirty_ratio|cache_target_dirty_high_ratio|cache_target_full_rati       o|cache_min_flush_age|cache_min_evict_age|auid|min_read_recency_for_promote|min_write_recency_for_promote|fast_read|hit_set_grade_decay_rate|hit_set_search_last_n|scrub_min_interval|scrub_max_inte       rval|deep_scrub_interval|recovery_priority|recovery_op_priority|scrub_priority|compression_mode|compression_algorithm|compression_required_ratio|compression_max_blob_size|compression_min_blob_size       |csum_type|csum_min_block|csum_max_block|allow_ec_overwrites <val> {--yes-i-really-mean-it}
                                    osd pool set-quota <poolname> max_objects|max_bytes <val>
                                    osd pool get-quota <poolname>
[...]

PR #505, opened 2017-08-02
igw: do not fail Stage 4 if command not recognized

smithfarm:
"ceph osd pool application enable" is a very new command, added quite late
in the luminous development cycle.

There are two basic scenarios in which this command might get run:

1. installation of recent luminous cluster - command should simply succeed
2. upgrade of an older cluster to luminous - command might not be recognized

Each of these has two subscenarios:

a. iscsi-image pool exists
b. iscsi-image pool does not exist (unlikely, but theoretically possible)

Provided the command is recognized, "ceph osd pool application enable" is of
course idempotent. The problem arises when the command is not recognized by the
MON we are talking to. In that case, Stage 4 currently fails, which is not a
satisfactory outcome for such a trivial error.

By using "onlyif", we try to handle all these scenarios, but it might happen
(during a cluster upgrade) that this command runs before the command is
recognized. In that case, Stage 4 will complete without error, but the cluster
might end up that the iscsi-images pool does not get assigned to an
application. This should be taken care of in the upgrade code.

Fixes: #504

### DIFF
--- a/srv/salt/ceph/igw/config/default.sls
+++ b/srv/salt/ceph/igw/config/default.sls
@@ -12,7 +12,7 @@ demo pool:
 
 demo pool enable application:
   cmd.run:
-    - name: "ceph osd pool application enable iscsi-images rbd"
+    - name: "ceph osd pool application enable iscsi-images rbd || :"
     - fire_event: True
 
 demo image:
--- a/srv/salt/ceph/mds/pools/default.sls
+++ b/srv/salt/ceph/mds/pools/default.sls
@@ -14,12 +14,20 @@ cephfs data:
       - "rados lspools | grep -q cephfs_data"
       - "ceph fs ls | grep -q ^name"
 
+cephfs data pool enable application:
+  cmd.run:
+    - name: "ceph osd pool application enable cephfs cephfs_data || :"
+
 cephfs metadata:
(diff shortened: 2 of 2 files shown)


smithfarm:
@Exotelis Does this fix the issue for you?

Exotelis:
@smithfarm Yes. Thank you

smithfarm:
@Exotelis I changed it slightly; can you try it again?

Exotelis:
@smithfarm It's still working with your latest changes

smithfarm:
This passes the `basic/health-igw.sh` integration test case: http://149.202.189.249:8081/ubuntu-2017-08-02_11:20:28-deepsea:basic:igw-ses5---basic-openstack/

swiftgist:
Instead of running the command twice, the return can be ignored with

demo pool enable application:
  cmd.run:
    - name: "ceph osd pool application enable iscsi-images rbd || :"
    - fire_event: True

smithfarm:
@swiftgist OK, made the change.

smithfarm:
@swiftgist The cephfs pools caused the same HEALTH_WARN in the nfs-ganesha integration test - added a commit which should take care of it in a harmless and zero-risk way ;-)

PR #506, opened 2017-08-02
wip: rgw: add examples for deploying rgw with ssl

theanalyst:
use rgw_configurations to create a role rgw-ssl alongside rgw, and have
an easier helper function that would copy certificates in pem format in
salt://ceph/rgw/cert/rgw.pem to the rgw nodes


theanalyst:
needs a test case in qa/ as well

theanalyst:
cp rgw.j2 rgw-ssl.j2

PR #508, opened 2017-08-02
Enable salt-api

swiftgist:


smithfarm:
This PR introduced a regression (stage.0 hangs). Reproducibility: 100% (apparently).

Reverting this commit fixes the issue.

swiftgist:
Realized last night while testing openATTIC that most pages were failures due to the salt-api being down.  Thought we had addressed this, but found no reference.

I verified the stage would start the salt-api, but forgot about the salt-master restart.  Not sure how to address that issue at the moment.

smithfarm:
Yeah, in the minion log, I see the salt-master.service getting restarted about 2 seconds after the master emits its last log message and hangs.

swiftgist:
This is reverted in #512 now.  New packages incoming.

PR #526, opened 2017-08-10
ceph_tgt support

swiftgist:
New mandatory variable to constrain the scope of minions considered for Ceph.  

The man page has details.  The /srv/pillar/ceph/ceph_tgt.sls has examples.  Note that 'make install' will still default to all minions.


jan--f:
I think adding the ```ceph_tgt``` validation to stage 1 would be helpful. Otherwise couldn't test yet due to a broken setup. _Looks_ good.

Issue #529, opened 2017-08-10
Allow running custom roles in a single calls to various stages


theanalyst:
Currently if a custom role for eg rgw configurations is set in global.yml before stage1, populate will not yet pick this up because a pillar.refresh is run at stage2, however since we need this info to create the directories and discover, it would be useful to have a refresh at stage1

PR #530, opened 2017-08-10
stage1: try doing a pillar refresh before we push proposals

theanalyst:
Doing a pillar refresh before we discover profiles, this can help cases
where a global/cluster.yml already contains a custom role and we need
populate to use this role so that we can use the role in stage2's
policy.cfg, an existing example would be using rgw_configurations and
creating the rgw-ssl role in global.yml before stage1 which would not be
populated unless a pillar.refresh is done

Fixes: #529

### DIFF
--- a/srv/salt/ceph/stage/discovery/default.sls
+++ b/srv/salt/ceph/stage/discovery/default.sls
@@ -4,9 +4,17 @@ ready:
     - name: minions.ready
     - timeout: {{ salt['pillar.get']('ready_timeout', 300) }}
 
+refresh_pillar0:
+  salt.state:
+    - tgt: {{ salt['pillar.get']('master_minion') }}
+    - sls: ceph.refresh
+
 discover roles:
   salt.runner:
     - name: populate.proposals
+    - require:
+        - salt: refresh_pillar0
+
 
 discover storage profiles:
   salt.runner:
--- a/srv/salt/ceph/stage/discovery/default.sls
+++ b/srv/salt/ceph/stage/discovery/default.sls
@@ -19,3 +19,5 @@ discover roles:
 discover storage profiles:
   salt.runner:
     - name: proposal.populate
+    - require:
+        - salt: refresh_pillar0


srv/salt/ceph/stage/discovery/default.sls
```diff
@@ -4,9 +4,17 @@ ready:
     - name: minions.ready
     - timeout: {{ salt['pillar.get']('ready_timeout', 300) }}
 
+refresh_pillar0:
+  salt.state:
+    - tgt: {{ salt['pillar.get']('master_minion') }}
```
jan--f: You only want to refresh the pillar on the master, not on all minions?

theanalyst: yea, since at this stage only the global/cluster.yml would be populated, everything else would be during/after stage2

theanalyst: would you need pillar data from other minions during your benchmarking runs... as far as custom roles go since the customizations go in the global/cluster yamls I think it is okay if we just do a refresh in master minion, thoughts?

swiftgist: The refresh is for the Stage 1 runner since it can generate roles (e.g. custom rgw and ganesha roles) from the pillar data.  Stage 2 will refresh all minions.

srv/salt/ceph/stage/discovery/default.sls
```diff
@@ -4,9 +4,17 @@ ready:
...
+
 discover roles:
   salt.runner:
     - name: populate.proposals
+    - require:
+        - salt: refresh_pillar0
+
 
 discover storage profiles:
   salt.runner:
```
theanalyst: I guess a `require: salt: refresh_pillar0` fits here as well? @jan--f @swiftgist

theanalyst: Basically ensuring that we refresh the pillar first before we push out / generate proposals

theanalyst:
@jan--f @swiftgist is this ok to go? It has passed local testing and a teuthology run while using rgw custom roles

PR #535, opened 2017-08-10
Enable salt-api, add validation

swiftgist:


swiftgist:
Dependent on #526 currently.  Checks should succeed once it's merged.

PR #538, opened 2017-08-11
Make install: set salt-master to minion_id from /etc/salt

theanalyst:
Currently we set the master minion_id as fqdn, in vagrant systems and
such if a machine has multiple host entries, for different interfaces
the mininon_id may not match what hostname -f reports. Since we assume
salt-master and salt-minion is installed anyway for make install, we can
safely assume that `/etc/salt/minion_id` is already populated


theanalyst:
@swiftgist @jan--f no strong opinions on this one, but this is just an irk of mine whenever i setup and teardown my setup many many times and minion_id always needs to be replaced by hand again.

swiftgist:
This sounds suspiciously simple... trying to imagine when this wouldn't work.

PR #539, opened 2017-08-11
stage2: keep default-notimeout in sync with default

theanalyst:
Since we're only overriding one function call and the rest remains the
same, use includes and excludes to achieve the same goal and avoid the
need to update this file when newer functions are called in default

### DIFF
--- a/srv/salt/ceph/stage/discovery/default-notimeout.sls
+++ b/srv/salt/ceph/stage/discovery/default-notimeout.sls
@@ -1,9 +1,9 @@
 
-ready:
-  salt.runner:
-    - name: minions.ready
+include:
+  - .default
 
-discover:
-  salt.runner:
-    - name: populate.proposals
-
+extend:
+  ready:
+    salt.runner:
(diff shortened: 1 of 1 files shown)


theanalyst:
Changelog: Moved to extend in favour of exclude and recreating the function

theanalyst:
@jan--f yeah it is not a commonly used default, I had used it sometime during my vagrant setup

theanalyst:
@swiftgist what do you think of this?

swiftgist:
@theanalyst I understand it, but I was wondering if somebody else looking at it would.  

I agree that duplicate content in sls files is annoying, but with only using a single include, would the extends get complicated for other files?  I think I'm fine with this particular file, but should we push this type of change throughout?

theanalyst:
interesting, yeah in this case the since it was `default-notimeout` which was just default with a notimeout extend was a perfect candidate, but in general many other cases are not exactly default examples anymore in which case a full copy would be more understandable I guess

PR #541, opened 2017-08-11
qa: common: bail out if DeepSea not installed

smithfarm:

### DIFF
--- a/qa/common/common.sh
+++ b/qa/common/common.sh
@@ -8,10 +8,15 @@ source $BASEDIR/common/json.sh
 source $BASEDIR/common/rbd.sh
 source $BASEDIR/common/rgw.sh
 
-SALT_MASTER=$(cat /srv/pillar/ceph/master_minion.sls | \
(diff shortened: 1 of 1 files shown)


qa/common/common.sh
```diff
@@ -8,10 +8,15 @@ source $BASEDIR/common/json.sh
 source $BASEDIR/common/rbd.sh
 source $BASEDIR/common/rgw.sh
 
-SALT_MASTER=$(cat /srv/pillar/ceph/master_minion.sls | \
-             sed 's/.*master_minion:[[:blank:]]*\(\w\+\)[[:blank:]]*/\1/' | \
-             grep -v '^$')
-
+MASTER_MINION_SLS=/srv/pillar/ceph/master_minion.sls
```
kshtsk: Just a suggestion. Wouldn't it better to use more general deepsea related file object check for existence. For example, '/srv/pillar/ceph' or '/srv/pillar'?

smithfarm: We want the check to fail when DeepSea is not properly installed. Someone might have a `/srv/pillar/ceph` directory without a working DeepSea installation.

PR #552, opened 2017-08-16
Remove rgw-nfs role

supriti:


swiftgist:
Agreed, we are going with client-ganesha.

PR #553, opened 2017-08-16
ui_rgw: don't glob over all conf files searching for rgws

theanalyst:
This commit makes ui_rgw only search for rgw conf bits in filenames that
we deploy the configuration files for ie. all the members of
`rgw_configurations` or `rgw` if nothing is configured, otherwise the
current runner just overrides the value with the last filename that
contains rgw_frontends which may not be valid anymore as more conf files
are added but not actually deployed

### DIFF
--- a/srv/modules/runners/ui_rgw.py
+++ b/srv/modules/runners/ui_rgw.py
@@ -124,13 +124,16 @@ class Radosgw(object):
         admin_path = 'admin'
 
         rgw_names = ['rgw']
-        if 'roles' in __pillar__:
-            if 'rgw_configurations' in __pillar__:
-                rgw_names = __pillar__['rgw_configurations'].keys()
+        for minion in cached:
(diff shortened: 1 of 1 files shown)


srv/modules/runners/ui_rgw.py
```diff
@@ -122,7 +122,15 @@ def endpoints(cluster='ceph'):
         port = '7480'  # civetweb default port
         ssl = ''
         admin_path = 'admin'
-        for rgw_conf_file_path in glob.glob("/srv/salt/ceph/configuration/files/ceph.conf.*"):
+
+        rgw_names = ['rgw']
+        if 'roles' in __pillar__:
+            if 'rgw_configurations' in __pillar__:
```
theanalyst: so this doesn't work in a runner, `__pillar__` is always empty only works on salt modules, need to use local.cmd or the pillar util from above

Issue #548, opened 2017-08-15
rgw role no longer creating users - REGRESSION


smithfarm:
The "NFS-Ganesha with RGW FSAL" qa tests started failing after the latest flurry of merges. As far as I can tell, the documented way of having DeepSea create rgw users is no longer working:

```
2017-08-15T07:46:24.717 INFO:teuthology.orchestra.run.target149202172037.stdout:    + radosgw-admin user list
2017-08-15T07:46:24.717 INFO:teuthology.orchestra.run.target149202172037.stdout:    []
2017-08-15T07:46:24.717 INFO:teuthology.orchestra.run.target149202172037.stdout:    + radosgw-admin user info --uid=demo
2017-08-15T07:46:24.717 INFO:teuthology.orchestra.run.target149202172037.stdout:    could not fetch user info: no user info saved
2017-08-15T07:46:24.718 INFO:teuthology.orchestra.run.target149202172037.stdout:    ++ echo 'Result: NOT_OK'
```

Link to suites/basic run: http://149.202.189.249:8081/ubuntu-2017-08-15_07:30:18-deepsea:basic-ses5---basic-openstack/

smithfarm:
@swiftgist @theanalyst @kmroz

smithfarm:
added the word REGRESSION to the title to draw attention to what might become a serious issue

swiftgist:
@smithfarm I can take a look in more detail a little later today.  I suspect it's the removal of the rgw.sls from the init.sls.  This was something I originally created and somewhat regretted.  While I was addressing the openattic.sls file, I thought I should address the rgw.sls as well.

I think I only half did it. :)

PR #556, opened 2017-08-17
Re-added the loading of rgw.sls to the pillar/ceph/init.sls

rjfd:
This commit fixes a regression introduced in #0051a27

Fixes: #548 


smithfarm:
Thanks!

PR #560, opened 2017-08-18
prometheus: make rule_files a top level config setting

tserong:
rule_files needs to be at the top level, not a child of global,
or prometheus won't start (tested with prometheus v2.beta.1)


jan--f:
LGTM. Will merge this after I checked if we need more changes for prometheus v2.

tserong:
Thanks :-)

jan--f:
Seems like this was actually wrong all along but prometheus 1.x never complained.

PR #563, opened 2017-08-18
qa: nfs-ganesha: test both NFSv3 and NFSv4

smithfarm:


supriti:
showmount works only for nfs version 3. But as we are defining both NFS = 3,4, it shows us the showmount point. So it can be used only with version3 to verify if nfs-ganesha server is running successfully.

smithfarm:
I used to have a "showmount loop" kludge in the nfs-ganesha tests to work around the transient mount hang. I added a commit dropping this kludge and I'm now running the nfs-ganesha tests multiple times to see if it's still needed.

smithfarm:
@supriti I can't seem to reproduce the mount hang anymore, so I commented out the workaround.

smithfarm:
Done, I think. All three NFS-Ganesha test cases now pass.

Issue #531, opened 2017-08-10
rpm uninstall does not remove /srv/modules/../*.pyc


Martin-Weiss:
when uninstalling the deepsea RPM the *.pyc files are left over:

/srv/modules/
/srv/modules/pillar
/srv/modules/pillar/stack.pyc
/srv/modules/runners
/srv/modules/runners/select.pyc
/srv/modules/runners/advise.pyc
/srv/modules/runners/benchmark.pyc
/srv/modules/runners/cephops.pyc
/srv/modules/runners/cephprocesses.pyc
/srv/modules/runners/configure.pyc
/srv/modules/runners/disengage.pyc
/srv/modules/runners/filequeue.pyc
/srv/modules/runners/minions.pyc
/srv/modules/runners/net.pyc
/srv/modules/runners/orderednodes.pyc
/srv/modules/runners/ready.pyc
/srv/modules/runners/proposal.pyc
/srv/modules/runners/push.pyc
/srv/modules/runners/remove.pyc
/srv/modules/runners/rescinded.pyc
/srv/modules/runners/sharedsecret.pyc
/srv/modules/runners/status.pyc
/srv/modules/runners/ui_ganesha.pyc
/srv/modules/runners/ui_iscsi.pyc
/srv/modules/runners/ui_rgw.pyc
/srv/modules/runners/upgrade.pyc
/srv/modules/runners/validate.pyc
/srv/modules/runners/populate.pyc

Can we somehow add these to the rpm uninstall process?

smithfarm:
The RPM install process should compile the `.py` files and the deepsea package should own both the `.py` and the `.pyc` files. Then this won't happen.

swiftgist:
I cannot recall the exact issue, but I thought we already ran into a problem when shipping pyc files.  This may have been related to the openATTIC rpm.

smithfarm:
I peeked inside an arbitrary python module RPM (the one I looked at was `python-numpy`) and it contains both `.py` and `.pyc` files. The RPM *must* own the `.pyc` files; otherwise they will get left behind like Martin noted.

smithfarm:
Python modules are typically packaged using `setuptools`. When you do `python setup.py build` the `.pyc` files get built. When you do `python setup.py install` they get installed. DeepSea needs to do something similar (and openATTIC, too, if it ships `.py` files).

jan--f:
[...]

PR #568, opened 2017-08-21
Build and package a pyc file for every py file

smithfarm:
Fixes: https://github.com/SUSE/DeepSea/issues/531


smithfarm:
This has been tested in the sense that I had OBS build it, and verified that the resulting RPM contains the `.pyc` files.

theanalyst:
In the long run maybe we should make setup.py do these things since that is the normal way python packages are meant to be. This is fine for now

.gitignore
```diff
@@ -1,4 +1,3 @@
-*.py[co]
```
theanalyst: I still think we are not going to add these files to git, so should we keep the gitignore?

smithfarm: This line is a duplicate. Even after this change, `*.pyc` will continue to be ignored :-)

theanalyst: ah ok :)

PR #571, opened 2017-08-21
Only use ceph.client.storage.keyring when it exists

swiftgist:
Fix removal of OSDs during Stage 5
Adjust migration to use its own removal

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -375,15 +375,31 @@ class OSDWeight(object):
         log.debug("Timeout expired")
         raise RuntimeError("Timeout expired")
 
+def _settings(**kwargs):
+    """
+    Initialize settings to use the client.storage name and keyring
+    if the keyring is available (Only exists on storage nodes.)
+
+    Otherwise, rely on the default which is client.admin. See
+    rados.Rados.
+    """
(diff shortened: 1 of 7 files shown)


srv/salt/_modules/osd.py
```diff
@@ -379,11 +379,15 @@ def zero_weight(id, wait=True, **kwargs):
     """
     Set weight to zero and wait until PGs are moved
     """
-    settings = {
+    settings = {}
+    storage_keyring = '/etc/ceph/ceph.client.storage.keyring'
+    if os.path.exists(storage_keyring):
+        settings = {
             'keyring': '/etc/ceph/ceph.client.storage.keyring',
```
jan--f: Use ```storage_keyring``` variable here.

srv/salt/_modules/osd.py
```diff
@@ -379,11 +379,15 @@ def zero_weight(id, wait=True, **kwargs):
...
             'client': 'client.storage'
-    }
-    settings.update(kwargs)
+        }
+        settings.update(kwargs)
```
jan--f: Update settings with kwargs only if the the ```storage_keyring``` path exists? Not quite obvious why...maybe a comment?

PR #572, opened 2017-08-22
populate,cephinspector: improve running cluster network detection

kmroz:
OSDs can listen on public and internal cluster IPs.  Thus, when deriving
the cluster network during the engulf process, filter out public IPs.

Signed-off-by: Karol Mroz <kmroz@suse.de>


kmroz:
Marking dnm as needs a quick test.

PR #589, opened 2017-08-28
Fix regeneration of ceph.conf, keep states in Stage 3

swiftgist:
Signed-off-by: Eric Jackson <ejackson@suse.com>

PR #591, opened 2017-08-29
config: only specifiy one source to force salt to look for it

jan--f:
Otherwise this can cause failure to distribute the config file.


smithfarm:
I can confirm that this fixes the breakage I described here: https://github.com/SUSE/DeepSea/pull/589#issuecomment-325498887

PR #597, opened 2017-08-31
Fix rgw config generation in ceph.conf

swiftgist:
Also, allow admin to override snippets in ceph.conf.d.

### DIFF
--- a/srv/modules/runners/select.py
+++ b/srv/modules/runners/select.py
@@ -1,9 +1,13 @@
 # -*- coding: utf-8 -*-
 
 import salt.client
+import logging
 import pprint
 import os
 import sys
+import re
+
+log = logging.getLogger(__name__)
 
 def minions(host = False, **kwargs):
     """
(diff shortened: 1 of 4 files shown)


srv/modules/runners/select.py
```diff
@@ -103,3 +107,44 @@ def attr(host = False, **kwargs):
...
+            for minion in minion_list:
+                grains_result = local.cmd(minion , 'grains.item',  list(args)).values()[0]
+                small = [ role ]
+                for arg in list(args):
+                    small.append(grains_result[arg])
+                results.append(small)
+                
+    return results
+
+__func_alias__ = {
```
theanalyst: interesting

srv/salt/ceph/configuration/files/ceph.conf.j2
```diff
@@ -22,9 +22,13 @@ mon health preluminous compat warning = false
 
 {% include "ceph/configuration/files/ceph.conf.d/global.conf" ignore missing %}
 
-{% for config in salt['rgw.configurations']() %}
-{% set client = config + "." + grains['host'] %}
+{% for config,host,fqdn in salt.saltutil.runner('select.from', pillar='rgw_configurations', attr='host, fqdn') %}
+{% set client = config + "." + host %}
+{% if salt['file.file_exists']("/srv/salt/ceph/configuration/files/ceph.conf.d/ceph.conf." + config) %}
+{% include "ceph/configuration/files/ceph.conf.d/ceph.conf." + config %}
```
theanalyst: would ui_rgw.py need any change because of this?

PR #603, opened 2017-09-02
Bug fixing the sort order of iperf result which was not natural

AvengerMoJo:
Taking out the dict, because salt return sort by key no matter what I do. 
So I can only return list when which look not as good as dict, unless someone has a better idea. 

What I do was taking out the text and convert the int, then sort. Instead of doing natural sort with string which I need to install another library or implement something much bigger with lots of corner case.


AvengerMoJo:
Thanks @jan--f  the latest update has update the display order.

swiftgist:
Works for me.

PR #602, opened 2017-09-01
qa: add openattic-with-rgw deployment test

smithfarm:
For the time being, it only deploys openATTIC and RGW and checks
the RGW deployment for sanity. It does not do any validation of the
oA deployment.

Signed-off-by: Nathan Cutler <ncutler@suse.com>

PR #604, opened 2017-09-04
validate: no conflict when rgw and oA are not on same node

smithfarm:
Fixes https://bugzilla.suse.com/show_bug.cgi?id=1057011 for the case when rgw and openattic roles are not colocated.

I discussed the colocated case with a couple folks and the consensus seemed to be that we should not support it. In other words, Stage 4 should fail if we detect colocation of rgw and openattic roles in the validation.

The colocated case is still WIP in #602 but I split this commit off into a separate PR since it's a simple fix for the usual (non-colocated) case.

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -590,7 +590,8 @@ class Validate(object):
         local = salt.client.LocalClient()
         for node in self.data.keys():
             if ('roles' in self.data[node] and
-                'openattic' in self.data[node]['roles']):
+                'openattic' in self.data[node]['roles'] and
+                'rgw' in self.data[node]['roles']):
                 # Would use file.contains if it supported '='
                 result = local.cmd(node, 'file.contains_regex', [ '/etc/ceph/ceph.conf', 'port\=80'], expr_form="glob")
                 if result[node]:


smithfarm:
I guess the objection to this approach is "what if the user deploys rgw manually?"

theanalyst:
It is not as much as user deploys rgw manually, more of since the configuration files in the /srv/salt/ceph/configuration are expected to be user edited we need validate also to ensure that that is not the case then

smithfarm:
@theanalyst OK, but if "rgw" role and "openattic" role are not shared on a node, we don't need to validate anything, right?

theanalyst:
yes, Also, I'm not opposed to this idea, it is just that getting this working requires us to handle all of these cases

swiftgist:
@smithfarm I already created a validation to prevent oA and rgw from using the same port on the same node.  If some PoC has limited physical nodes but plenty of memory, I do not see why we should dismiss their configuration.

swiftgist:
Okay, I read the related bug and see that the recent changes to ceph.conf is putting the rgw.conf sections on all minions now. That's what is different.

smithfarm:
> I already created a validation to prevent oA and rgw from using the same port on the same node. If some PoC has limited physical nodes but plenty of memory, I do not see why we should dismiss their configuration.

Does DeepSea provide a way for the user to specify an arbitrary port (not just 80/443) for rgw and openattic to listen on?

theanalyst:
well right now for rgw, the only way is to modify the port value in ceph.conf.rgw

Martin-Weiss:
It seems that the conflict handling in modules/runners/validate.py does not work as required for something like this:

`rgw frontends = "civetweb port=8080+8443s ssl_certificate=/etc/ceph/rgw.pem"`

--> here a conflict is reported because the match on "port\=80" is valid.

Do we need to re-open this issue or do we need to create a new issue?

theanalyst:
@Martin-Weiss lets open a new issue, it seems our regex match matches anything with 80 which  is wrong

Martin-Weiss:
ok issue #652 created

Issue #314, opened 2017-06-09
default rgw system user missing


swiftgist:
We should really include the system user in the DeepSea configuration.  Otherwise, the admin must research/guess what to put where.  As a workaround, anyone can add the following to /srv/pillar/ceph/stack/ceph/cluster.yml:

```
rgw_configurations:
  rgw:
    users:
      - { uid: "admin", name: "Admin", email: "admin@demo.nil", system: True }
```

prior to running Stage 2.  (Or add this to the pillar and run Stages 2-4.)  A cached version of the system user will appear in /srv/salt/ceph/rgw/cache/user.admin.json.  The rgw salt module does this.  When openATTIC starts, the same credentials are retrieved by the ui_rgw runner.

The permanent entry needs to end up in /srv/pillar/ceph/stack/ceph/default/cluster.yml which comes from the populate runner.

theanalyst:
Just to be explicit here, a system user is not the same as the one who can access the RGW admin APIs, a system user is sort of a cluster super user, whereas what oA actually requires a user with caps to all the admin APIs

abonillasuse:
Please include this in SES5.

swiftgist:
@theanalyst What is needed for openATTIC?  How do we distinguish which user that is?  Currently, the logic searches for the system user and returns the credentials.

PR #607, opened 2017-09-07
Split users from rgw_configuration, add system user

swiftgist:
Users are not tied to configurations (or zones).  This has been inherently wrong.   Users can optionally be maintained under .../ceph/rgw/users/users.d with this change.  The /srv/pillar/ceph/rgw.sls is redundant with the structure under /srv/pillar/ceph/stack.  The rgw_configurations has become a simple array and can be maintained in either the global.yml or ceph/cluster.yml.

Fixes: https://github.com/SUSE/DeepSea/issues/314

### DIFF
--- a/srv/salt/_modules/rgw.py
+++ b/srv/salt/_modules/rgw.py
@@ -37,45 +37,41 @@ def configuration(role):
     the ganesha roles silver and silver-common will both return silver.
     """
     if role == 'ganesha':
-        role = 'rgw'
+        return 'rgw'
     if 'roles' in __pillar__:
         if 'rgw_configurations' in __pillar__:
-            for rgw_config in  __pillar__['rgw_configurations'].keys():
+            for rgw_config in  __pillar__['rgw_configurations']:
                 if rgw_config in role:
                     return rgw_config
-    return 
+    return
 
 
 
-def users(role):
+def users(realm='default'):
     """
-    Return the list of users.  Consider the default rgw and ganesha roles
-    equivalent.
+    Return the list of users for a realm.
     """
-    if 'roles' in __pillar__:
-        if 'rgw_configurations' in __pillar__:
-            if role == 'ganesha' or role == 'rgw':
-                # Special case for default names
-                users = [ u['uid'] for u in __pillar__['rgw_configurations']['rgw']['users'] ]
-                log.info("users: {}".format(users))
-                return users
-            if role in __pillar__['rgw_configurations']:
-                users = [ u['uid'] for u in __pillar__['rgw_configurations'][role]['users'] ]
-                log.info("users: {}".format(users))
-                return users
-        if 'rgw' in __pillar__['roles']:
-            return []
+    cmd = "radosgw-admin user list --rgw-realm={}".format(realm)
+    log.info("cmd: {}".format(cmd))
+    proc = Popen(cmd.split(), stdout=PIPE, stderr=PIPE)
+    proc.wait()
+    log.debug("rc: {}".format(proc.returncode))
+    if proc.returncode != '0':
+        return json.loads(proc.stdout.read())
     return []
 
-def add_users(pathname="/srv/salt/ceph/rgw/cache"):
+def add_users(pathname="/srv/salt/ceph/rgw/cache", jinja="/srv/salt/ceph/rgw/files/users.j2"):
     """
-    Write each user to its own file
+    Write each user to its own file.
     """
-    if 'rgw_configurations' not in __pillar__:
+    users = __salt__['slsutil.renderer'](jinja)
+    log.debug("users rendered: {}".format(users))
+
+    if users is None or 'realm' not in users:
         return
-    
-    for role in __pillar__['rgw_configurations']:
-        for user in __pillar__['rgw_configurations'][role]['users']:
+
+    for realm in users['realm']:
+        for user in users['realm'][realm]:
             if 'uid' not in user or 'name' not in user:
                 raise ValueError('ERROR: please specify both uid and name')
 
@@ -128,7 +124,7 @@ def create_bucket(**kwargs):
 
 def _key(user, field, pathname):
     """
-    Read the filename and return the key value.  
+    Read the filename and return the key value.
     """
     data = None
     filename = "{}/user.{}.json".format(pathname, user)
@@ -143,7 +139,7 @@ def _key(user, field, pathname):
 
 def access_key(user, pathname="/srv/salt/ceph/rgw/cache"):
     if not user:
-        raise ValueError("ERROR: no user specified") 
+        raise ValueError("ERROR: no user specified")
     return _key(user, 'access_key', pathname)
 
 def secret_key(user, pathname="/srv/salt/ceph/rgw/cache"):
--- a/srv/salt/ceph/rgw/files/users.j2
+++ b/srv/salt/ceph/rgw/files/users.j2
@@ -1,0 +1,12 @@
+
+realm:
+  default:
+{% if salt['file.file_exists']('/srv/salt/ceph/rgw/users/system.user.yml') %}
+{% include "ceph/rgw/users/system.user.yml" %}
+{% else %}
+{% include "ceph/rgw/files/system.user.yml" %}
+{% endif %}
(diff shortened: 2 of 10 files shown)


srv/salt/ceph/rgw/files/users.j2
```diff
@@ -0,0 +1,11 @@
+
+realm:
+  default:
+{% if salt['file.file_exists']('/srv/salt/ceph/rgw/users/system.user.yml') %}
+{% include "ceph/rgw/users/system.user.yml" %}
```
rjfd: Why do you want to support more than one location for the system user info?
It's better if `system.user.yml` is only allowed to be in a single location

srv/salt/ceph/rgw/users/users.d/README
```diff
@@ -0,0 +1,8 @@
+Initial Users
+
+The /srv/salt/ceph/rgw/files/users.j2 will include any rgw users created in
+this directory.  Add users to one or more files with a yml extension.  Each entry is indented six spaces to form a proper dictionary.
```
rjfd: I believe we are shooting ourselves in the foot with this: `Each entry is indented six spaces to form a proper dictionary.` 
It's so easy to miss this and by mistake get the wrong number of indentation spaces.

I believe the user should just declare a list of users with zero spaces of indentation, and in `/srv/salt/ceph/rgw/files/users.j2` we process this list in jinja and add the indentation spaces.
What do you think?

srv/salt/_modules/rgw.py
```diff
@@ -40,42 +40,37 @@ def configuration(role):
...
-    if 'rgw_configurations' not in __pillar__:
+    users = __salt__['slsutil.renderer'](jinja)
+
+    if users is None or 'realm' not in users:
         return
-    
-    for role in __pillar__['rgw_configurations']:
-        for user in __pillar__['rgw_configurations'][role]['users']:
+
+    for realm in users['realm']:
```
rjfd: From the structure of the file `/srv/salt/ceph/rgw/files/users.j2`, the value of the `realm` key is not a list but a dict, which means that you should change this to:
```
for realm, user_list in users['realm'].items():
    for user in user_list:
```

srv/salt/_modules/rgw.py
```diff
@@ -40,42 +40,37 @@ def configuration(role):
...
     return []
 
-def add_users(pathname="/srv/salt/ceph/rgw/cache"):
+def add_users(pathname="/srv/salt/ceph/rgw/cache", jinja="/srv/salt/ceph/rgw/files/users.j2"):
     """
-    Write each user to its own file
+    Write each user to its own file.
     """
-    if 'rgw_configurations' not in __pillar__:
+    users = __salt__['slsutil.renderer'](jinja)
```
rjfd: we need to be careful with this because this call will only succeed if this function is called in the master node minion. Otherwise slsutil.renderer will not find the file `/srv/salt/ceph/rgw/files/users.j2`

srv/salt/ceph/rgw/users/users.d/README
theanalyst: Agree, this would be something that would end up being set wrong, and hard to track down later

smithfarm:
If this changes `rgw_configurations` then I guess the following qa code will need to be adjusted? https://github.com/SUSE/DeepSea/blob/c467dc9693d0ea0758bcfecc52b602a7c45bfa40/qa/common/rgw.sh#L5-L15

srv/salt/ceph/rgw/files/users.j2
swiftgist: Because I do not want the admin modifying files supplied by the rpm.  Upgrades are bad enough in general.  

If the admin does not care for our name, etc., then all of their changes are under the rgw/users directory.

srv/salt/ceph/rgw/users/users.d/README
swiftgist: Will slsutil.renderer process it correctly?  I am not reading the file.  I am relying on Salt.

As far as tracking it down, we can have the module complain verify the yaml.  We do that during Stage 2 and it's not foolproof either.

Technically, this is no different than what was in the pillar.  Users can get the structure wrong there as well.

srv/salt/_modules/rgw.py
swiftgist: Many of our modules are in this category.  Much of it has to do with permissions as root or having a proper functioning return since orchestrations ignore runner return codes.

srv/salt/_modules/rgw.py
swiftgist: I am hardcoding default for the moment and did not plan to extend this yet.  My ulterior goal is to set us up for rgw multi-site without too much reorganization.

swiftgist:
@smithfarm It's there. :)

srv/salt/ceph/rgw/files/users.j2
rjfd: Makes sense, we just need to make sure that this is properly documented then.

srv/salt/ceph/rgw/users/users.d/README
rjfd: The idea is to instead of just include the user's file as in [srv/salt/ceph/rgw/files/users.j2:10](https://github.com/SUSE/DeepSea/pull/607/files#diff-c62df72fdd6c23b5f05fe33ba66bdcebR10),
process the file contents it a for loop and correctly create the indentation steps.

Something like this (not sure if it works, didn't test it):

```
{% for user in salt['sls.renderer'](file) %}
      - { uid: "{{ user['uid'] }}", name: "{{ user['name'] }}", email: "{{ user['email'] }}" }
{% endfor %}
```

swiftgist: More fields are allowed than just those.  And yaml supports more than one structure, so I wasn't trying to suggest/enforce the one liner per entry.

The double for loop could be an issue, but I can see if something silly will work (i.e. indent the include directive).

srv/salt/ceph/rgw/files/users.j2
```diff
@@ -0,0 +1,12 @@
...
```
swiftgist: Created a macro to do the indenting and updated the README

smithfarm:
@swiftgist I ran the basic suite on this PR, and there were two failures - both in the nfs-ganesha test cases that use the RGW FSAL (which mean they also use DS to create rgw users). Note that the regular rgw test passed - it does not create any rgw users, though.

http://149.202.181.82:8081/ubuntu-2017-09-11_11:01:42-deepsea:basic-ses5---basic-openstack/

rjfd:
@swiftgist taken from the minion log:

```
realm:
  default:


      - { uid: "admin", name: "Admin", email: "admin@admin.nil", system: True }




            - { uid: "demo", name: "Demo", email: "demo@demo.nil" }
            - { uid: "demo1", name: "Demo1", email: "demo1@demo.nil" }
```

Apparently the filter your using `indent(6, true)` is adding 12 spaces instead of 6...

swiftgist:
The extra spaces were in the qa/common/rgw.sh.

smithfarm:
New test run here: http://149.202.181.82:8081/ubuntu-2017-09-11_15:03:32-deepsea:basic-ses5---basic-openstack/

If it passes, consider the PR approved by me.

smithfarm:
The nfs-ganesha RGW FSAL tests fail because DS is not populating `/etc/ganesha/ganesha.conf` correctly. It should have EXPORT sections for both demo users as well as an RGW section, but these are missing. (The file can be found in the logs of the failed tests by grepping for `/etc/ganesha/ganesha.conf`.)

srv/salt/_modules/rgw.py
rjfd: I just noticed that iteration over a python dict return the key values, which means that this part of code should work as expected.

smithfarm:
nfs-ganesha **running** http://149.202.181.82:8081/ubuntu-2017-09-13_02:28:43-deepsea:basic:nfs-ganesha-ses5---basic-openstack/

If it passes, I'll run the whole suite again just to be sure.

PR #610, opened 2017-09-07
deepsea-cli: A CLI tool for visualizing and running DeepSea stages

rjfd:
This PR adds the implementation of the DeepSea CLI tool that allows a user to visualize the progress of DeepSea stages' execution.

The tool provides two modes:
* **the monitoring mode**: the user runs the monitor and executes salt-run commands in another terminal session
* **the running mode**: the user runs and visualizes the stage using the CLI tool

Example of some commands:

```
$ deepsea --help
$ deepsea monitor
$ deepsea stage show ceph.stage.0
$ deepsea stage run ceph.stage.0
```

```
```


cli/deepsea.py
```diff
@@ -0,0 +1,256 @@
...
+from __future__ import print_function
+
+import logging.config
+import logging
+import os
+import pkg_resources
+import signal
+import sys
+import time
+import click
```
smithfarm: This looks like a new dependency; can you add a `Requires: python-click` to the spec file? (We'll also need to make sure we ship this dependency...)

rjfd: It's already there: https://github.com/rjfd/DeepSea/blob/wip-monitor-rebase/deepsea.spec#L57

deepsea.spec
```diff
@@ -45,16 +45,34 @@ BuildArch:      noarch
...
+#################################################################################
+# subpackages
+#################################################################################
+%package cli
+Summary:        DeepSea CLI
+Group:          Systems/Filesystems
+BuildRequires:  python-setuptools
+Requires:       deepsea = %{version}-%{release}
+Requires:       python-setuptools
+Requires:       python-click
```
smithfarm: ah, ok, now I see this, but I don't see the `%package cli` and `%description cli` sections?

smithfarm: argh, nevermind. I'm turning blind in my old age.

cli/monitor.py
```diff
@@ -0,0 +1,700 @@
...
+            if event.jid in self._dynamic_steps:
+                step = self._dynamic_steps[event.jid]
+                step.finish(event)
+                return step
+
+        return None
+
+    def state_result_step(self, event):
+        """
+        Consumes teh current step
```
tserong: Typo in comment (s/teh/the/)

cli/monitors/terminal_outputter.py
```diff
@@ -0,0 +1,770 @@
...
+            if isinstance(arg, dict):
+                for key, val in arg.items():
+                    if first:
+                        args += "{}={}".format(key, val)
+                        first = False
+                    else:
+                        args += ", {}={}".format(key, val)
+            else:
+                if first:
+                    first = True
```
tserong: Isn't that redundant?

cli/monitors/terminal_outputter.py
```diff
@@ -0,0 +1,770 @@
...
```
tserong: As above -- `if first: first = True` seems to be setting first to what it's already set to.

cli/monitors/terminal_outputter.py
rjfd: hehe, good catch, it should be `first = False`

cli/monitor.py
```diff
@@ -0,0 +1,700 @@
...
+        for dep in curr_step.step.on_fail_deps:
+            for i in range(1, self.current_step+1):
+                dep_step = self._steps[self.current_step-i]
+                if dep_step.step.desc == dep.desc:
+                    if dep_step.success:
+                        curr_step.skipped = True
+                        self.current_step += 1
+                        return curr_step
+
+        return None
```
theanalyst: Why are we returning None, can't we return True/False instead?

deepsea.spec
```diff
@@ -32,35 +32,22 @@ Url:            https://github.com/suse/deepsea
...
 
 BuildRequires:  salt-master
+BuildRequires:  python-setuptools
 Requires:       salt-master
 Requires:       salt-minion
 Requires:       salt-api
 Requires:       python-ipaddress
 Requires:       python-netaddr
 Requires:       python-rados
+Requires:       python-setuptools
```
smithfarm: why is `python-setuptools` both a build dependency and a runtime dependency?

rjfd:
@tserong many thanks for your review. 

Yes, I would also like to have a manpage, let's see if I can prepare it on time.

deepsea.spec
rjfd: It's necessary for the "console script" (deepsea command)  created by the setuptools installer to work.

cli/monitor.py
rjfd: Because we are returning an object in the above returns, and I think we should not mix different types for a return function.

ricardoasmarques:
If we run "deepsea" without root permissions this tool will output a python stacktrace.

It would be great if you can display a pretty message like:

**"Root privileges are required to run this tool."**

rjfd:
@ricardoasmarques done :-)

rjfd:
@smithfarm @jan--f @swiftgist I found a way to change the version number on the `deepsea.spec` that resides inside the tarball during the rpmbuild process. See https://github.com/rjfd/DeepSea/blob/wip-monitor-rebase/deepsea.spec#L57

The only problem now is that the version number generated by the OBS service is using the "+" character to concatenate the version number with the GIT hash, which is not a valid character for defining version number of python packages. See the output of `python setup.py install` when using the "+" character

```
dist.py:294: UserWarning: The version specified ('0.7.26.cli+git.0.42374c3') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details
```

So, do you think it's possible to change the version number to something that does not use a "+", for instance: `0.7.26.cli_git.0.42374c3` or `0.7.26.cli.git.0.42374c3` ?

smithfarm:
The integration test currently fails with "Package 'python-click' not found." Working on it.

smithfarm:
> So, do you think it's possible to change the version number to something that does not use a "+", for instance: 0.7.26.cli_git.0.42374c3 or 0.7.26.cli.git.0.42374c3 ?

@rjfd According to this [1] "The version number is normally taken verbatim from the package's version. The only restriction placed on the version is that it cannot contain a dash "-"."

So my guess is that either way (`.` or `_`) will be fine. Personally I'd go with the `.`.

[1] http://ftp.rpm.org/max-rpm/ch-rpm-file-format.html

smithfarm:
After generating new OVH images with python-click preinstalled, I got the new health-ok-cli.sh test to run, and it got all the way to stage 3 before blowing up: http://149.202.181.82:8081/ubuntu-2017-09-08_20:24:35-deepsea:basic:health-ok-cli-ses5---basic-openstack/

rjfd:
@smithfarm can we get access to `/var/log/deepsea.log' of that run?

smithfarm:
@rjfd Unfortunately, no - at least, not from that particular run. Here is the code I'm currently using to harvest the salt and ganesha logs: https://github.com/SUSE/ceph/blob/ses5/qa/tasks/deepsea.py#L194-L215

As you can see, it works well for gathering all logs that appear in a directory, i.e. `/var/log/$FOO/*`

Either I can implement a `gather_individual_log_file()` function in the deepsea task, or you could modify the code in this PR to log to something like `/var/log/deepsea/monitor.log` instead? I'm OK either way, but it seems like other parts of deepsea might want to log to their own files at some point, so having a directory might make sense. What do you think?

smithfarm:
Another option would be to have the deepsea task dump `/var/log/deepsea.log` (if it exists) to stdout in the teardown phase, regardless of success or failure. That will make it easier to analyze this particular log, since it will be included in teuthology.log.

rjfd:
@smithfarm why don't you run deepsea-cli with the `--log-file` option like:

```
$ deepsea --log-file=/var/log/salt/deepsea.log --log-level=debug stage run ceph.stage.0
```

That way the log will be created inside the salt directory, that should work, right?

smithfarm:
@rjfd Here's the log: http://149.202.181.82/ubuntu-2017-09-09_08:19:38-deepsea:basic:health-ok-cli-ses5---basic-openstack/38/remote/target137074026204/

smithfarm:
> that should work, right?

@rjfd Sure, something like this:

```
diff --git a/qa/common/helper.sh b/qa/common/helper.sh
index 87c46a4..4f387f3 100644
--- a/qa/common/helper.sh
+++ b/qa/common/helper.sh
@@ -31,7 +31,13 @@ function _run_stage {
       echo "************ using DeepSea CLI **************"
       echo "*********************************************"
       echo ""
-      deepsea stage run ceph.stage.${stage_num} --simple-output
+      deepsea \
+          --log-file=/var/log/salt/deepsea.log \
+          --log-level=debug \
+          stage \
+          run \
+          ceph.stage.${stage_num} \
+          --simple-output
       return
   fi
```

rjfd:
@smithfarm I think I already know what the problem is. In this suite, is there any minion configured with the RGW role?

smithfarm:
Nope.

```
2017-09-09T08:25:50.552 INFO:teuthology.orchestra.run.target137074026204.stderr:+ cat_policy_cfg
2017-09-09T08:25:50.553 INFO:teuthology.orchestra.run.target137074026204.stderr:+ cat /srv/pillar/ceph/proposals/policy.cfg
2017-09-09T08:25:50.553 INFO:teuthology.orchestra.run.target137074026204.stdout:# Cluster assignment
2017-09-09T08:25:50.554 INFO:teuthology.orchestra.run.target137074026204.stdout:cluster-ceph/cluster/*.sls
2017-09-09T08:25:50.554 INFO:teuthology.orchestra.run.target137074026204.stdout:# Common configuration
2017-09-09T08:25:50.554 INFO:teuthology.orchestra.run.target137074026204.stdout:config/stack/default/global.yml
2017-09-09T08:25:50.554 INFO:teuthology.orchestra.run.target137074026204.stdout:config/stack/default/ceph/cluster.yml
2017-09-09T08:25:50.554 INFO:teuthology.orchestra.run.target137074026204.stdout:# Role assignment - master
2017-09-09T08:25:50.555 INFO:teuthology.orchestra.run.target137074026204.stdout:role-master/cluster/target137074026204.teuthology*.sls
2017-09-09T08:25:50.555 INFO:teuthology.orchestra.run.target137074026204.stdout:# Role assignment - admin
2017-09-09T08:25:50.555 INFO:teuthology.orchestra.run.target137074026204.stdout:role-admin/cluster/*.sls
2017-09-09T08:25:50.555 INFO:teuthology.orchestra.run.target137074026204.stdout:# Role assignment - mon
2017-09-09T08:25:50.555 INFO:teuthology.orchestra.run.target137074026204.stdout:role-mon/cluster/*.sls slice=[:1]
2017-09-09T08:25:50.555 INFO:teuthology.orchestra.run.target137074026204.stdout:role-mgr/cluster/*.sls slice=[:1]
2017-09-09T08:25:50.556 INFO:teuthology.orchestra.run.target137074026204.stdout:role-mon/stack/default/ceph/minions/*.yml slice=[:1]
2017-09-09T08:25:50.556 INFO:teuthology.orchestra.run.target137074026204.stdout:# Hardware Profile
2017-09-09T08:25:50.556 INFO:teuthology.orchestra.run.target137074026204.stdout:profile-default/cluster/*.sls
2017-09-09T08:25:50.556 INFO:teuthology.orchestra.run.target137074026204.stdout:profile-default/stack/default/ceph/minions/*yml
2017-09-09T08:25:50.557 INFO:teuthology.orchestra.run.target137074026204.stdout:
```

rjfd:
Ok, I already know what to do. I'll push a fix later in the day.

rjfd:
@smithfarm pushed the fix for the teuthology errors, and also applied the patch to `qa/common/helper.sh`

rjfd:
Pushed change that fixes the color scheme for lighter terminal backgrounds

rjfd:
Added man pages (still incomplete)

man/deepsea-monitor.1
```diff
@@ -0,0 +1,66 @@
+.TH deepsea-monitor 1 "DeepSea CLI man page"
+.SH NAME
+deepsea-monitor - Starts the stage execution monitor.
```
smithfarm: "Starts the DeepSea stage execution progress monitor" ? (also see next comment, below)

man/deepsea-monitor.1
```diff
@@ -0,0 +1,66 @@
...
+
+.SH SYNOPSIS
+deepsea monitor [--help] [--show-dynamic-steps] [--show-state-steps]
+                [--simple-output]
+
+.SH DESCRIPTION
+Starts DeepSea progress monitor.
```
smithfarm: "Starts the DeepSea stage execution progress monitor." ? (to unify the description throughout the manpage)

man/deepsea-monitor.1
```diff
@@ -0,0 +1,66 @@
...
+
+This allows to visualize DeepSea execution progress when running stages
+using salt-run commands in other terminal sessions.
```
smithfarm: "using salt-run state.orch commands" ? (see next paragraph)

man/deepsea-monitor.1
```diff
@@ -0,0 +1,66 @@
...
+deepsea-monitor - Starts the stage execution monitor.
+
+.SH SYNOPSIS
+deepsea monitor [--help] [--show-dynamic-steps] [--show-state-steps]
+                [--simple-output]
+
+.SH DESCRIPTION
+Starts DeepSea progress monitor.
+
+This allows to visualize DeepSea execution progress when running stages
```
smithfarm: "The progress monitor provides a detailed, real-time visualization of what is happening during execution of stages" ?

man/deepsea-monitor.1
```diff
@@ -0,0 +1,66 @@
...
+Enables the visualization of Salt state intermediate steps. Only
+state steps that have the option
+.B fire_event=True
+are shown
+.RE
+.RS
+.B Caution:
+If you enable this option, the stage parsing stage that occurs when the stage
+starts its execution might take much longer, which will defer the execution
+progress visualization for an undertermined period of time.
```
smithfarm: typo "undetermined"

man/deepsea-monitor.1
```diff
@@ -0,0 +1,66 @@
...
+
+.RE
+.B --simple-output
+.RS
+Enables a minimalistic visualization layout without colors.
+Very useful for testing purposes or for output redirection.
```
smithfarm: Useful when redirecting the monitor output to a text file.

man/deepsea-stage-run.1
```diff
@@ -0,0 +1,75 @@
+.TH deepsea-stage-run 1 "DeepSea CLI man page"
+.SH NAME
+deepsea-stage-run - Runs DeepSea stages.
```
smithfarm: "Run DeepSea stages with real-time, user-friendly output." ?

man/deepsea-stage-run.1
```diff
@@ -0,0 +1,75 @@
...
+
+.SH SYNOPSIS
+deepsea stage run [--help] [--hide-dynamic-steps] [--hide-state-steps]
+                  [--simple-output]
+                  <stage_name>
+
+.SH DESCRIPTION
+Runs a DeepSea stage and shows its execution progress.
```
smithfarm: Runs a DeepSea stage, showing its execution progress in real time.

man/deepsea-monitor.1
```diff
@@ -0,0 +1,66 @@
...
+
+.SH DESCRIPTION
+Starts DeepSea progress monitor.
+
+This allows to visualize DeepSea execution progress when running stages
+using salt-run commands in other terminal sessions.
+
+The user must start the monitor before running any
+.I salt-run state.orch
+so that the monitor can detect the start of stages' execution.
```
smithfarm: the start of the stage's execution.

man/deepsea-stage-run.1
```diff
@@ -0,0 +1,75 @@
...
+
+This commands is an alternative to the Salt based commands to run DeepSea
```
smithfarm: the Salt-based commands for running

man/deepsea-stage-run.1
```diff
@@ -0,0 +1,75 @@
...
+stages. For instance:
```
smithfarm: for running DeepSea stages (or any DeepSea orchestration state file).

man/deepsea-stage-run.1
```diff
@@ -0,0 +1,75 @@
...
+.RS
+.B deepsea stage run ceph.stage.0
+.RE
+is equivalent to run:
```
smithfarm: is equivalent to:

man/deepsea-stage-run.1
```diff
@@ -0,0 +1,75 @@
...
+.B --hide-state-steps
+.RS
+Disables the visualization of Salt state intermediate steps. This
+may speed up the initialization time for running a stage.
+
+.RE
+.B --simple-output
+.RS
+Enables a minimalistic visualization layout without colors.
+Very useful for testing purposes or for output redirection.
```
smithfarm: Useful when redirecting the monitor output to a text file.

man/deepsea-stage-show.1
```diff
@@ -0,0 +1,69 @@
...
+.SH NAME
+deepsea-stage-show - Shows DeepSea stage execution steps
+
+.SH SYNOPSIS
+deepsea stage show [--help] [--hide-state-steps] [--only-visible-steps]
+                   [--clear-cache] [--no-cache]
+                   <stage_name>
+
+.SH DESCRIPTION
+Shows a DeepSea stage steps ordered by execution order.
```
smithfarm: Parses a DeepSea stage and generates a detailed report showing all steps in order of execution.

man/deepsea-stage-show.1
```diff
@@ -0,0 +1,69 @@
...
+
+This command can be used to understand what are the steps involved in each
+stage. For instance, it shows the total number of steps; the type of each step,
```
smithfarm: This command can be used to understand what a stage will do, without actually running the stage.

man/deepsea-stage-show.1
```diff
@@ -0,0 +1,69 @@
...
+
+.RE
+.B --hide-state-steps
+.RS
+Disables the parsing of Salt state intermediate steps.
+
+.RE
+.B --only-visible-steps
+.RS
+Only show state operations that declared
```
smithfarm: is there a difference between "operation" and "step" of a stage? If not, let's use "step" to keep the terminology consistent.

"Only show steps that have `fire_event=True` in their Salt configuration." ?

man/deepsea-stage-show.1
```diff
@@ -0,0 +1,69 @@
...
+.B fire_event=True
+
+By default, all states defined in each state file are shown.
```
smithfarm: "all steps defined in each state file" ?

man/deepsea.1
```diff
@@ -0,0 +1,78 @@
...
+DeepSea CLI
+
+.SH SYNOPSIS
+deepsea [--version] [--help] [-l <log_level> | --log-level=<log_level>]
+        [--log-file=<path>]
+        <command> [<args>]
+
+.SH DESCRIPTION
+DeepSea CLI is a command-line-interface tool to run and visualize DeepSea's
+stages execution progress.
```
smithfarm: is a command-line tool for running DeepSea stages and visualizing their execution progress.

man/deepsea.1
```diff
@@ -0,0 +1,78 @@
...
+
+Two modes are supported for visualizing a stage's execution progress.
+.RS
+.RE
+.B 1) monitoring mode:
+which allows to visualize the stage execution progress triggered by a salt-run
```
smithfarm: s/which allows to visualize/for visualizing/

man/deepsea.1
```diff
@@ -0,0 +1,78 @@
...
+command issued in another terminal session; and
```
smithfarm: s/which allows to visualize/visualizes/

man/deepsea.1
```diff
@@ -0,0 +1,78 @@
...
+.RS
+.RE
+.B 2) stand-alone mode:
+which allows to run and visualze the stage execution progress from DeepSea CLI
+only.
```
smithfarm: runs a DeepSea stage while providing real-time visualization of its component steps as they are executed.

man/deepsea.1
smithfarm: visualizes the execution progress of a DeepSea stage triggered by a salt-run . . .

man/deepsea.1
```diff
@@ -0,0 +1,78 @@
...
+.B error
+.B debug
+.B silent
+Default:
+.B info
+
+.RE
+.B --log-file=<path>
+.RS
+Sets the log file location.
```
smithfarm: Defaults to `/var/log/deepsea.log`

rjfd:
@smithfarm thank you very much for your thorough review of the man pages. I addressed all the comments and pushed the changes

smithfarm:
@rjfd An idea that just came to me - the `deepsea stage show` subcommand is essentially a "dry run". Would it make sense to change the syntax to `deepsea stage dry-run` ?

rjfd:
@smithfarm nice suggestion. I renamed `stage show` to `stage dry-run`.

rjfd:
Pushed the new `salt-run state.orch` alias, and updated man pages accordingly.

@smithfarm could you run deepsea teuthology suites again on this PR?

smithfarm:
full run **pass** http://149.202.181.82:8081/ubuntu-2017-09-13_06:55:51-deepsea:basic-ses5---basic-openstack/

PR #614, opened 2017-09-08
Don't create/append partitions when deploying encrypted disks

jschmid1:
ceph-disk expects the raw device and handles the creation of the next higest partition in addition to encrypting the wal/db/device.


jschmid1:
closing in favor of #620

PR #631, opened 2017-09-14
Display more statistics on the RGW Grafana dashboard

votdev:
Display more user usage information in the Grafana dashboard (OP-2669)
See https://tracker.openattic.org/browse/OP-2669 for more information.

### DIFF
--- a/srv/salt/ceph/monitoring/grafana/files/ceph-rgw-users.json
+++ b/srv/salt/ceph/monitoring/grafana/files/ceph-rgw-users.json
@@ -80,7 +80,7 @@
                 "legendFormat": "Total",
                 "metric": "",
                 "refId": "A",
-                "step": 3600
+                "step": 60
               },
               {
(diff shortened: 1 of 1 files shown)


jan--f:
I'm a little worried about the exporter. Have you done any experiments when there is a lot of metrics to export?
I'm wondering if a similar setup as for rbd wouldn't be more appropriate. I.e. export metrics to a file and let the node_exporter handle the http end of things. What do you think @votdev ?

votdev:
@jan--f 
I was not able to test this with many metrics. What you are worried about exactly? Do you think the exporter is not able to process and deliver the metrics within the defined scrape time (1min)? Should we do that every 5 minutes or so?
Using the node_exporter should be fine, too. It will simplify the setup and does not need another systemd unit and a running HTTP server process.
I will create a new pull request for that because this PR is not related to that issue.

PR #634, opened 2017-09-14
Add sysctl, set aio-max-nr to 1M - 16x default

### LINKED ISSUE #475: adjust fs.aio-max-nr in larger deployments (or just always?)
ceph-disk fails silently when you hit the threshold defined in `/proc/fs/aio-max-nr` 

https://www.ibm.com/developerworks/linux/library/l-async/
https://www.novell.com/support/kb/doc.php?id=3445451

there should be no drawback in increasing the default max-nr as it doesn't pre-allocate.

swiftgist:
Fixes #475 


srv/salt/ceph/sysctl/default.sls
```diff
@@ -0,0 +1,14 @@
+
+/etc/sysctl.d/deepsea-aio-max-nr.conf:
```
l-mb: This is not a good filename, since without numbers, it's not clear in which order the - possibly conflicting - sysctls will be applied.

swiftgist: Created #681

ddiss:
Upstream Ceph will soon have matching functionality via https://github.com/ceph/ceph/pull/17894 . We should probably consider reverting this once the Ceph change is merged.

Issue #636, opened 2017-09-15
Filesystem rights for the RGW SSL certificate /etc/ceph/rgw.pem need to be secure


Martin-Weiss:
Could we adjust the filesystem rights for /etc/ceph/rgw.pem during the deployment so that only the user (ceph) that is used to run the radosgw can read the private key?

PR #637, opened 2017-09-15
rgw: ssl: make pem files readable only to ceph user & group

theanalyst:
Fixes: https://github.com/SUSE/DeepSea/issues/636

Reported-by: Martin Weiss <mweiss@suse.com>

### DIFF
--- a/srv/salt/ceph/rgw/cert/default.sls
+++ b/srv/salt/ceph/rgw/cert/default.sls
@@ -3,3 +3,6 @@ deploy the rgw.pem file:
   file.managed:
     - name: /etc/ceph/rgw.pem
     - source: salt://ceph/rgw/cert/rgw.pem
+    - user: ceph
+    - group: ceph
+    - mode: 600
--- a/qa/common/common.sh
+++ b/qa/common/common.sh
@@ -440,3 +440,18 @@ EOF
   # FIXME: assert script not running on the iSCSI gateway node
   _run_test_script_on_node $TESTSCRIPT $CLIENTNODE
 }
+
+function validate_rgw_cert_perm {
+    local TESTSCRIPT=/tmp/test_rados_put.sh
+    local RGWNODE=$(_first_x_node rgw-ssl)
+    cat << 'EOF' > $TESTSCRIPT
+set -ex
+trap 'echo "Result: NOT_OK"' ERR
+RGW_PEM=/etc/ceph/rgw.pem
+test -f "$RGW_PEM"
+test "$(stat -c'%U' $RGW_PEM)" == "ceph"
+test "$(stat -c'%G' $RGW_PEM)" == "ceph"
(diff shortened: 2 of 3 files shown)


Martin-Weiss:
Does this also set 600 ?

srv/salt/ceph/rgw/cert/default.sls
```diff
@@ -3,3 +3,5 @@ deploy the rgw.pem file:
   file.managed:
     - name: /etc/ceph/rgw.pem
     - source: salt://ceph/rgw/cert/rgw.pem
+    - user: ceph
+    - group: ceph
```
smithfarm: as @Martin-Weiss mentioned, it needs `mode: 600`

theanalyst:
@smithfarm @Martin-Weiss thanks for the input, updated and added a test case, let me run this locally to see if everything works fine.

qa/common/rgw.sh
```diff
@@ -123,3 +123,17 @@ function rgw_ssl_init {
...
 }
+
+
+function validate_rgw_cert_perm {
+    local TESTSCRIPT=/tmp/test_rados_put.sh
+    local RGWNODE=$(_first_x_node rgw-ssl)
+    cat << 'EOF' > $TESTSCRIPT
+set -ex
+trap 'echo "Result: NOT_OK"' ERR
+test $(stat -c'%U' /etc/ceph/rgw.pem) == "ceph"
```
smithfarm: If the file does not exist, we get an unexpected error:

```
# test $(stat -c'%U' /non/existent/rgw.pem) == "root"
stat: cannot stat '/non/existent/rgw.pem': No such file or directory
bash: test: ==: unary operator expected
```

This can be avoided by putting the lval in double-quotes like so:

```
test "$(stat -c'%U' /non/existent/rgw.pem)" == "ceph"
stat: cannot stat '/non/existent/rgw.pem': No such file or directory
```

Also, could we put the `/etc/ceph/rgw.pem` in a variable and test that the file exists, something like:

```
RGW_PEM=/etc/ceph/rgw.pem
test -f "$RGW_PEM"
test "$(stat -c'%U' $RGW_PEM)" == "ceph"
test "$(stat -c'%G' $RGW_PEM)" == "ceph"
test "$(stat -c'%a' $RGW_PEM)" -eq 600
```

theanalyst: thanks, will add this

theanalyst:
changelog: move the test snippet from rgw.sh -> common.sh

theanalyst:
fixed a typo, we were calling a non existant function earlier, +validate_rgw_cert_perm -validate_rgw_cert_pem

theanalyst:
@smithfarm passes the run in my vagrant env.

smithfarm:
Oops, this was missing an `echo "Result: OK"` line at the end of `validate_rgw_cert_perm`

Fixing in #643

PR #632, opened 2017-09-14
Prevents disruptive service restarts

jschmid1:
Prevents disruptive service restarts change in one of the corresponding config files and allow a restart if conditions are met.

Followup discussion:
What config files changes affect which service?

I.e. if we change the mon configs do we also want to restart the osds/rgw etc.
same questions apply for nfs-ganesha, mds, mgr etc.  

tackles #584 

Signed-off-by: Joshua Schmid jschmid@suse.com

PR #620, opened 2017-09-12
Properly handle bluestore args

jschmid1:
Issues:

* Encryption:
ceph-disk expects the raw device and handles the creation
of the next higest partition in addition to encrypting
the wal/db/device.

* Bluestore deployment:
refer to #615 

Also refactoring the decision paths in `_bluestore_args()`.

resolves #615 

Tested:

* wal/db defined / no size specified / unencrypted
* wal/db defined  no size specified / encrypted
* wal/db defined / size specified / unencrypted
* wal/db defined / size specified / encrypted
* no wal/db defined / no size specified / unencrypted
* no wal/db defined  no size specified / encrypted
* no wal/db defined / size specified / unencrypted
* no wal/db defined / size specified / encrypted

Signed-off-by: Joshua Schmid <jschmid@suse.de>

Tackles: https://github.com/SUSE/DeepSea/issues/755
EDIT: tests can be found in #623

PR #642, opened 2017-09-15
cli: gracefully handle SLS rendering errors

rjfd:
The deepsea-cli was not handling the case when some SLS file contained syntax errors, and an ugly python stack trace was being thrown.

This PR implements the handling of SLS syntax errors.

Video with demo: https://asciinema.org/a/ap5G06hRIxR2JO2x6YOp7hAY6

### DIFF
--- a/cli/stage_executor.py
+++ b/cli/stage_executor.py
@@ -19,9 +19,9 @@ import threading
 import time
 import sys
 
-from .common import PrettyPrinter as PP
 from .monitor import Monitor
 from .monitors.terminal_outputter import SimplePrinter, StepListPrinter
+from .stage_parser import RenderingException
 
 
 # pylint: disable=C0103
@@ -74,9 +74,12 @@ def run_stage(stage_name, hide_state_steps, hide_dynamic_steps, simple_output):
     mon = Monitor(not hide_state_steps, not hide_dynamic_steps)
     printer = SimplePrinter() if simple_output else StepListPrinter(False)
     mon.add_listener(printer)
-    mon.parse_stage(stage_name)
-    mon.start()
+    try:
+        mon.parse_stage(stage_name)
+    except RenderingException:
+        return 2
 
+    mon.start()
     executor = StageExecutor(stage_name)
 
     # pylint: disable=W0613
--- a/cli/stage_parser.py
+++ b/cli/stage_parser.py
@@ -14,6 +14,7 @@ import StringIO
 from collections import OrderedDict, defaultdict
 
 import salt.client
+import salt.exceptions
 
 from .common import redirect_stdout, redirect_stderr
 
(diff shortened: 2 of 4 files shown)


smithfarm:
@rjfd This is a nit and it doesn't relate directly to this PR, but . . . I noticed that `check_root_privileges` exits immediately, and it's called on every invocation of deepsea, even ones that include the `--help` option. So, for example:

```
cephadm@ip-10-0-4-10:~> deepsea --help
Root privileges are required to run this tool
cephadm@ip-10-0-4-10:~> deepsea stage --help
Root privileges are required to run this tool
cephadm@ip-10-0-4-10:~> deepsea foobarbaz
Root privileges are required to run this tool
```

smithfarm:
**Test runs of interest**

These two runs include #620, #632, and #643 in addition to this PR:

* http://149.202.189.253:8081/ubuntu-2017-09-16_09:19:24-deepsea:basic-ses5---basic-openstack/
* http://149.202.189.253:8081/ubuntu-2017-09-16_10:15:31-deepsea:basic:rgw-ses5---basic-openstack/

Just this PR and #643:

* http://149.202.189.253:8081/ubuntu-2017-09-16_10:57:16-deepsea:basic:rgw-ses5---basic-openstack/

cli/stage_parser.py
```diff
@@ -49,9 +94,18 @@ def render(file_name):
...
         err = StringIO.StringIO()
         out = StringIO.StringIO()
+        exception = None
         with redirect_stderr(err):
             with redirect_stdout(out):
-                result = __caller__.cmd('slsutil.renderer', file_name)
+                try:
+                    result = __caller__.cmd('slsutil.renderer', file_name)
+                except salt.exceptions.SaltException as ex:
+                    exception = StageRenderingException(file_name, ex.strerror)
```
tserong: Why not raise the exception here? Or does that cause trouble becuase stderr and stdout are redirected?

cli/stage_parser.py
```diff
@@ -35,6 +36,50 @@ class OrchestrationNotFound(Exception):
...
+        super(RenderingException, self).__init__()
+        self.error_desc = error_desc
+
+    def pretty_error_desc_str(self):
+        """
+        Returns a more user-friendly message of the exception
+        """
+        output = ""
+        if self.error_desc.find("SaltRenderError") == -1:
+            found_ex_line = True
```
tserong: I don't understand this piece. That reads to me as if it would set found_ex_line = True if "SaltRenderError" was *not* found? Then later, it checks if each line starts with "SaltRenderError" anyway?

cli/stage_executor.py
```diff
@@ -71,12 +73,30 @@ def run_stage(stage_name, hide_state_steps, hide_dynamic_steps, simple_output):
...
+            PP.println(PP.bold("An error occurred while rendering one of the following states:"))
+            for state in ex.states:
+                PP.print(PP.cyan("    - {}".format(state)))
+                PP.println(" ({})".format("/srv/salt/{}".format(state.replace(".", "/"))))
+        else:
+            PP.println(PP.bold("An error occurred while rendering the stage file:"))
+            PP.println(PP.cyan("    {}".format(ex.stage_file)))
+        PP.println()
+        PP.println(PP.bold("Error description:"))
+        PP.println(PP.red(ex.pretty_error_desc_str()))
```
tserong: The above block looks like it's the same as the block in start_stage() in monitor.py. Can these two be sanely put into a separate helper function?

cli/stage_parser.py
rjfd: Right, if I raise the exception here, all the output made by the code that catches the exception is redirected.

cli/stage_executor.py
rjfd: Yes, you're right. I'll refactor this to have a single location with the prints for this kind of exception.

cli/stage_parser.py
rjfd: The idea here is to remove the stack trace part of a `SaltRenderError` exception and keep the rest of the exception output, but keep the stack trace part of any other exception that might be caught.

I agree that the way this is written is confusing, I'll change this into a more readable piece of code.

rjfd:
@smithfarm 

>  noticed that check_root_privileges exits immediately, and it's called on every invocation of deepsea, even ones that include the --help option.

Yes, need to fix that. Currently every code path checks if the user is "root", even when just using the --help option.
I also need to make the CLI to run the "stage parsing" phase as the "salt" user, otherwise the when runners are executed due to the rendering of sls files, some runner may create files, which only root can read or write.

rjfd:
@tserong addressed your comments, and pushed to the PR.
I moved the output code to the place where it should belong, which is in the `terminal_outputter.py` module.

rjfd:
@smithfarm care to run the teuthology suite to see if the latest changes didn't break anything?

smithfarm:
**running** http://149.202.189.253:8081/ubuntu-2017-09-21_12:04:29-deepsea:basic-ses5---basic-openstack/

PR #644, opened 2017-09-16
Wip conditional restarts

swiftgist:
This is a continuation of #632 to support rgw configurations and to rename the rgw configuration files.  

I did disable bucket creation because endpoints is failing with custom rgw configurations.  Restarting the services correctly is more involved than one would think and these commits are touching a few files. 
 Since #470 migrates the functionality to a module, we can enable bucket creation there.


theanalyst:
https://jonathanchang.org/blog/pushing-to-a-pull-request-on-github/ explains how to edit someone's pr, as far as modify pull request from collaborators is ticked this applies and is an easier workflow in this case

srv/modules/runners/changed.py
```diff
@@ -0,0 +1,153 @@
...
+        log.debug("dependencies: {}".format(self.dependencies))
+
+    def dependencies(self):
+        """
+        Services might depend on each other and have to trigger a restart.
+        I.e. Changes in the MDS section might affect the NFS-Ganesha service
+        hence the service should also be restarted.
+        # TODO for other services and complete tree
+        """
+        return { 'mds': ['mds'],
```
jschmid1: 'rgw': ['rgw'] is missing here

swiftgist: It's not needed.... it happens in rgw_configurations.

jschmid1: didn't see the changed.config(service)..

swiftgist:
Merged into #632

PR #649, opened 2017-09-18
Refactor RGW Prometheus exporter to use node_exporter (OP-2678)

votdev:
https://tracker.openattic.org/browse/OP-2678

- Do not run the Prometheus RGW exporter as a webserver
- Remove systemd unit to handle the RGW exporter
- Introduce cron job to execute the RGW exporter every 5 minutes and write the content to a *.prom file
- Use node_exporter to process the metrics
- RGW exporter: Create a new registry, otherwise unwanted default collectors are added automatically
- Relocate installation of Prometheus RGW exporter from ceph.stage.deploy to ceph.stage.radosgw
- The Prometheus RGW exporter will be installed on the 'master' node if there is at least one 'rgw' node. The role 'master' takes care that the required admin keyring is installed which is required by the radosgw-admin CLI tool that is used by the exporter script. See https://bugzilla.suse.com/show_bug.cgi?id=1059621.


jan--f:
Oh and I'd prefer the change to the select runner to be in a separate PR.

PR #660, opened 2017-09-20
Realize DEV_ENV=true also in case there is exactly one minion

Martin-Weiss:
In case there is only one single minion (test-environment with everything on a single server) - the result of en(self.data.keys()) is 1 and with that this is not greater than 1 and DEV_ENV=true is not detected. Changing to ">=" to allow also 1.

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -170,7 +170,7 @@ class Validate(object):
     def __dev_env(self):
         if 'DEV_ENV' in os.environ:
             return os.environ['DEV_ENV'].lower() != 'false'
-        elif len(self.data.keys()) > 1:
+        elif len(self.data.keys()) >= 1:
             any_minion = self.data.keys()[0]
             if 'DEV_ENV' in self.data[any_minion]:
                 return self.data[any_minion]['DEV_ENV']


smithfarm:
@Martin-Weiss I'm curious how the "DEV_ENV=true is not detected" is manifesting on your single-node test cluster?

I ask because we regularly run a full deployment test on a single-node cluster, relying on DEV_ENV="true" to achieve a successful deployment.

smithfarm:
During Stage 3 initialization I am seeing "DEV_ENV:       true" and I double-checked that there is only one node in the cluster (this single node is doing everything, including DeepSea, Salt Master, Salt Minion, etc.)

Martin-Weiss:
How do you configure DEV_ENV=true in your environment? 

I am setting DEV_ENV: true in /srv/pillar/ceph/stack/global.yml as a pillar and with this I see the issue.
In case I use "export DEV_ENV=true" I do NOT see this issue.

srv/modules/runners/validate.py
```diff
@@ -170,7 +170,7 @@ def __init__(self, name, data=None, grains=None, printer=None):
     def __dev_env(self):
         if 'DEV_ENV' in os.environ:
             return os.environ['DEV_ENV'].lower() != 'false'
-        elif len(self.data.keys()) > 1:
+        elif len(self.data.keys()) >= 1:
```
rjfd: change this line to: `elif self.data:`

This will test if the `self.data` is not empty, which is a more correct (pythonic) way than checking if the number of keys is greater than 0.

rjfd:
The current code would only use the `DEV_ENV` option from the pillar if more than one `key` existed in the pillar data variable `self.data`, which is wrong. Nice catch @Martin-Weiss.

No such problem exists when suing DEV_ENV as an environment variable.

srv/modules/runners/validate.py
Martin-Weiss: Thanks for the feeback! As I am not a real developer - I do not care how this is fixed ;-)
>= works for be - but if someone can rest it I would also be fine with elif self.data: .

The major question I have is more about "why do we need to test for existing minions here at all?"

rjfd: It's just because the way the pillar data is being assigned to that `self.data` variable. the `self.data` variable is assigned with the output of the `salt '*' pillar.items` salt command, which corresponds to a dictionary where each key is the `minion_id` and the value is the pillar data.

Long story short, technical reasons.

smithfarm:
@Martin-Weiss: I opened another PR #679 implementing @rjfd 's suggestion. If you agree, this one could be closed.

swiftgist:
#679 is merged.

Issue #657, opened 2017-09-20
stage.0: Module function cephprocesses.wait threw an exception. Exception: 'openattic'


Martin-Weiss:
On an already deployed cluster I get this error when executing stage.0:

```
  Name: wait until the cluster has recovered before processing ses-5-single.emea.utopia.novell.com - Function: salt.state - Result: Changed Started: - 09:08:20.869260 Duration: 6408.758 ms
----------
          ID: check if all processes are still running after processing ses-5-single.emea.utopia.novell.com
    Function: salt.state
      Result: False
     Comment: Run failed on minions: ses-5-single.emea.utopia.novell.com
              Failures:
                  ses-5-single.emea.utopia.novell.com:
                  ----------
                            ID: wait processes
                      Function: module.run
                          Name: cephprocesses.wait
                        Result: False
                       Comment: Module function cephprocesses.wait threw an exception. Exception: 'openattic'
                       Started: 09:08:27.525172
                      Duration: 83.194 ms
                       Changes:

                  Summary for ses-5-single.emea.utopia.novell.com
                  ------------
                  Succeeded: 0
                  Failed:    1
                  ------------
                  Total states run:     1
                  Total run time:  83.194 ms
     Started: 09:08:27.278118
    Duration: 350.72 ms
     Changes:

Summary for ses-5-single.emea.utopia.novell.com_master
-------------
Succeeded: 14 (changed=8)
Failed:     1
-------------

```
The openattic.service us up an running well and I can access openATTIC with a web-browser without any problems.

Any idea what might be wrong, here?

swiftgist:
Did your cluster go into HEALTH_ERR?  The steps in Stage 0 for a minion are serialized for an already running cluster.  It's in /srv/salt/ceph/stage/0/minion/default.sls.  The ceph.wait state is simply paranoia on our part that the previous update on some minion caused an issue and the cluster did not recover.  We bail out.

Unfortunately, the HEALTH_ERR status isn't terribly granular so we do not have a systematic guarantee of correlating cause and effect.

Martin-Weiss:
> Did your cluster go into HEALTH_ERR?

No - the cluster and all services are up and running well and I am testing to go through stages 0..5 where I do not expect failures in case I did not change anything.

Any idea how to get more debug information, here?

jschmid1:
@Martin-Weiss 

There is a module called cephprocesses.py which checks for services to be up. 

You can try it either on the respective node with:

[...]

PR #661, opened 2017-09-20
Add openattic to list of checked processes

jschmid1:
EDIT: moved rgw-ssl role addition to #680

~Furthermore~ we don't respect openattic at all although I was pretty sure we had that in. Did i miss the reason to drop it?

### DIFF
--- a/srv/salt/_modules/cephprocesses.py
+++ b/srv/salt/_modules/cephprocesses.py
@@ -26,6 +26,7 @@ def check(**kwargs):
                  'rgw': ['radosgw'],
                  'ganesha': ['ganesha.nfsd', 'rpcbind', 'rpc.statd'],
                  'admin': [],
+                 'openattic': ['openattic'],
                  'master': []}
 
     running = True


jschmid1:
#657 might be related

theanalyst:
I guess openattic was never in this file, (probably something we missed for a long time and no one found since they hadn't run stage 0 again after stage4 maybe?)

Can you explain why the rgw_conf. iterations doesn't populate this?

jschmid1:
iirc there is now the usecase where you can specify `rgw_init: default-ssl` and add `rgw-ssl` role to the policy.cfg but never add custom `rgw_configurations`. I'm not a 100% sure about this assumptions as so many things are changing in rgw right now :)

theanalyst:
yeah that wouldn't work by default, unless they bring their own ceph.conf in the ceph.conf.d dir. but it is okay to add this to the list since we're shipping it in the rpm anyway

smithfarm:
The health-openattic.sh test is reproducing #657 now on master. Presumably the failure will go away once this is merged.

swiftgist:
I checked and rgw-ssl is not a role.  It's really rgw with a different default.  So, the question is "which way are we going?".  I did not think we were making rgw-ssl a proper role.

For this PR, technically we could just add both and remove it later, but I think it adds to the confusion when you see it in this function but missing from the populate runner.

smithfarm:
@jschmid1 Could you split this into two PRs - one for oA and another for rgw-ssl? It appears only the rgw-ssl part is controversial.

jschmid1:
@smithfarm done

PR #662, opened 2017-09-20
Run stage.0 after deploying as bevahiour changes

jschmid1:
Stage.0 behaves differently (executes updates sequentially and checks for cluster health and service status). To increase the exercised codepath we should run this _after_ the initial deployment.

### DIFF
--- a/qa/suites/basic/health-igw.sh
+++ b/qa/suites/basic/health-igw.sh
@@ -72,5 +72,6 @@ ceph_health_test
 iscsi_kludge # see bsc#1049669
 igw_info
 iscsi_mount_and_sanity_test
+run_stage_0 "$CLI"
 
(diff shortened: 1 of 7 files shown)


jschmid1:
@smithfarm 
can we have test runs with that patch?

theanalyst:
This is good to have, this is how I found the other issue with rgw-configurations and cephprocesses as well

smithfarm:
@jschmid1 http://149.202.189.253:8081/ubuntu-2017-09-20_15:56:25-deepsea:basic-ses5---basic-openstack/ (ignore the dead jobs)

I looked at the health-ok log and could see that the first invocation of Stage 0 ran 14 jobs, while the second one (at the end) ran 25.

PR #663, opened 2017-09-20
Do not call rados if not installed

jschmid1:
resolves: 
https://bugzilla.suse.com/show_bug.cgi?id=1057062


srv/salt/_modules/cephimages.py
```diff
@@ -6,20 +6,30 @@
...
 from subprocess import Popen, PIPE
 
+def _is_installed(name):
+    try:
+        devnull = open(os.devnull)
+        Popen([name], stdout=devnull, stderr=devnull).communicate()
+    except OSError as e:
+        if e.errno == os.errno.ENOENT:
+            return False
+    return True
```
smithfarm: Does this actually execute the binary? That's fine for "rados" but if the binary does something like `rm -rf` (deliberate exaggeration) it might have "side effects". To find out if "foo" is installed and in the PATH, you can do "type foo" and examine the exit code:

```
vanguard1:/home/smithfarm # type rados
rados is hashed (/usr/bin/rados)
vanguard1:/home/smithfarm # echo $?
0
vanguard1:/home/smithfarm # type foo
bash: type: foo: not found
vanguard1:/home/smithfarm # echo $?
1
```

Anyway, this is "just an idea".

theanalyst: cant you use zypper module or the pkg module in this case?

srv/salt/_modules/cephimages.py
```diff
@@ -6,20 +6,30 @@
...
+    if _is_installed('rados'):
+        proc = Popen([ 'rados', 'lspools' ], stdout=PIPE, stderr=PIPE)
+        for line in proc.stdout:
+            pool = line.rstrip('\n')
+            cmd = [ '/usr/bin/rbd', '-p', pool, 'ls' ]
+            rbd_proc = Popen(cmd, stdout=PIPE, stderr=PIPE)
+            for rbd_line in rbd_proc.stdout:
+                if pool not in images:
+                    images[pool] = []
+                images[pool].append(rbd_line.rstrip('\n'))
```
theanalyst: we should probably used the librados and librbd python bindings in these cases, ok for now

srv/salt/_modules/cephimages.py
jschmid1: Although this is neither a public API, nor a user specified variable you are probably right in terms of not taking the risk. `type/which` has the same effect and reduces the overall risk and maybe even the invocation speed. 

updating.

srv/salt/_modules/cephimages.py
jschmid1: true. want to open an issue to track this, @theanalyst

srv/salt/_modules/cephimages.py
```diff
@@ -6,20 +6,31 @@
 from glob import glob
 from subprocess import Popen, PIPE
 
+def _is_installed(name):
+    try:
+        devnull = open(os.devnull)
+        cmd = "type ".format(name)
```
smithfarm: Further down we are constructing the cmd like this: `cmd = [ '/usr/bin/rbd', '-p', pool, 'ls' ]`

If I'm reading this correctly, here it will be something like `['type rados']` - would `['type', name]` be better?

srv/salt/_modules/cephimages.py
```diff
@@ -6,20 +6,31 @@
...
+        Popen([cmd], stdout=devnull, stderr=devnull).communicate()
```
smithfarm: in other words, this could be changed to `Popen(['type', name], ...`

srv/salt/_modules/cephimages.py
theanalyst: ack, created https://github.com/SUSE/DeepSea/issues/665

srv/salt/_modules/cephimages.py
```diff
@@ -6,20 +6,30 @@
 from glob import glob
 from subprocess import Popen, PIPE
 
+def _is_installed(name):
+    try:
+        devnull = open(os.devnull)
+        Popen(['type', name], stdout=devnull, stderr=devnull).communicate()
```
rjfd: Can't you use the `__salt__['pkg.info_installed']` module function?

The code would look like this:
```
try:
    res = __salt__['pkg.info_installed`](name)
    return name in res
except Exception:
    return False
```
Then for the case of checking the availability of the `rados` command, you can call this with `_is_installed("ceph-common")`

srv/salt/_modules/cephimages.py
rjfd: I think we can already check in this with the python bindings. Something like this:

```
images = {}
if not _is_installed('ceph-common'):
    return images
import rados
import rbd
cluster = rados.Rados(conffile='/etc/ceph/ceph.conf')
cluster.connect()
try:
    for pool in cluster.list_pools():
        ioctx = cluster.open_ioctx(pool)
        try:
            rbd_inst = rbd.RBD()
            images[pool] = rbd_inst.list(ioctx)
        finally:
            ioctx.close()
finally:
    cluster.shutdown()
return images
```

theanalyst: actually you can do a try catch on import itself, ie. 
```py
try:
    import rados
except ImportError:
...
```

swiftgist:
With the conditional, this will fail silently.  Is this mainly to get rid of the "[ERROR]: Rados could not be loaded" messages when the rpm is installed and the minions initially load the module?

swiftgist:
Ignore the part about "Rados could not be loaded".  The concern I have about failing silently is that the return will be an empty structure that the Ceph mine will cache.  Will that not cause an issue for oA when trying to list the images?

jschmid1:
please run basic suite

kshtsk:
WARNING. This PR is not tested with teuthology, please run the tests again if required.

jschmid1:
> Ignore the part about "Rados could not be loaded". The concern I have about failing silently is that the return will be an empty structure that the Ceph mine will cache. Will that not cause an issue for oA when trying to list the images?

The current behavior is not much different. When the cluster is in an unclean state oA will not get any information and additionally leaves a zombie process behind.

Coming back to this after some time made me realize another potential issue:

When the cluster is in a bad state we return `{}` which would overwrite the current cached images, wouldn't it?
Rather then always returning `{}` in case of an issue we'd load the cache `cephimages.list` values before and return those instead.

kshtsk:
retest this please

Issue #652, opened 2017-09-19
validate.py reports conflict between openAttic and RGW even though there is none


Martin-Weiss:
It seems that the conflict handling in modules/runners/validate.py does not work as required for something like this:

`rgw frontends = "civetweb port=8080+8443s ssl_certificate=/etc/ceph/rgw.pem"`

--> here a conflict is reported because the match on "port=80" is valid.

Do we need to re-open this issue or do we need to create a new issue?

PR #666, opened 2017-09-20
validate: openattic validation succeeds for non 80 ports

theanalyst:
Use file.search, which supports a regex (since file.contains_regex was
deprecated), and only match port=80, earlier we use to match port=8000
which cause issue when oa and rgw are colocated. This still fails in an
obnoxious case when someone tries ssl over port 80, since 80s would
still not match, and pass validate, and would try to install oa and
rgw-ssl with the same port 80 but 80s is a source of trouble for
many other cases anyway

Fixes: https://github.com/SUSE/DeepSea/issues/652
Reported-by: Martin Weiss <mweiss@suse.com>

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -600,7 +600,7 @@ class Validate(object):
                 'openattic' in self.data[node]['roles'] and
                 'rgw' in self.data[node]['roles']):
                 # Would use file.contains if it supported '='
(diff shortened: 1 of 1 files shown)


Martin-Weiss:
I am not a programmer - so do not really understand the \b - but does the new implementation match for port=80+8443s or port=8443s+80 or port=80s and does not match for port=800s+8000?

theanalyst:
so if port=80s is passed it would still succeed (whereas it is supposed to fail), in which case openattic would also be deployed at 80, all the other cases with `+` it still confirms to the valid behaviour

PR #671, opened 2017-09-21
monitoring: run rbd exporter only on fast-diff enabled images

jan--f:

### DIFF
--- a/srv/salt/ceph/monitoring/prometheus/exporters/files/rbd.sh
+++ b/srv/salt/ceph/monitoring/prometheus/exporters/files/rbd.sh
@@ -11,10 +11,10 @@ function clean_up {
 }
 
 if mkdir /var/lock/$me.lock; then
-    echo "$me lock succeeded"
+    (>&2 echo "$me lock succeeded")
     trap clean_up EXIT SIGINT SIGTERM
 else
-    echo "$me couldn't take lock...giving up"
+    (>&2 echo "$me couldn't take lock...giving up")
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


srv/salt/ceph/monitoring/prometheus/exporters/files/rbd.sh
```diff
@@ -37,9 +37,13 @@ do
     pools="$(ceph -c $conf osd lspools 2>/dev/null | sed 's/[[:digit:]] \([^,]*\),/\1 /g')"
     for pool in $pools
     do
-        rbd -p $pool --format json -c $conf du 2>/dev/null |
-        jq ".images[] | [.name, .provisioned_size, .used_size] | @csv" |
-        sed -e 's/^"//' -e 's/"$//' -e 's/\\"//g' | # stripe some quotes
-        awk -F , -v cluster=$cluster -v pool=$pool "$awk_cmd"
+        for img in `rbd -p $pool ls`; do
+            if rbd -p $pool info $img | grep "features:.*fast-diff.*" > /dev/null; then
```
rjfd: you can use `grep -q` option and avoid redirecting the stdout to `/dev/null`

`rbd -p $pool info $img | grep -q "features:.*fast-diff.*"`

jan--f:
@rjfd thx good catch.

swiftgist:
If anybody is ever looking for things to do upstream, I think having the rbd command support an option that doesn't requiring invoking three separate binaries to massage the data would help in the maintenance area.

PR #673, opened 2017-09-21
cli: Fixes root privileges checking

rjfd:
The existing root checking procedure was too conservative. Even `--help` messages were not possible to see if the user was not root.

This PR fixes the above problem.


smithfarm:
Tested locally in the sense that I can run `deepsea --help` as non-root now +1

Also running a suite on it: http://149.202.189.253:8081/ubuntu-2017-09-22_13:25:01-deepsea:basic-ses5---basic-openstack/

PR #679, opened 2017-09-22
validate: detect DEV_ENV=true from pillar in single minion case

smithfarm:
In case there is only one single minion (test-environment with everything on a
single server) and DEV_ENV=true is being set in the pillar, the result of
len(self.data.keys()) is 1 and with that this is not greater than 1 and
DEV_ENV=true is not detected.

Reported-by: Martin Weiss <Martin.Weiss@suse.com>


smithfarm:
Continuation of #660

PR #680, opened 2017-09-22
Add rgw-ssl to list of checked processes

jschmid1:
In a rgw-ssl deployment we define an actual[ 'role-rgw-ssl' ](https://github.com/SUSE/DeepSea/blob/1fcfea497493c9b3931dd8858c3d8345291dcb5e/qa/common/common.sh#L251)in the policy.cfg & pillar.

We already do add custom rgw_configurations [here](https://github.com/SUSE/DeepSea/blob/master/srv/salt/_modules/cephprocesses.py#L34) but iterate over all [roles](https://github.com/SUSE/DeepSea/blob/master/srv/salt/_modules/cephprocesses.py#L34) resulting in a key error.


jschmid1:
Not needed anymore, reopening if needed

PR #683, opened 2017-09-22
cli: gracefully handle non-started skipped steps

rjfd:
This PR handles the case where some step of the execution was parsed by the CLI parser, but during salt-run execution such step is skipped.

### DIFF
--- a/cli/monitor.py
+++ b/cli/monitor.py
@@ -632,16 +632,33 @@ class Monitor(threading.Thread):
 
         if nstep:
             if isinstance(step, Stage.TargetedStep):
-                logger.info("State step finished without 'ret' event: [%s/%s] name=%s(%s) on=%s",
-                            step.order,
-                            self._running_stage.total_steps(), step.name, step.args_str,
-                            step.targets.keys())
-                self._fire_event('step_state_finished', step)
+                if not step.jid:
+                    # step not even started
+                    step.skipped = True
+                    logger.info("Skipping non started state step: [%s/%s] name=%s(%s)",
+                                step.order, self._running_stage.total_steps(), step.name,
+                                step.args_str)
+                    self._fire_event('step_state_started', step)
+                else:
+                    logger.info("State step finished without 'ret' event: [%s/%s] name=%s(%s) "
+                                "on=%s",
+                                step.order,
+                                self._running_stage.total_steps(), step.name, step.args_str,
+                                step.targets.keys())
+                    self._fire_event('step_state_finished', step)
             else:
-                logger.info("Runner step finished without 'ret' event:: [%s/%s] name=%s(%s)",
-                            step.order, self._running_stage.total_steps(), step.name,
-                            step.args_str)
-                self._fire_event('step_runner_finished', step)
+                if not step.jid:
+                    # step not even started
+                    step.skipped = True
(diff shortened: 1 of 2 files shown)


rjfd:
@tserong good question! I think my choice of words to describe this PR was not the best. Here's the full story:
 
The CLI parses the SLS files, starting always from an orchestration file, to delineate a linear plan of execution formed by a list of steps. Each step corresponds to a runner execution or a state execution.

There are two situations where execution steps are skipped at runtime:
* a) when the next step to execute in our plan contains pre-conditions that are not satisfied. For instance, whenever a state declaration contains a `require: { salt: state_id }` primitive. This state only executes if `state_id` was successful.
* b) when the CLI parses the SLS files, some state declaration that is guarded by a jinja conditional block, appears in our execution plan, but when effectively running the stage the same state declaration is removed from execution.

Case a) was already being handled in [cli/monitor.py:689-699](https://github.com/SUSE/DeepSea/blob/master/cli/monitor.py#L689-L699) and the way we handle is by calling the `step_(runner | state)_started` with a step that has a boolean attribute named `skipped`, which in this case has the value `True`. The contract for the listeners of the `step_(runner | state)_started` is that if `step.skipped == True` then no `step_(runner | state)_finished` will be called on that step.

Case b) is a situation that ideally should never occur. If it occurs is because the evaluation of the jinja code is producing external side-effects that change the evaluation results in each run. Since, I was already handling Case a), I decided to handle this case the same way. 

The bad choice of words is due to that in both cases the events were never started, which makes the event name we used to signal skipped events a bit awkward.

Do you think it's worth to create a new event called `step_(runner | state)_skipped` instead of using the `step_(runner | state)_started` in conjunction with the step's `skipped` attribute?

swiftgist:
Looking from the outside, I think step_*_skipped is less confusing and matches the intentions.  

I do not know enough about the inner workings, but if states or runners are absent, then why fire an event at all?  My assumption is that the monitor will not display starting/finished messages for absent states/runners.  If the event is absent, wouldn't that just work out?

rjfd:
> Looking from the outside, I think step_*_skipped is less confusing and matches the intentions.

Yeah, I think I'll create a new event to the skipped steps.

> I do not know enough about the inner workings, but if states or runners are absent, then why fire an event at all?

From the implementation point of view, we are aware of skipped steps and therefore we should make them visible to the "outputters" implementation. In terms of progress visualization we may ignore such "skipped steps" events, but IMO they are relevant and useful for the user.

rjfd:
@tserong @swiftgist pushed a commit that adds the `step_*_skipped` event for notifying about skipped steps.

rjfd:
@smithfarm could you run the teuthology suite with this PR to check if it does not break anything. Thanks!

smithfarm:
full run on master (not including this PR): http://149.202.180.39:8081/ubuntu-2017-09-26_12:54:19-deepsea:basic-ses5---basic-openstack/

cli run on master + this PR: http://149.202.180.39:8081/ubuntu-2017-09-26_12:58:25-deepsea:basic-ses5---basic-openstack/

smithfarm:
@rjfd Unfortunately, the GRA1 region of OVH is acting up. This happens from time to time. Trying a different region.

PR #684, opened 2017-09-23
filesystem inspection and btrfs subvol creation/migration

kmroz:
The actual goal of this work is to help those using btrfs, to create/migrate the Ceph variable state directory to a btrfs subvolume.  The hope is that it can be extended for other filesystem related things in the future.

Runner adds:
`salt-run fs.inspect_var` - Inspection of /var/lib/ceph
`salt-run fs.create_var` - Creation of /var/lib/ceph as a btrfs subvolume
`salt-run fs.migrate_var` - Migration of existing /var/lib/ceph to a btrfs subvolume
`salt-run fs.correct_var_attrs` - Removing copy-on-write from /var/lib/ceph and it's children, while omitting /var/lib/ceph/osd/*

Use:
Automatic: `salt-run state.orch ceph.migrate.subvolume`
Controlled: `salt-run fs.inspect_var` and read/follow recommendations

Requires some testing beyond my 4 node virtual cluster.  Main pain points are actually with `systemctl start/stop` and determining if Ceph processes have indeed been terminated/started (hence the various sleeps and loops/checks/timeouts, etc).

### DIFF
--- a/deepsea.spec
+++ b/deepsea.spec
@@ -152,6 +152,7 @@ systemctl try-restart salt-api > /dev/null 2>&1 || :
 %dir /srv/salt/ceph/migrate/nodes
 %dir /srv/salt/ceph/migrate/osds
 %dir /srv/salt/ceph/migrate/policy
+%dir /srv/salt/ceph/migrate/subvolume
 %dir /srv/salt/ceph/mines
 %dir /srv/salt/ceph/mines/files
 %dir /srv/salt/ceph/mon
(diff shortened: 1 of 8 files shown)


swiftgist:
Once I sync'd all the modules, the "salt-run state.orch ceph.migrate.subvolume" ran fine.  I had two nodes that had to be created and the remainder had to be migrated.

The /srv/salt/ceph/migrate/subvolume is missing from the deepsea.spec file.

kmroz:
@swiftgist - Thanks, spec updated. I also added some man page entries, but let's discuss those (ie. not sure I want to expose _all_ fs modules for casual use).

Issue #681, opened 2017-09-22
Rename deepsea-aio-max-nr.conf


swiftgist:
Prefix the file with 99-.

PR #689, opened 2017-09-25
Prefix for sorting

swiftgist:


swiftgist:
Fixes #681

smithfarm:
Upstream ceph has https://github.com/ceph/ceph/pull/17894 staged - maybe we should take it instead?

swiftgist:
@smithfarm That's fine when it gets here.  I think the 1M is pretty much a guess either way and if the admin wants to change it, having this one could make that easier.

smithfarm:
@swiftgist Won't it be confusing for the admin when the same configuration parameter is set by two configuration files, one added by Ceph and the other by DeepSea? Which of the two will take precedence?

swiftgist:
@smithfarm A few questions: why are we setting this on anything other than storage nodes?  The only known issue is some storage nodes with a decent number of disks, but the Ceph package will set it on gateways, etc.

Next, we set it to 1M.  It will likely be fine, but what if we're wrong?  I expect that someone will say that's the admin's job to set a  new value, manage the file and push it out to the affected minions.  I do not believe anybody would be interested in repackaging Ceph to change the value.  I do think these admins would be encouraged to use Salt.  Why make them figure it out?

The files sort naturally and the last one takes precedence.  In this case, no real effect. If somebody finds it confusing, we can remove the ceph.sysctl from Stage 3, but keep the state around.

smithfarm:
@swiftgist Those questions might stand a better chance of answer/discussion if you posed them in https://github.com/ceph/ceph/pull/17894 . . .

ddiss:
> why are we setting this on anything other than storage nodes? The only known issue is some storage nodes with a decent number of disks, but the Ceph package will set it on gateways, etc.

IIUC, the upstream change adds `90-ceph-osd-sysctl.conf` to the ceph-osd package only, which would ideally only be installed on storage nodes.

> Next, we set it to 1M. It will likely be fine, but what if we're wrong? I expect that someone will say that's the admin's job to set a new value, manage the file and push it out to the affected minions. I do not believe anybody would be interested in repackaging Ceph to change the value. I do think these admins would be encouraged to use Salt. Why make them figure it out?

IMO an incorrect value configured by the package should be fixed via a package update.

> The files sort naturally and the last one takes precedence. In this case, no real effect. If somebody finds it confusing, we can remove the ceph.sysctl from Stage 3, but keep the state around.

I'd like to see the DeepSea functionality dropped in future, in favour of using the upstream Ceph changes. That said, it should be fine to have both merged for now, let's just make sure both changes are tracked.

Issue #688, opened 2017-09-25
deepsea CLI breaks when LC_ALL=C 


fghaas:
The deepsea CLI appears to choke on the unicode check mark (✓) when `LC_ALL` is set to C.

```
$ sudo deepsea stage run ceph.stage.prep
Starting stage: ceph.stage.prep
Parsing ceph.stage.prep steps... Traceback (most recent call last):
  File "/usr/bin/deepsea", line 9, in <module>
    load_entry_point('deepsea==0.7.27+git.0.274c55d', 'console_scripts', 'deepsea')()
  File "/usr/lib/python2.7/site-packages/deepsea/deepsea.py", line 311, in main
    cli(prog_name='deepsea')
  File "/usr/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/deepsea/deepsea.py", line 269, in stage_run
    ret = run_stage(stage_name, hide_state_steps, hide_dynamic_steps, simple_output)
  File "/usr/lib/python2.7/site-packages/deepsea/stage_executor.py", line 77, in run_stage
    mon.parse_stage(stage_name)
  File "/usr/lib/python2.7/site-packages/deepsea/monitor.py", line 505, in parse_stage
    self._fire_event('stage_parsing_started', stage_name)
  File "/usr/lib/python2.7/site-packages/deepsea/monitor.py", line 569, in _fire_event
    getattr(listener, event)(*args)
  File "/usr/lib/python2.7/site-packages/deepsea/monitors/terminal_outputter.py", line 600, in stage_parsing_started
    PP.println(SP.WAITING)
  File "/usr/lib/python2.7/site-packages/deepsea/common.py", line 220, in println
[...]

PR #691, opened 2017-09-26
cli: Use ASCII when no UTF-8 glyphs are supported

rjfd:
Fixes: #688

### DIFF
--- a/cli/common.py
+++ b/cli/common.py
@@ -53,6 +53,20 @@ def requires_root_privileges(func):
     return func_wraper
 
 
+def check_terminal_utf8_support():
+    """
+    Checks whether the terminal supports UTF-8 glyphs.
+    """
+    symbol = u"\u23F3"
+    try:
+        sys.stdout.write(symbol)
+        sys.stdout.write("\x1B[K\n")
+        sys.stdout.write("\x1B[A")
+        return True
+    except UnicodeEncodeError:
+        return False
+
+
 class PrettyPrinter(object):
     """
     Helper class to pretty print
--- a/cli/monitors/terminal_outputter.py
+++ b/cli/monitors/terminal_outputter.py
@@ -13,7 +13,7 @@ import re
 import threading
 import time
 
-from ..common import PrettyPrinter as PP
+from ..common import PrettyPrinter as PP, check_terminal_utf8_support
 from ..monitor import MonitorListener
 from ..stage_parser import StateRenderingException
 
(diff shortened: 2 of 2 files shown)


rjfd:
Output in a terminal with no support for UTF-8 glyphs:

deepsea-cli_no_utf8

cli/common.py
```diff
@@ -53,6 +53,20 @@ def func_wraper(*args, **kwargs):
     return func_wraper
 
 
+def check_terminal_utf8_support():
+    """
+    Checks weather the terminal supports UTF-8 glyphs.
```
smithfarm: s/weather/whether/

cli/common.py
```diff
@@ -53,6 +53,20 @@ def func_wraper(*args, **kwargs):
...
+    """
+    symbol = u"\u23F3"
+    try:
+        sys.stdout.write(symbol)
+        sys.stdout.write("\x1B[K\n")
+        sys.stdout.write("\x1B[A")
+        return True
+    except UnicodeEncodeError:
+        return False
```
smithfarm: Nice :-) I was wondering how you were going to accomplish this check.

cli/monitors/terminal_outputter.py
```diff
@@ -279,9 +279,9 @@ class StepListPrinter(MonitorListener):
     uses its own thread to allow the output of time clock counters for each step
     """
     # pylint: disable=C0103
-    OK = PP.green(PP.bold(u"\u2713"))
-    FAIL = PP.red(u"\u274C")
-    WAITING = PP.orange(u"\u23F3")
+    OK = PP.green(PP.bold(u"\u2713")) if check_terminal_utf8_support() else PP.green("OK")
+    FAIL = PP.red(u"\u274C") if check_terminal_utf8_support() else PP.red("Fail")
+    WAITING = PP.orange(u"\u23F3") if check_terminal_utf8_support() else PP.orange("Waiting")
```
smithfarm: I guess it's not really a performance problem to call this function three times in a row, but since the output is guaranteed (?) not to change between calls, wouldn't it be cleaner to call it just once and save the output in a local variable?

rjfd: Yes, you're right, I'll change that.

cli/common.py
rjfd: ah, I never remember how to write that word correctly :-)

rjfd:
@smithfarm pushed fixes to your comments

PR #699, opened 2017-09-28
Fix stack.cfg, move pillar examples to doc

swiftgist:


swiftgist:
@tserong Added linebreaks :)

Issue #694, opened 2017-09-26
query_disktype is not checking exit code of smartctl


matejzero:
I was trying to generate proposal where one SSD on the server would be used as wal, but salt was detecting my SSD as rotational (HDD).

With a lot of help from eric on #ceph, we discovered there is a bug in `query_disktype` function in `cephdisks.py`.

Looking at the code, you are not checking for smartctl exit code and only output you are checking is "A mandatory SMART command failed".

In my case, smartctl error was:
```bash
/usr/sbin/smartctl -i /dev/sdac -d megaraid,27
smartctl 6.2 2013-11-07 r3856 [x86_64-linux-4.4.82-6.9-default] (SUSE RPM)
Copyright (C) 2002-13, Bruce Allen, Christian Franke, www.smartmontools.org

Smartctl open device: /dev/sdac [megaraid_disk_27] failed: INQUIRY failed
```

In my configuration, I have a system drive on megaraid controller and a HBA, where JBOD is attached and all drives are in the JBOD, so we shouldn't pass `-d ...` parameter to `smartctl`.

Problematic piece of code (https://github.com/SUSE/DeepSea/blob/master/srv/salt/_modules/cephdisks.py):
```python
        try:
            cmd = "{} -i /dev/{} -d {},{}".format(smartctl_path,
                                                  device,
                                                  raid_ctrl['controller_name'],
                                                  bus_id)
            print(cmd)
            proc = Popen(cmd, stdout=PIPE, stderr=PIPE, shell=True)
            for line in proc.stdout:
                # ADD PARSING HERE TO DETECT FAILURE
                if "A mandatory SMART command failed" in line:
                    log.warn("Something went wrong during smartctl query")
                m = re.match("([^:]+): (.*)", line)
                if m:
                    if (m.group(1) == "Rotation Rate"):
                        c = re.match("^\s+ Solid State Device", m.group(2))
                        if c:
                            return '0'
            return '1'
```

It would probably be good to do some exit code checking and run `_is_rotational` if smartctl fails.

swiftgist:
Fixed in #700

PR #700, opened 2017-09-28
Check returncode, fallback to default detection

swiftgist:


swiftgist:
Intended to fix #694

swiftgist:
@jschmid1 I took a look at the unit tests and helper/output.py and I'll say that it's a bit of mental gymnastics to determine what I should put where.  Changing the returncode isn't enough because I need to fallback to the default detection.

swiftgist:
Added unit test to check that is_rotational is called.

PR #701, opened 2017-09-28
Remove redundant error message

swiftgist:


swiftgist:
The error message is triggered by the Salt minion loading the file.  Since the individual functions already produce an error message, removing this one seems the simplest.

PR #706, opened 2017-10-03
Fix Makefile, spec file for man installation

swiftgist:


swiftgist:
Fix for final build.

PR #716, opened 2017-10-05
benchmarks: use client_glob for baseline benchmark

mogeb:
Signed-off-by: Mohamad Gebai <mgebai@suse.com>

### DIFF
--- a/srv/modules/runners/benchmark.py
+++ b/srv/modules/runners/benchmark.py
@@ -252,9 +252,17 @@ def baseline(margin=10, verbose=False, **kwargs):
     trigger 'ceph tell osd.$n bench' on all $n OSDs and check the results for
     slow outliers
     '''
-    # get all osd ids
-    osd_list = local_client.cmd('I@cluster:ceph and I@roles:storage',
-                                'osd.list', [], expr_form='compound')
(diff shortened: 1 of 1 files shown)


srv/modules/runners/benchmark.py
```diff
@@ -252,9 +252,17 @@ def baseline(margin=10, verbose=False, **kwargs):
...
+    client_glob = kwargs.get('client_glob',
+                             'I@roles:storage and I@cluster:ceph')
+    log.info('client glob is {}'.format(client_glob))
+
+    # get all osd ids for a given client_glob
+    osd_list = local_client.cmd(client_glob,
+        'osd.list', [], expr_form='compound')
+
+    if not osd_list:
+        raise Exception('No minions found for glob {}'.format(client_glob))
```
jan--f: Maybe change this exception message. osd_list will also be empty if the minions that match client_glob don't return any OSDs. Maybe ```No OSDs found on minons matching {}```

mogeb: Right, good point. Will do.

mogeb: Done.

Issue #720, opened 2017-10-06
CLI: sometimes not all steps are reported as passed, though none failed


smithfarm:
As reported by @kshtsk - in the summary printed by `deepsea stage run` sometimes not all of the steps are reported as passed - for example:

```
16:26:45 + deepsea --log-file=/var/log/salt/deepsea.log --log-level=debug stage run ceph.stage.0 --simple-output
16:26:45 Starting stage: ceph.stage.0
16:26:45 Parsing stage ceph.stage.0 steps... done
16:28:31 
16:28:31 Stage initialization output:
16:28:31 deepsea_minions          : valid
16:28:31 master_minion            : valid
16:28:31 ceph_version             : valid
16:28:31 
16:28:31 [init] Executing runner validate.setup... 
16:28:31          |_  cmd.shell(/usr/bin/zypper info ceph)... 
16:28:31                in staging-target-213-032-031-203.ovh.suse.de... ok
16:28:31                in staging-target-213-032-030-092.ovh.suse.de... ok
16:28:31                in staging-target-167-114-224-035.ovh.suse.de... ok
16:28:31                in staging-target-213-032-031-092.ovh.suse.de... ok
16:28:31 [init] Executing runner validate.setup... ok
16:28:31 [init] Executing runner advise.salt_run... ok
16:28:31 [init] Executing runner cephprocesses.mon... ok
16:28:31 
16:28:31 [1/14] Executing state ceph.salt-api... 
16:28:31          |_  staging-target-213-032-031-203.ovh.suse.de: salt-api ok
16:28:31          staging-target-213-032-031-203.ovh.suse.de... ok
16:28:31 [2/14] Executing state ceph.sync... 
16:28:31          |_  staging-target-213-032-031-203.ovh.suse.de: load modules(saltutil.sync_all) ok
16:28:31          staging-target-213-032-031-203.ovh.suse.de... ok
16:28:31 [3/14] Executing state ceph.repo... 
16:28:31          staging-target-213-032-031-203.ovh.suse.de... ok
16:28:31 [4/14] Executing state ceph.updates... 
16:28:31          |_  staging-target-213-032-031-203.ovh.suse.de: packagemanager patch only kernel(packagemanager.up) ok
16:28:31          |_  staging-target-213-032-031-203.ovh.suse.de: packagemanager update regular(packagemanager.up) ok
16:28:31          staging-target-213-032-031-203.ovh.suse.de... ok
[...]

PR #727, opened 2017-10-10
stage 0: update mines using mines.update instead of minion restart

rjfd:
The current behavior of updating the salt mines in stage 0 using `service.running` module is making the salt-minion to lose the job information of the following step, which is `ceph.updates`, because it restarts the salt-minion. The consequence is that stage 0 finishes the execution before the ceph.updates job is finished.

This PR fixes the issue by calling the `mine.update` module instead.

### DIFF
--- a/srv/salt/ceph/mines/default.sls
+++ b/srv/salt/ceph/mines/default.sls
@@ -1,15 +1,20 @@
 
 
-configure_mines:
+configure_mine_functions_conf:
   file.managed:
     - name: /etc/salt/minion.d/mine_functions.conf
     - source: salt://ceph/mines/files/mine_functions.conf
     - fire_event: True
 
+add_mine_cephdisks.list_to_minion:
+  module.run:
+    - name: mine.send
+    - func: cephdisks.list
+
 manage_salt_minion_for_mines:
-  service.running:
-    - name: salt-minion
+  module.run:
+    - name: mine.update
     - watch:
-      - file: configure_mines
+      - file: configure_mine_functions_conf
     - fire_event: True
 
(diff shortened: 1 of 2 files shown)


smithfarm:
@rjfd add `Fixes: https://github.com/SUSE/DeepSea/issues/720` to the commit message?

rjfd:
@smithfarm yes, it is supposed to fix that issue. Will add that to the commit message

rjfd:
I just find out that this patch does not fix the problem. I'll update this PR with another solution.

rjfd:
@swiftgist  I just pushed a new fix for this issue.
@smithfarm can you trigger the tests in teuthology?

smithfarm:
Test results:

http://213.32.28.133:8081/ubuntu-2017-11-20_06:47:34-deepsea:basic-ses5---basic-openstack/

swiftgist:
Okay, so the general lesson here is not to use certain features for Salt itself.  The original configuration came from https://docs.saltstack.com/en/latest/ref/states/requisites.html#watch

rjfd:
@swiftgist the problem is not related with the `watch` requisite, but rather with the fact that we were restarting the salt-minion within a state execution.

swiftgist:
Fair enough, but if you have to add a mine.send in addition to the mine.update call to make the minion catch that there's been a configuration change since we cannot restart the salt-minion, then the typical watch command example does not really apply.

I believe Salt from a year ago used the watch example for restarting daemons only as needed in many talks.  Now, we have an "onchanges" keyword too.  Neither of these is addressing Salt configuration changes.

The longer term fear/issue is that the Salt master/minions cannot pickup configuration changes gracefully.  That is, we have no SIGHUP equivalent that exists for other services.  Eventually, a dynamic change will not match the configuration file.  The admin will restart a Salt minion at some later date and have an issue that is unrelated to the reason for restarting the minion.

I think we are safe with this change since if the mine.send fails for some reason, then that node will appear to have no storage.  That will cause an immediate investigation.

PR #735, opened 2017-10-11
qa: Add basic restart mechanism for ceph.restart tests

jschmid1:
Implementing basic validation for restarted services
for now only: MON, OSD, RGW

Depends on: #734 

### DIFF
--- a/qa/common/common.sh
+++ b/qa/common/common.sh
@@ -113,6 +113,24 @@ function run_stage_5 {
 # see https://github.com/SUSE/DeepSea/tree/master/srv/salt/ceph/configuration/files/ceph.conf.d
 #
 
+function change_rgw_conf {
+    cat <<EOF >> /srv/salt/ceph/configuration/files/ceph.conf.d/rgw.conf
+foo = bar
+EOF
+}
+
+function change_osd_conf {
+    cat <<EOF >> /srv/salt/ceph/configuration/files/ceph.conf.d/osd.conf
+foo = bar
+EOF
+}
+
+function change_mon_conf {
+    cat <<EOF >> /srv/salt/ceph/configuration/files/ceph.conf.d/mon.conf
+foo = bar
+EOF
+}
+
 function ceph_conf_small_cluster {
   local STORAGENODES=$(json_storage_nodes)
   test ! -z "$STORAGENODES"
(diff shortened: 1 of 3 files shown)


qa/common/common.sh
```diff
@@ -149,7 +167,39 @@ EOF
...
+#
+
+function restart_services {
+  salt-run state.orch ceph.restart
+}
+
+function mon_restarted {
+  local exit_status 
+  local expectation=$1
+  salt --static --out json -C 'I@roles:mon' cmd.shell "journalctl -u ceph-mon@* " | grep -i terminated
```
smithfarm: If we're telling Salt to generate JSON output, would it make more sense to use `jq` instead of `grep` to parse the output?

smithfarm:
@jschmid1 I have been meaning to ping you to discuss the possibility of integrating `health-ok-encrypted.sh` into `health-ok.sh` so the user would run `health-ok.sh --encrypted` instead of maintaining a separate script.

The same would apply to this. Instead of a separate script, could we do `health-ok.sh --restart`? (Especially now that the command-line parameter parsing infrastructure is there.)

qa/common/common.sh
smithfarm: In this case, if we're just checking for the presence of a string "terminated", we could just drop `--static --out json`?

smithfarm: What if there are multiple nodes with the `mon` role, and only one restarts? The test will pass?

qa/common/common.sh
```diff
@@ -149,7 +167,39 @@ EOF
...
+  test ${exit_status} = ${expectation}
+}
+
+function osd_restarted {
+  local exit_status 
+  local expectation=$1
+  salt --static --out json -C 'I@roles:storage' cmd.shell "journalctl -u ceph-osd@* " | grep -i terminated
+  exit_status={$?}
+  test ${exit_status} = ${expectation}
+}
```
smithfarm: Typically there are multiple `storage` nodes and each one has multiple `ceph-osd@*` services. For test to pass, it is enough if just one `ceph-osd@` unit restarts on one of the nodes?

jschmid1:
> @jschmid1 I have been meaning to ping you to discuss the possibility of integrating health-ok-encrypted.sh into health-ok.sh so the user would run health-ok.sh --encrypted instead of maintaining a separate script.
> 
> The same would apply to this. Instead of a separate script, could we do health-ok.sh --restart? (Especially now that the command-line parameter parsing infrastructure is there.)


I agree with having the --encrypted switch because this changes the deployment path. `--restart` however is an essential functionality that should always work and does not depend on a certain setup.

qa/common/common.sh
jschmid1: True; we could iterate over each OSD and invoke checks upon those.

qa/common/common.sh
jschmid1: It will pass, yes. Iterating over hosts will definitely improve the test. Let me add it.

smithfarm:
> --restart however is an essential functionality that should always work and does not depend on a certain setup.

In that case, can you just add the functionality to `health-ok.sh` instead of making a new test script?

jschmid1:
> In that case, can you just add the restart functionality to health-ok.sh instead of making a new test script?

I'd prefer that.

Issue #730, opened 2017-10-10
Stage 2 may need to do "zypper refresh" before installing prometheus packages


smithfarm:
See errors like this one occasionally when running integration tests in teuthology:

```
2017-10-10T18:29:08.083 INFO:teuthology.orchestra.run.target167114250015.stdout:[13/14] Executing state ceph.monitoring...
2017-10-10T18:29:08.083 INFO:teuthology.orchestra.run.target167114250015.stdout:         |_  select.minions(cluster=ceph, host=False)... ok
2017-10-10T18:29:08.083 INFO:teuthology.orchestra.run.target167114250015.stdout:         |_  select.minions(cluster=ceph, host=False)... ok
2017-10-10T18:29:08.083 INFO:teuthology.orchestra.run.target167114250015.stdout:         |_  select.minions(cluster=ceph, roles=rgw)... ok
2017-10-10T18:29:08.084 INFO:teuthology.orchestra.run.target167114250015.stdout:         |_  target167114250015.teuthology: golang-github-prometheus-prometheus fail
2017-10-10T18:29:08.084 INFO:teuthology.orchestra.run.target167114250015.stdout:         |_  target167114250015.teuthology: /etc/prometheus/prometheus.yml ok
2017-10-10T18:29:08.084 INFO:teuthology.orchestra.run.target167114250015.stdout:         |_  target167114250015.teuthology: /etc/prometheus/ses_nodes/target167114250015.teuthology.yml ok
2017-10-10T18:29:08.084 INFO:teuthology.orchestra.run.target167114250015.stdout:         |_  target167114250015.teuthology: /etc/prometheus/ses_nodes/target167114250143.teuthology.yml ok
2017-10-10T18:29:08.084 INFO:teuthology.orchestra.run.target167114250015.stdout:         |_  target167114250015.teuthology: golang-github-prometheus-alertmanager ok
2017-10-10T18:29:08.084 INFO:teuthology.orchestra.run.target167114250015.stdout:         |_  target167114250015.teuthology: grafana ok
2017-10-10T18:29:08.085 INFO:teuthology.orchestra.run.target167114250015.stdout:         target167114250015.teuthology... fail
2017-10-10T18:29:08.085 INFO:teuthology.orchestra.run.target167114250015.stdout:[14/14] Executing state ceph.monitoring.prometheus.exporters.node_exporter...
[...]

PR #738, opened 2017-10-12
refresh repo database before installing pkgs

jschmid1:
potential bugs:
* [prometheus package](https://github.com/SUSE/DeepSea/compare/master...jschmid1:wip-refresh-pgk-installed?expand=1#diff-9c15a59ff451861ec015955c08e85fcbR4)
* [grafana package](https://github.com/SUSE/DeepSea/compare/master...jschmid1:wip-refresh-pgk-installed?expand=1#diff-b7cf5b0b97e32158e7f5efda4e9d73a9R3)

Did you leave the `name: <pkg>` parameter by intention?

Fixes: https://github.com/SUSE/DeepSea/issues/730

### DIFF
--- a/srv/salt/ceph/igw/default.sls
+++ b/srv/salt/ceph/igw/default.sls
@@ -3,6 +3,7 @@ lrbd:
   pkg.installed:
     - pkgs:
       - lrbd
+    - refresh: True
 
 enable lrbd:
   service.running:
--- a/srv/salt/ceph/igw/import/default.sls
+++ b/srv/salt/ceph/igw/import/default.sls
@@ -4,6 +4,7 @@ lrbd:
   pkg.installed:
     - pkgs:
       - lrbd
+    - refresh: True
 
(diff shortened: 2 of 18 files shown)


jan--f:
If ```pkg.installed``` doesn't find a name key, the state name is used...so should work either way.

smithfarm:
Why are we firing an event sometimes, and other times not? Does that have something to do with the CLI?

smithfarm:
@rjfd Would it make sense to fire an event every time `pkg.installed` runs?

rjfd:
@smithfarm yes, I think it makes sense. That way the CLI can show those intermediate steps to the user.
As a matter of fact, all states should fire events...

jschmid1:
@rjfd @smithfarm let's handle that with a different PR

rjfd:
@jschmid1 sure.

jschmid1:
created https://github.com/SUSE/DeepSea/issues/742

PR #744, opened 2017-10-16
Update grains unconditionally

swiftgist:

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -1698,20 +1698,15 @@ class OSDGrains(object):
         """
         Load and save grains when changed
         """
-        if storage:
-            content = {}
-            if os.path.exists(filename):
-                with open(filename, 'r') as minion_grains:
-                    content = yaml.safe_load(minion_grains)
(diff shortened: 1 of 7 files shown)


srv/salt/_modules/osd.py
```diff
@@ -1698,20 +1698,19 @@ def _grains(self, storage, filename="/etc/salt/grains"):
...
-                    content['ceph'] = storage
-                    self._update_grains(content)
-                else:
-                    log.debug("No update for {}".format(filename))
-            else:
+        content = {}
+        if os.path.exists(filename):
+            with open(filename, 'r') as minion_grains:
+                content = yaml.safe_load(minion_grains)
+        if 'ceph' in content:
```
jan--f: Could this be simplified like so:
```
if 'ceph' in content and content['ceph'] == storage:
    log.debug("No update for {}".format(filename))
else:
    content['ceph'] = storage
    self._update_grains(content)

jan--f: To replace line 1705 - 1713

PR #745, opened 2017-10-16
[RFC] Add RBD benchmarks

mogeb:
This is a Request For Comment, do not merge.


srv/salt/ceph/rbd/benchmarks/cleanup_master.sls
```diff
@@ -1,6 +1,12 @@
-delete benchmark image:
+delete benchmark pool:
   cmd.run:
-    - name: rbd rm deepsea_benchmark
+    - name: ceph osd pool rm {{ salt['pillar.get']('rbd_benchmark_pool', 'deepsea_rbd_benchmark') }} {{ salt['pillar.get']('rbd_benchmark_pool', 'ds_rbd_bench_pool') }} --yes-i-really-really-mean-it
```
mogeb: This depends on having the mon setting that allows pools to be deleted

srv/salt/ceph/rbd/benchmarks/create_image.sls
```diff
@@ -1,3 +1,11 @@
-create benchmark image:
+create benchmark pool:
   cmd.run:
-    - name: rbd create --size 2048 deepsea_benchmark
+    - name: ceph osd pool create {{ salt['pillar.get']('rbd_benchmark_pool', 'deepsea_rbd_benchmark') }} 256 256
```
mogeb: The pg num needs to be computed.

srv/salt/ceph/rbd/benchmarks/create_image.sls
```diff
@@ -1,3 +1,11 @@
...
+
+{% for host in salt.saltutil.runner('select.minions', cluster='ceph', roles='storage') %}
+create {{ host }} benchmark image:
+  cmd.run:
+    - name: rbd create --pool {{ salt['pillar.get']('rbd_benchmark_pool', 'deepsea_rbd_benchmark') }} --size 2048 {{ salt['pillar.get']('rbd_benchmark_image_prefix', 'rbd_bench_img_') }}{{ host }}
```
mogeb: Do we want to give the flexibility of changing the pool name?

srv/salt/ceph/rbd/benchmarks/create_image.sls
smithfarm: While we're at it, let's compute pg num for all pools DS creates.

Martin-Weiss:
Do I understand right that this is going to create an RBD with a size of 2 GB?
Does such a small size make sense for performance measurements?

I had several customers using 2 TB RBDs for their ESX datastores - so maybe we should create RBDs in 200 GB or 2 TB size?

mogeb:
@Martin-Weiss, good point, and yes I agree that 2 GB is definitely too small. Do you think 200 GB or 2 TB is good in general? Do you think it should be a function of the cluster size (number of nodes)?

Martin-Weiss:
> Do you think 200 GB or 2 TB is good in general? Do you think it should be a function of the cluster size (number of nodes) instead?

I believe we can not calculate this size automatically. All clusters have a different setup and size and due to that I assume we should have some sort of configuration file where we specify the parameters for the benchmarking. The RBD size would be one of the parameters - but also other parameters are required - i.e. which pool to use and which block sizes / how many clients / which clients / how many threads etc...

Background: Depending on the setup / environment we might have pools with EC, pools on SSD or NVMe, cache-tiers, different size settings, different hierarchies and rules in the crushmap etc. - and due to that we might need a larger set of parameters for "setup for the tests", "execution of the tests", "optional cleanup after the tests".

Also keep in mind that some of the tests might be destructive so we might also need some "are you sure"  checks..

srv/salt/_modules/calculate.py
```diff
@@ -0,0 +1,39 @@
...
+def _nearest_power_of_two(target):
+    if target <= 0:
+        raise Exception('Can\'t find nearest power of two of invalid value')
+
+    i = 0
+
+    # Based on the logic presented on http://ceph.com/pgcalc/
+    while True:
+        current = 2 ** i
+        next = 2 ** (i + 1)
```
jschmid1: Although `next` is not a keyword, it's part of python's built-ins and shouldn't be used as variable name. `next_` is probably fine.

srv/salt/ceph/rbd/benchmarks/cleanup_master.sls
jan--f: We could inject this flag, remove and reinstate the default? Though this would get more complex if the setting already exists.

srv/salt/ceph/rbd/benchmarks/create_image.sls
jan--f: hmm good question. I'd certainly recommend using a more expicit pool name (maybe ```rbd_deepsea_benchmark_pool```). We could also add this feature later...should be as simple as pulling a name from the pillar.

jschmid1:
add to whitelist

swiftgist:
conflicts resolved in #889

swiftgist:
Oops... closed wrong ticket

PR #748, opened 2017-10-18
openattic: Check oA config file in /etc/default and always overwrite settings

rjfd:
This PR is related to the oA issue: https://tracker.openattic.org/browse/OP-2808


sebastian-philipp:
see also: https://bitbucket.org/openattic/openattic/pull-requests/666/moved-suse-config-file-from-etc-sysconfig/diff

PR #756, opened 2017-10-26
qa: fix regression and rip out health-ok-encrypted.sh

smithfarm:
* fix regression introduced by 1e99b8e 
* rip out health-ok-encrypted.sh


jschmid1:
closing in favor of #757

PR #759, opened 2017-10-27
qa: health-ok.sh: liberally accept both --encrypted and --encryption

smithfarm:
This fixes a regression introduced by 01833ee983e6da8b41706c3819a07ce9a86b0f49


smithfarm:
Tested - the health-ok-encrypted test case passes when this fix is applied.

PR #760, opened 2017-10-27
Support for multiple ntp servers in ntp.conf

### LINKED ISSUE #752: Support a list of ntp servers
We should enable multiple servers for our ntp configuration.
We can simply check if time_server is a list and iterate over it if so.

jan--f:
Fixes #752


jschmid1:
@jan--f is this already in use by a major user?

PR #764, opened 2017-10-30
"NaN ns" in Dashboard (bsc1065590)

votdev:
To fix this issue a mapping for "0" => "0 ns" is added.

The other solution to set the decimal precision to 1 seems not to work, the widget still shows 'xxx s' for example instead of 'xxx.y s' as usual.

If the decimal precision is set to any value, Grafana displays 0 values as 'ns', so the mapping above with the unit 'ns' should be the best solution.

### DIFF
--- a/srv/salt/ceph/monitoring/grafana/files/ceph-cluster.json
+++ b/srv/salt/ceph/monitoring/grafana/files/ceph-cluster.json
@@ -1020,8 +1020,8 @@
             "valueMaps": [
               {
                 "op": "=",
-                "text": "N/A",
-                "value": "null"
+                "text": "0 s",
+                "value": "0"
               }
             ],
             "valueName": "avg"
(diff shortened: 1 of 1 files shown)


srv/salt/ceph/monitoring/grafana/files/ceph-cluster.json
```diff
@@ -1022,6 +1022,11 @@
                 "op": "=",
                 "text": "N/A",
                 "value": "null"
+              },
+              {
+                "op": "=",
+                "text": "0 ns",
```
jan--f: Lets use 's' as the base unit.

srv/salt/ceph/monitoring/grafana/files/ceph-cluster.json
```diff
@@ -1100,6 +1105,11 @@
...
```
jan--f: also seconds please

votdev:
The useless mappings have been removed after I've tested it by purging the metrics.

PR #765, opened 2017-10-30
Lint fixes

swiftgist:


jschmid1:
we should run a set of tests at the very least.. fixing 'syntax' in 58 files can very easy lead to indentation & rename errors..
@smithfarm could you launch a barrage of tests on this branch?

PR #780, opened 2017-11-08
add sortbitwise flag and split them up nicely

jschmid1:
Make sure to set the osd flag before we upgrade the cluster to not run
into http://tracker.ceph.com/issues/20416.


swiftgist:
This is only being set to handle upgrades from Pre-Jewel installations on the off-chance that sortbitwise is not already set.

PR #782, opened 2017-11-09
backporting iteration #2

jschmid1:
backporting to stable branch. See changelog for inclusions.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Nov 9 10:15:12 UTC 2017 - jschmid@suse.de
+
+- Version 0.8.01 bb9069893
+- prevents bsc#1067088
+- don't fall back to 'upgrade all' if no cluster exists
+- set the 'sortbitwise' osd flag before the upgrade
+
+-------------------------------------------------------------------
 Mon Nov 6 10:16:07 UTC 2017 - jschmid@suse.de
 
 - Version 0.8 ecbc24759d047be517a309c05674a0a7e9b5ddc2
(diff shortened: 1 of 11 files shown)


smithfarm:
@jschmid1 Does this PR need to have the fix for the 'ceph.packages.remove' issue, too? Elsewhere you mentioned it's included in 0.8 but the product is still on 0.7.35 afaict.

jschmid1:
@smithfarm Yes, it's included in 0.8 which I though he shipped in the first place.

The fix we are talking about is in https://github.com/SUSE/DeepSea/commit/c8af8e0db32ebd1ef7db7ddbd63933951b911c2e which is included in the SES5 branch and already staged for maintenance

jschmid1:
good observation, but we discussed this PR already. 

#780 is merged now.

carltonf:
@jschmid1 I note that in this MU preparation, #776 is not included. It still affects cases like bsc#1066502.

jecluis:
Folks, this is a public - upstream - repository. Talking about Maintenance Updates and referring internal bugzilla tickets that are way behind a required login and proper permissions is not cool.
I fully understand that we keep thinking about SUSE/DeepSea as something we only consume, but, please, let's pretend it's not and interact with it as if other people were consuming this project.

jschmid1:
@carltonf This is preparation PR which is still in progress. #776 i.e. has the 'backport' flag, which means that it will eventually be backported to the 'stable' branch. Due to the lack of time, I was not able to proceed with this task yet.

carltonf:
@jschmid1 Oh, thanks for the explanation. I didn't notice it and thought it got lost for all these works.

@jecluis Sure, you are right about that.

jschmid1:
closing, backporting each PR separately

PR #785, opened 2017-11-16
Fix is_prepared

swiftgist:
Correct the detection of an NVMe OSD


jschmid1:
And we should stick to our approach of merging fixes to master first and backport them later.

jschmid1:
the code looks good, have to verify it though

jschmid1:
#786 was merged in master.

PR #791, opened 2017-11-16
Add client processes to list of checked processes

dzedro:
Fix same error as in https://github.com/SUSE/DeepSea/pull/661 with client-cephfs, client-nfs, client-radosgw and client-iscsi roles

Signed-off-by: Jozef Pupava <jpupava@suse.com>


swiftgist:
The names may be changing in some cases soon, but that's fine.

PR #795, opened 2017-11-17
CentOS 7 support changes

rjfd:
This PR pushes changes to make DeepSea to work in CentOS 7 distro.

A demo can be found in:
https://asciinema.org/a/147996

I've created a github repo to hold the packaging files for fedora Copr. That repo is available in:
https://github.com/rjfd/deepsea_rpm
You can take a look if you want.

The RPM package for CentOS is available in:
https://copr.fedorainfracloud.org/coprs/rjdias/home/packages/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,26 @@
 # Override this to install docs somewhere else
 DOCDIR = /usr/share/doc/packages
 
+OS=$(shell source /etc/os-release ; echo $$ID)
+ifeq ($(OS), opensuse)
+USER=salt
+GROUP=salt
+PKG_INSTALL=zypper -n install
+else
+ifeq ($(OS), sles)
+USER=salt
+GROUP=salt
+PKG_INSTALL=zypper -n install
+else
+USER=root
+GROUP=root
+ifeq ($(OS), centos)
+PKG_INSTALL=yum install -y
+endif
+endif
+endif
+
+
 usage:
 	@echo "Usage:"
 	@echo -e "\tmake install\tInstall DeepSea on this host"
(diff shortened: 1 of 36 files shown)


swiftgist:
I think the generic replacements of the hardcoded zypper/rpm logic are fine.  I would like less Jinja in general.  For example,

```
{% if grains.get('os_family', '') == "Suse" %}
￼   {% set user = "salt" %}
￼   {% set group = "salt" %}
￼ {% else %}
￼   {% set user = "root" %}
￼   {% set group = "root" %}
￼ {% endif %}

```

in a Salt module could handle many distros without additionally changing each sls file.  Then

```
￼ - user: {{ user }}
￼ - group: {{ group }}
```

would become

```
￼ - user: {{ salt['master.user']() }}
￼ - group: {{ salt['master.group'() }}
```


For organization, I think the repo additions should end up in /srv/salt/ceph/repo.  I'm curious if there's any counter arguments.

For both repos and services, I think having a default.sls (or changing the init.sls in these directories) to incorporate `grains.get('os', '')`.  Something like

```
{% set default_service = salt['grains.get']('os', 'default') %}

include:
  - .{{ salt['pillar.get']('ganesha_service', default_service) }}
```
Then files in these directories would become CentOS.sls, SUSE.sls.  The above will need some additional thought because `default.sls` should work if the os version does not exist.  The one advantage I can see here is that the Salt admin can run the state file directly without any question of whether Jinja conditionals rendered correctly.  Also, the override still exists for those that need it.

Otherwise, I think it's a nice job and I will try to help with this if everyone thinks the above makes sense.

rjfd:
@swiftgist 

> would become
>```
>￼ - user: {{ salt['master.user']() }}
>￼ - group: {{ salt['master.group'() }}
>```

I agree. I also wasn't liking to copy and paste that jinja snippet whenever the user and group was needed.

> For organization, I think the repo additions should end up in /srv/salt/ceph/repo. I'm curious if there's any counter arguments.

My only argument against is that I think we should only install a repo if it's really needed. For instance, the repo we are installing for ganesha, if the user does not configured any ganesha role than we should not be installing any ganesha repo. All others repos that are required for packages always installed by DeepSea should be indeed put in `ceph/repo`.

> For both repos and services, I think having a default.sls (or changing the init.sls in these directories) to incorporate grains.get('os', ''). Something like

Sounds like a good way to organize the code and makes the extension to other distros easier.

I'll make these changes and push them to the PR.

Makefile
```diff
@@ -1,6 +1,26 @@
 # Override this to install docs somewhere else
 DOCDIR = /usr/share/doc/packages
 
+OS=$(shell egrep ^ID= /etc/os-release | sed -e 's/ID="\(.*\)"/\1/g')
```
smithfarm: @rjfd I always wonder, when confronted with code that parses `/etc/os-release` using sed, awk, etc., whether it would be possible to eliminate that code by simply sourcing `/etc/os-release` as was originally intended. For example

```
OS=$(source /etc/os-release ; echo $ID)
```

rjfd: Fixed!

rjfd:
@smithfarm do you care to run this PR in teuthology to check if does not break anything in SUSE distro?

smithfarm:
@rjfd First test run is here: http://213.32.28.133:8081/ubuntu-2017-11-21_16:19:21-deepsea:basic-ses5---basic-openstack/

The rgw failure is OVH infrastructure noise. The nfs-ganesha (RGW FSAL) failure needs investigation. For now I'm going to re-run it.

Note that your `wip-centos` branch is now slightly behind master. I pushed a `wip-centos-test` branch to smithfarm/DeepSea.git and will use that for the next run.

jschmid1:
please test this

jschmid1:
Detected a failure in 


>
>           ID: salt-api
>     Function: salt.state
>       Result: False
>      Comment: Run failed on minions: target158069088033.teuthology
>               Failures:
>                   target158069088033.teuthology:
>                       Data failed to compile:
>                   ----------
>                       Rendering SLS 'base:ceph.salt-api.configure' failed: Jinja variable 'salt.utils.templates.AliasedLoader object' has no attribute 'deepsea.user'
>      Started: 10:27:10.636217
>     Duration: 1298.145 ms
>      Changes:


check http://158.69.81.28/ubuntu-2017-12-14_10:19:29-deepsea:basic:health-ok-ses5---basic-openstack/100/teuthology.log

swiftgist:
Resolved conflicts in #889

PR #797, opened 2017-11-20
Add glance, cinder and nova keyrings/pools

swiftgist:


swiftgist:
These states are not part of any orchestration.  For those that may use OpenStack, this will simply serve as an initial configuration or basic example.

jan--f:
@l-mb is that what you had in mind for easier openstack - ceph integration?

jschmid1:
add swiftgist to whitelist

Issue #783, opened 2017-11-10
openATTIC settings configuration: "Object Gateway API connection failed"


callithea:
Error message: "Non system user capabilities, or wrong admin resource path"
I was able to reproduce this behavior twice (fresh DeepSea setup based on master branch, commit ded2b14da1b97218bd6b7efeefbf2da668a06727).
Fixed this manually by modifying the admin user and setting "system" to true with radosgw-admin.
From ceph -s output:
```
health: HEALTH_OK
rgw:     1 daemon active
```

ceph version: ceph version 12.2.1-361-gaea79b8b7a (aea79b8b7a65a8ca5cc1de84c7fe7d99eaa72ec0) luminous (stable)

Anyone else with this issue?

sebastian-philipp:
https://tracker.openattic.org/browse/OP-2916

PR #804, opened 2017-11-22
radosgw: fix creation of radosgw system user

rjfd:
Fixes a regression introduced in commit 0b5116e of PR #765 

In the commit referenced above, the verification of the 'system' flag was removed and thus currently DeepSea does not create the admin user with the `--system` option.

@smithfarm could you check why our radosgw suite did not catch this regression?

**Credits for finding this bug goes to @tspmelo (Tiago Melo)**

### DIFF
--- a/srv/salt/_modules/rgw.py
+++ b/srv/salt/_modules/rgw.py
@@ -102,6 +102,8 @@ def add_users(pathname="/srv/salt/ceph/rgw/cache", jinja="/srv/salt/ceph/rgw/fil
                 args = ''
                 if 'email' in user:
                     args += " --email={}".format(user['email'])
+                if 'system' in user and user['system']:
+                    args += " --system"
                 if 'access_key' in user:
                     args += " --access-key={}".format(user['access_key'])
 


callithea:
I guess this will fix the issue I've created a while ago: https://github.com/SUSE/DeepSea/issues/783 ?

rjfd:
@callithea good catch! I'll update the commit message to reference the issue number as well

theanalyst:
I don't think rgw deepsea suites have any tests on system users as such

rjfd:
@swiftgist I don't think this needs to be backported because commit 0b5116e is not present in SES5 branch, right?

swiftgist:
On Wednesday, November 22, 2017 04:46:42 PM Ricardo Dias wrote:
> @swiftgist I don't think this needs to be backported because commit 0b5116e
> is not present in SES5 branch, right?

hmm... should only be because of the lint change right?  well... probably 
worth the check :)

rjfd:
@swiftgist yes, this was caused by the lint changes.

PR #811, opened 2017-11-23
stage/remove: fail hard if osd drain fails to prevent data loss

jan--f:


swiftgist:
The Travis failures are addressed in #790.

PR #805, opened 2017-11-22
Radosgw exporter "radosgw-admin usage show --show-log-sum=false" hangs forever with lots of objects (bsc1067453)

votdev:
Add the ability to customize the Prometheus RGW exporter.

For bsc1067453 (DeepSea issue #778)

There are two solutions to workaround the current situation. To do so add the following to
   /srv/pillar/ceph/stack/global.yml:

1. Disable the Prometheus RGW exporter.
```monitoring_prometheus_exporters_ceph_rgw_exporter_cron: default-disable```

2. Disable the collection of the usage metrics:
```monitoring_prometheus_exporters_ceph_rgw_exporter_cron: default-disable-usage-metrics```

After that re-run DeepSea stage 4.
```
# salt '*' saltutil.pillar_refresh
# salt-run state.orch ceph.stage.services
```

PR #812, opened 2017-11-23
spec/fix: add new Prometheus RGW exporter files

carltonf:
This commit adds needed SPEC changes for SR#805:
> Refactor Prometheus RGW exporter customization according to the DeepSea conventions.

### DIFF
--- a/deepsea.spec
+++ b/deepsea.spec
@@ -165,6 +165,9 @@ systemctl try-restart salt-api > /dev/null 2>&1 || :
 %dir /srv/salt/ceph/monitoring/grafana/files
 %dir /srv/salt/ceph/monitoring/prometheus
 %dir /srv/salt/ceph/monitoring/prometheus/exporters
+%dir /srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter
+%dir /srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter/cron
+%dir /srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter/files
 %dir /srv/salt/ceph/monitoring/prometheus/exporters/files
 %dir /srv/salt/ceph/monitoring/prometheus/files
 %dir /srv/salt/ceph/noout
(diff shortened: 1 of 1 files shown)


votdev:
Thx for adding the missing changes in the spec file.

rjfd:
I have one question. PR #805 removed `/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter.sls` and added a new directory `/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter`, which is being added to the spec file in this PR.

When updating deepsea, the old `/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter.sls` file will remain and Salt will use it in favor of the new directory. Is there a way to specify in the spec file that the `/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter.sls` file must be deleted during an update?

@smithfarm @votdev @carltonf

smithfarm:
@rjfd Yes, with something like this:

```
%postun
if [ $1 -ge 1 ] ; then
  rm -rf /srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter.sls
fi
```

@carltonf Could you add that to this PR?

votdev:
What about the Makefile?

smithfarm:
Actually, if `/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter.sls` is no longer used at all, maybe we should delete it in `%postun` unconditionally? I.e. so it is deleted upon update *and* remove.

rjfd:
> What about the Makefile?

The makefile is only used for development. We don't need to support upgrades :)

rjfd:
@smithfarm 
> Actually, if /srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter.sls is no longer used at all, maybe we should delete it in %postun unconditionally?

Yes

carltonf:
> the /srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter.sls file must be deleted during an update?

It's not necessary to manually remove old files, the RPM update process will do it automatically.

The update process is:
> 1. Installs the desired package.
> 2. Erases all older versions of the package, if any exist.

The erasure process will : 
> It reviews the RPM database to find every file listed as being part of the package, and if they do not belong to another package, deletes them.

/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter.sls will not be listed as part of the new Deepsea package so it will be deleted automatically.

See http://ftp.rpm.org/max-rpm/ch-rpm-upgrade.html

rjfd:
@carltonf great! one less thing to worry

Issue #806, opened 2017-11-22
rgw suite: add radosgw system user validation


rjfd:
The RadosGW suite needs to validate that the "admin" user created by DeepSea is created as a `system` rgw user.

This means that in the json output of `radosgw-admin user info --uid=admin` a `system` key must be present with the value `true`.

rjfd:
This issue is related with the regression referred in PR #804

PR #813, opened 2017-11-23
qa-rgw: added rgw system user validation

rjfd:
Fixes: #806

### DIFF
--- a/qa/common/rgw.sh
+++ b/qa/common/rgw.sh
@@ -26,6 +26,22 @@ EOF
   _run_test_script_on_node $TESTSCRIPT $RGWNODE
 }
 
+function rgw_validate_system_user {
+  #
+  # prove the system user "admin" was really set up
+  #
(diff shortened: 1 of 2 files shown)


qa/common/rgw.sh
```diff
@@ -26,6 +26,21 @@ EOF
...
+function rgw_validate_system_user {
+  #
+  # prove the system user "admin" was really set up
+  #
+  local TESTSCRIPT=/tmp/rgw_validate_system_user.sh
+  local RGWNODE=$(_first_x_node rgw)
+  cat << EOF > $TESTSCRIPT
+set -ex
+trap 'echo "Result: NOT_OK"' ERR
+radosgw-admin user info --uid=admin | grep system | grep -q true
```
smithfarm: I guess this won't produce any output for the log? Could we run it once without the greps, and then with?

rjfd: Sure

rjfd: done!

PR #820, opened 2017-11-29
[backport] qa: show deepsea RPM version in case deepsea was installed from RPM

jschmid1:
(cherry picked from commit 289c996adff9c75672a890679c26c8dd188ee875)


rjfd:
@jschmid1 PR #722 is already part of SES5 branch and therefore it does not need to be backported.

jschmid1:
ah, true. That's the case for a couple of the PRs

PR #837, opened 2017-11-30
Correct variable name

swiftgist:


jschmid1:
introduced here: https://github.com/SUSE/DeepSea/pull/823

PR #841, opened 2017-12-04
Staged restart behavior of stage.3

jschmid1:
While we have staged restart in stage.0, I think we have room for  complications in stage.3 as we call
`ceph.{osd,mon,mgr}.restart` on _all_ `roles:{osd,mon,mgr}` at the same time. We do have the iteration around the individual services in the respective state but effectively we are restarting all nodes with a role at the same time.
We also have the check for a failed service but when all services are restarted at the same time, the check becomes quite useless.
I *think* we have to either import the runner like we do in stage.0 or add 3 iterations for every host... Including seems to be the less redudant choice. 

Adding the `- include:` block to the bottom of stage.3/default.sls results in being executed _first_. Working around that by including stage.3 itself and the individual restart orchestrations in a top-level file like we do in stage.0.

I also see an issue on bigger deployments with our cross checking. In stage.3 calls `cephprocessses.wait`. Every restart orchestration calls `cephprocesses.wait`.
And every restart _state_ calls `cephprocesses.wait` (only on the node level). Being really fussy is a good thing imho but it will slow down the stages by quite some time in any deployment > 100 OSDs.
We might want to offer a a alterating _init that only does a minimum ammount of cross checking.

Candidate for backporting; belongs to staged restart feature #808 


jschmid1:
linter passes locally.. :/

srv/salt/ceph/stage/deploy/core/init.sls
```diff
@@ -0,0 +1,4 @@
+
+include:
+  - .{{ salt['pillar.get']('stage_deploy_core_init', 'default') }}
```
swiftgist: stage_deploy_core_init -> stage_deploy_core

swiftgist:
Two things that I ran across: 
  The ceph.wait for each restart needs a failhard. 
  The cephprocesses.wait should target the role.

jschmid1:
> Two things that I ran across:
> The ceph.wait for each restart needs a failhard.

done

> The cephprocesses.wait should target the role.

That was part of my initial concern; cross-checks after every restart are maybe overly cautious but improve the safety of all staged operations, such as restarting and updating. There is certainly a trade-off between performance and stability..

We have checks in:

The restart orchestration:
  - target: all nodes
  - runs_on: each node iteration

The restart state:
  - target: the operating host (all roles on it)
  - runs_on: each invokation (or any iteration in the case of osds & rgws)

If we cut out the check in the restart orchestration we are only ever checking for up and running services on the node itself, which reduces the time spent dramastically. The second option would be to provide an alternative state that implements the safer variation, and get rid of it in the default.

srv/salt/ceph/stage/deploy/core/init.sls
jschmid1: done

jschmid1:
retest this please

jschmid1:
test this please

jschmid1:
test this please

jschmid1:
please test this

jschmid1:
please test this

swiftgist:
Looks good.

smithfarm:
All integration tests except "health-ok" and "health-igw" are now failing reproducibly. The failure happens when the test script tries to run Stage 3. It gets to cephprocesses.wait and fails like this:

`'salt_|-check if mon processes are still running on target137074025153.teuthology after restarting mons_|-check if mon processes are still running on target137074025153.teuthology after restarting mons_|-state': {'comment': 'Run failed on minions: target137074025153.teuthology\nFailures:\n    target137074025153.teuthology:\n    ----------\n              ID: wait processes\n        Function: module.run\n            Name: cephprocesses.wait\n          Result: False\n         Comment: Module function cephprocesses.wait executed\n         Started: 23:34:53.080566\n        Duration: 136867.361 ms\n         Changes:   \n                  ----------\n                  ret:\n                      False\n    \n    Summary for target137074025153.teuthology\n    ------------\n    Succeeded: 0 (changed=1)\n    Failed:    1\n    ------------\n    Total states run:     1\n    Total run time: 136.867 s\n', 'name': 'check if mon processes are still running on target137074025153.teuthology after restarting mons', 'start_time': '23:34:52.707820', 'result': False, 'duration': 137300.998, '__run_num__': 18, '__jid__': '20171212233452742212', 'changes': {}, '__id__': 'check if mon processes are still running on target137074025153.teuthology after restarting mons'}`

Does that mean `cephprocesses.wait` is timing out?

jschmid1:
@smithfarm yes, that means that at least one monitor is not coming up after restart it. Would it be possible to get access to one of machines?

smithfarm:
@jschmid1 Yes, possible, by making an instrumented version of the test script that goes into a deep sleep after Stage 3 fails, instead of bailing out.

Another option would be to run e.g. `health-rgw.sh` in vagrant - have you tried that?

jschmid1:
@smithfarm #868 should fix this issue

smithfarm:
@jschmid1 Thanks; I'll give it a whirl.

PR #847, opened 2017-12-06
Refactor Prometheus RGW exporter

votdev:
Collect and display bucket statistics, e.g. actual size, utilized size and number of objects in Grafana.

This PR uses `radosgw-admin bucket stats` to collect bucket statistics like:
- Number of buckets
- Actual bucket size
- Utilized (compressed) bucket size
- Number of objects in bucket

The following widgets have been added in the 'Ceph - Object Gateway' dashboard:
- Top 10 - Buckets per user
- Top 10 - Bucket size
- Top 10 - Objects in bucket

screenshot02

The following widgets have been added in the 'Ceph - Object Gateway users' dashboard:
- Number of buckets
- Objects in bucket (Overall, per bucket)
- Bucket size (Overall, per bucket)

screenshot01

Additionally the very expensive metrics collected via 'radosgw-admin usage show' have been completely removed.

The following widgets have been removed in the 'Ceph - Object Gateway' dashboard:
- Top 10 - Operations
- Top 10 - Bandwidth (Sent)
- Top 10 - Bandwidth (Received)

The new dashboard:
screenshot_new_02

The following widgets have been removed in the 'Ceph - Object Gateway users' dashboard:
- Operations (Overall, per bucket, per bucket + category)
- Bandwidth (Overall, per bucket, per bucket + category)

The new dashboard:
auswahl_001

### DIFF
--- a/srv/salt/ceph/monitoring/grafana/files/ceph-rgw-users.json
+++ b/srv/salt/ceph/monitoring/grafana/files/ceph-rgw-users.json
@@ -19,6 +19,12 @@
         "id": "prometheus",
         "name": "Prometheus",
         "version": "1.0.0"
+      },
+      {
+        "type": "panel",
+        "id": "singlestat",
+        "name": "Singlestat",
+        "version": ""
       }
     ],
     "annotations": {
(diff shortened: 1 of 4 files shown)


votdev:
What about getting rid of the `radosgw usage show` (see [here](https://github.com/votdev/DeepSea/blob/f3be6f146081b93ea16034f576aee053392e0474/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter/files/ceph_rgw.py#L40))? This is a really expensive operation and will make troubles on greater installations (see [bsc1067453](https://bugzilla.suse.com/show_bug.cgi?id=1067453)). In my opinion the collected data is not that much interesting for admins, or i'm wrong? With #805 this can be disabled, butin this case the Grafana dashboard still displays the widgets without any data. So my suggestion is to remove this completely from the Prometheus exporter and Grafana. Any comments on that?

LenzGr:
Nice work, I like these new "Top 10" stats. What metrics in particular would we lose by disabling the calls to `radosgw usage show`?

votdev:
We will loose the following widgets on the dashboards

Ceph - Object Gateway
- Top 10 - Operations
- Top 10 - Bandwidth (Sent)
- Top 10 - Bandwidth (Received)

Ceph - Object Gateway users
- Operations (Overall, per bucket, per bucket + category)
- Bandwidth (Overall, per bucket, per bucket + category)

Categories are: create_bucket, delete_obj, get_obj, list_bucket, list_buckets, put_obj, stat_bucket

The bandwidth widgets do not display any data if the bucket is shared via NFS. Another point to get rid of it in my opinion.

jschmid1:
@votdev If you want this to be a backport, please add the 'backport' label rather than adding it to the title. The actual backport against the SES5 branch will have that title.

votdev:
@jschmid1 I do not have the privileges to do so. I also can not add reviewers, too.

jschmid1:
@votdev I think you need to be added to SUSE organization. I'll send you the instructions via irc

LenzGr:
I think these new metrics are more valuable and give a better insight. So LGTM.

votdev:
To backport this PR it is necessary that #805 is backported, too.

jschmid1:
add to whitelist

PR #856, opened 2017-12-07
stage 3: don't update ceph packages, only install

jan--f:


jan--f:
not sure if we should add ```refresh: True``` but probably?

jschmid1:
+1 to `refresh`.


Adding a `failhard`  would be nice aswell.

PR #868, opened 2017-12-15
allow indipendent service checks

jschmid1:
Fixes regression:

With https://github.com/SUSE/DeepSea/pull/841 we added restart of individual 'core' services right after the deployment finished. We tried to reduce the noise and crosschecks by targeting only the respective roles.
i.e. only target roles:mon  when you check for running services after restarting mons, instead of doing the full roundtrip and check services on all cluster nodes.

One problem in an initial deployment still persists. When you happen to have a node which holds two roles, one being a core (osd,mon,mgr) and the other being a non-core (i.e. rgw) cephprocesses.check performs on the node and checks for _all_ defined roles for that node. As you haven't executed stage.4 yet the rgw role is _defined_ for the node but not yet _deployed_ which is treated as a failure, hence the stage is aborted.
This doesn't lead to a  complete failure because the restart mechanism is executed _after_ the cluster is deplolyed, but isn't nice nevertheless. The easy-to-fix workaround is to just run stage.4 right after.

With this PR I added a 'ceph.processes.$role' which checks _only_ for the role specified.

'ceph.processes' keeps it's current behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -594,6 +594,24 @@ copy-files:
 	# state files - check processes
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/processes
 	install -m 644 srv/salt/ceph/processes/*.sls $(DESTDIR)/srv/salt/ceph/processes/
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/processes/admin
+	install -m 644 srv/salt/ceph/processes/admin/*.sls $(DESTDIR)/srv/salt/ceph/processes/admin
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/processes/rgw
+	install -m 644 srv/salt/ceph/processes/rgw/*.sls $(DESTDIR)/srv/salt/ceph/processes/rgw
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/processes/osd
+	install -m 644 srv/salt/ceph/processes/osd/*.sls $(DESTDIR)/srv/salt/ceph/processes/osd
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/processes/openattic
+	install -m 644 srv/salt/ceph/processes/openattic/*.sls $(DESTDIR)/srv/salt/ceph/processes/openattic
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/processes/mon
+	install -m 644 srv/salt/ceph/processes/mon/*.sls $(DESTDIR)/srv/salt/ceph/processes/mon
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/processes/mgr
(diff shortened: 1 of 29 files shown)


jschmid1:
please run rgw suite

smithfarm:
Test run details: http://213.32.78.239:8081/ubuntu-2017-12-18_11:21:05-deepsea:basic-ses5---basic-openstack/

Note that the openattic suite is failing currently because there aren't enough OSDs. The error is:

```
2017-12-18T11:46:20.101 INFO:teuthology.orchestra.run.target149202190190.stdout:    cephfs: Command "ceph fs new cephfs cephfs_metadata cephfs_data" run
2017-12-18T11:46:20.101 INFO:teuthology.orchestra.run.target149202190190.stdout:        stdout:
2017-12-18T11:46:20.102 INFO:teuthology.orchestra.run.target149202190190.stdout:        stderr: Error ENOENT: pool 'cephfs_metadata' does not exist
2017-12-18T11:46:20.102 INFO:teuthology.orchestra.run.target149202190190.stdout:    cephfs metadata: Command "ceph osd pool create cephfs_metadata 128" run
2017-12-18T11:46:20.102 INFO:teuthology.orchestra.run.target149202190190.stdout:        stdout:
2017-12-18T11:46:20.102 INFO:teuthology.orchestra.run.target149202190190.stdout:        stderr: Error ERANGE:  pg_num 128 size 3 would mean 1248 total pgs, which exceeds max 1200 (mon_max_pg_per_osd 200 * num_in_osds 6)
```

As a quick fix, I opened a PR [1] increasing the number of disks in the environment teuthology creates for this test.

There is also a failure in `deepsea:basic/nfs-ganesha/{cluster/2nodes3disks.yaml distros/sle_12.3.yaml exec/both-cli.yaml tasks.yaml}`, which is reproducible but, I believe, unrelated. I will need to bisect to see which commit introduced it (or maybe it's an issue in nfs-ganesha itself).

The dead job is just infrastructure noise.

jschmid1:
I triggered a openattic suite myself [0] which was successful. Is there possibly something wrong with the configuration?

[0] http://158.69.81.28/ubuntu-2017-12-18_10:42:47-deepsea:basic:openattic-ses5---basic-openstack/124/teuthology.log
http://158.69.81.28/ubuntu-2017-12-18_10:42:47-deepsea:basic:openattic-ses5---basic-openstack/123/teuthology.log

smithfarm:
No doubt there is something wrong with the configuration, but I don't immediately see why Stage 4 is creating 424 PGs in your case, and 1248 PGs in mine.

callithea:
I've seen this error a few days ago and filed an issue in our upstream tracker: https://tracker.openattic.org/browse/OP-3008.
@smithfarm did the tests pass after adding more disks?

kmroz:
Tested by adding a new rgw role for a node#3 of an existing cluster, and running through stages 2-4.  Works well for me 👍

jschmid1:
rebased

PR #875, opened 2017-12-19
pass role arg as kwargs to match process the requested role

jschmid1:
the signature of cephprocesses.check is 

``` python
def check(results=False, quiet=False, **kwargs):
```

You can call 

``` shell
salt-call cephprocesses.check roles=['mon']
```

because salt converts this as kwargs anyways. If you do that in a module context like before the commit, the params will be just silently ignored and _all_ roles for the requested _search_ pattern are checked.

instead we have to do:

``` python
kwarg={'roles': [role]}
```

note that `kwarg` is the right syntax although the signature actually asks for kwargs ->  see the [docs](https://docs.saltstack.com/en/latest/ref/clients/#localclient)

### DIFF
--- a/srv/modules/runners/cephprocesses.py
+++ b/srv/modules/runners/cephprocesses.py
@@ -36,7 +36,7 @@ def help_():
 
 
 # pylint: disable=dangerous-default-value
-def check(cluster='ceph', roles=[], tolerate_down=0, verbose=True):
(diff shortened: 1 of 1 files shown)


jschmid1:
please run basic suite

jschmid1:
please run rgw suite

PR #882, opened 2017-12-21
Use deepsea.spec in order to make the buildservice happy

jschmid1:
Temp revert of #877 / #872 respectively.

Has to be properly fixed in master and backported. 


jschmid1:
Going the proper way

PR #892, opened 2018-01-10
Add restart mechanism to stage.4 related services

### LINKED ISSUE #891: Add restart mechanism to stage.4 related services
We added `ceph.restart.$role.(lax)` to stage.3 with https://github.com/SUSE/DeepSea/pull/841 and https://github.com/SUSE/DeepSea/pull/808

Although we have all related `ceph.restart.$role` orchestrations we still miss them in the actual stage.4 invocation.

jschmid1:
Special case for iscsi/igw/lrbd:

1) restart vs reload

`systemctl restart lrbd` effectively is a:
``` bash
systemctl stop lrbd
systemctl start lrbd
```
which causes all LUNs to be dropped and re-initialized every time you run stage.4

`systemctl reload lrbd` on the other hand just triggers a invocation of `lrbd` which is non-disruptive but also does not apply all bigger changes. This can only be archived via a `restart`.

There is now a `ceph.igw.restart.force` which executes a `systemctl restart` and a `ceph.igw.restart.reload` which executes a `systemctl reload`.
Calling `ceph.igw.restart` defaults to `reload` now.

2) Detecting readyness

`cephprocesses.check` performs a check on running services. Since lrbd is a onshot python script that runs for the duration of it's upstart we don't require it to be 'up'. However it's necessary to detect if lrbd _is_ still running to determine whether we can move forward with the restart procedure without violating our non-disruptive policy. There is now a _inverted_ list of processes that require to be 'down' in order to make the `.check` function return `True`.

3) Including restarts in stage.4

Stage.4 now performs restart after it's corresponding service was deployed.

 ```yaml

include:
  - ..iscsi
  - ...restart.igw.lax
  - ..cephfs
  - ...restart.mds.lax
  - ..radosgw
  - ...restart.rgw.lax
  - ..ganesha
  - ...restart.ganesha.lax
  - ..openattic
  - ...restart.openattic.lax
```

^^^^ 
now moved to sub-orchestrations

resolves #891 

### CODE REVIEW

File: srv/salt/_modules/cephprocesses.py [hunk 1]
```diff
@@ -27,7 +27,7 @@
              'mgr': ['ceph-mgr'],
              'storage': ['ceph-osd'],
              'mds': ['ceph-mds'],
-             'igw': [],
```
swiftgist: Is this needed?  lrbd will never get checked below and gives the impression that the service needs to be up.

srv/salt/_modules/cephprocesses.py
```diff
@@ -38,8 +38,14 @@
              'client-radosgw': [],
              'master': []}
 
+# Processes like lrbd have an inverted logic
+# if they are running it means that the service is _NOT_ ready
+# as opposed to the the services in the 'processes' map.
+processes_inverted = {'igw': ['lrbd']}
```
swiftgist: I'm just bikeshedding here, but I think absent_processes might match the intention more.

srv/salt/_modules/cephprocesses.py
jschmid1: You get the res['up'/'down'] map based on the services from `processes`.

srv/salt/_modules/cephprocesses.py
jschmid1: Ah, yes. That might make it more clear. Changing

Martin-Weiss:
Does this also influence the way how openATTIC is setting new configs?

As "lrbd reload" does not handle all the adjustments in lrbd.conf (renames, deletions, meta-data changes / parameter changes on existing LUNs, targerts etc.) how should these changes be handled after stage.4 does not restart completely?

jschmid1:
@Martin-Weiss 

> Does this also influence the way how openATTIC is setting new configs?
> 

This PR only targets the general technique we are using to control restarts/reloads. Reacting on a config change is a subsequent task which will be tackled separately.

> As "lrbd reload" does not handle all the adjustments in lrbd.conf (renames, deletions, meta-data changes / parameter changes on existing LUNs, targerts etc.) how should these changes be handled after stage.4 does not restart completely?

Detecting whether a restart or a reload is required is tricky. Ideally this would be archived in lrbd itself. However, this is not possible right now. The current strategy is to rely on the admins proficiency to compare the desired results and perform a staged restart if the regular reload just wasn't enough.

`salt-run state.orch ceph.restart.igw.force` would be your friend then.

jschmid1:
please run basic suite

jschmid1:
please run rgw suite

swiftgist:
Take a look at
https://github.com/SUSE/DeepSea/commit/bb006642db1a16ff588df4085f8f950a61f4b76a

I moved the individual restarts to their respective services.

jschmid1:
> I moved the individual restarts to their respective services.

that's even better. I'll cherry pick that atop this PR

PR #895, opened 2018-01-12
Handle missing wal/db

swiftgist:

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -1259,6 +1259,7 @@ class OSDCommands(object):
         log.info("activate: {}".format(cmd))
         return cmd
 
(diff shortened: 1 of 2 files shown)


jan--f:
Is there any bug related to this? Did this misbehave?

jschmid1:
I think this method is quite hard to comprehend without any context. I'd like to see a test case that describes the context and makes sure to not run into this issue anymore.

swiftgist:
@jan--f Yes, this came from Martin.  Admin configures stand-alone OSDs.  Then, admin reconsiders and decides to separate wal/db.  The osd.report will return nothing wrong and migrate.osd will skip these OSDs as well.

@jschmid1 I created a handful of unit tests for is_incorrect.  The two related to this bug are the bluestore ones that end with _no_device.

PR #901, opened 2018-01-19
add exception for mailformed policy.cfg

### LINKED ISSUE #897: Add debug statements to push.py
Tracing a bug that was caused by a missing comment in the policy.cfg was not easy, stracktraces were very missleading.

Make debugging easier by adding log.debug statements to:

https://github.com/SUSE/DeepSea/blob/master/srv/modules/runners/push.py#L247

denisok:
closes #897


jschmid1:
@denisok added you to the list of admins

PR #903, opened 2018-01-24
Block device benchmark module

ddiss:
@jan--f you probably have something to say about the srv/salt/ceph/benchmarks/blockdev/ path usage ;-) . I still think it makes sense to nest all of the benchmark backend files under the parent _benchmarks_ path, would you be open to using separate subdirectories per module here?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -385,6 +385,9 @@ copy-files:
 	install -m 644 srv/salt/ceph/rbd/benchmarks/*.sls $(DESTDIR)/srv/salt/ceph/rbd/benchmarks/
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/rbd/benchmarks/files
 	install -m 644 srv/salt/ceph/rbd/benchmarks/files/keyring.j2 $(DESTDIR)/srv/salt/ceph/rbd/benchmarks/files/
+	# state files - benchmark-blockdev
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/benchmarks/blockdev/
+	install -m 644 srv/salt/ceph/benchmarks/blockdev/*.sls $(DESTDIR)/srv/salt/ceph/benchmarks/blockdev/
 	# state files - reactor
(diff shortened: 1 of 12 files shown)


jan--f:
The motivation for keeping orchestration files under ```benchmarks``` and state files related to benchmarks under ```<backend>/benchmarks``` (say ```cephfs/benchmarks```) was mostly to not mix orchestration and state files in a sub directory tree. I'm certainly not married to this idea, I'm merely interested in consistency ;)
What do others think... @swiftgist @jschmid1 @mogeb ?

jschmid1:
> The motivation for keeping orchestration files under benchmarks and state files related to benchmarks under <backend>/benchmarks (say cephfs/benchmarks) was mostly to not mix orchestration and state files in a sub directory tree. I'm certainly not married to this idea, I'm merely interested in consistency ;)
> What do others think... @swiftgist @jschmid1 @mogeb ?

Right, that's what we did with most of our states/orchestration hierarchy. I don't see a big problem here, though. We have `ceph.benchmarks.blockdev` which is an orchestration and should live under the benchmark namespace. The called modules/states like `ceph.tool.fio` etc are included

ddiss:
@jan--f @jschmid1 Thanks for the feedback.
So aside from fixing the spec file, I'll work on a separate PR that moves the remaining benchmark backend orchestration files under srv/salt/ceph/benchmarks/<backend>/ , assuming there are no objections to the relocation.

ddiss:
These latest changes are rebased against master and include an extra spec file patch.

ddiss:
@jschmid1 thanks for the review! I encountered one issue on redeployment, which needs the following change:
```
--- a/srv/salt/_modules/cephprocesses.py
+++ b/srv/salt/_modules/cephprocesses.py
@@ -36,6 +36,7 @@ processes = {'mon': ['ceph-mon'],
              'client-iscsi': [],
              'client-nfs': [],
              'client-radosgw': [],
+             'benchmark-blockdev': [],
              'master': []}
```
I'll squash it in with this series.

PR #912, opened 2018-01-26
ganesha: install: Refrain from installing recommended packages

smithfarm:
By default, this pkg.installed block will do:

zypper --non-interactive in nfs-ganesha nfs-ganesha-ceph nfs-ganesha-rgw nfs-ganesha-utils

which brings in 170 packages (including entire GUI stack). Adding `--no-recommends` would fix this, but we're not running zypper directly here.

The Salt docs say that "install_recommends" is supported only with APT, but maybe it
has been implemented for zypper in the meantime. Anyway, it's worth a try!

### DIFF
--- a/srv/salt/ceph/ganesha/install/default.sls
+++ b/srv/salt/ceph/ganesha/install/default.sls
@@ -20,4 +20,5 @@ install_ganesha:
 {% if grains.get('os_family', '') == 'Suse' %}
         - nfs-ganesha-utils
 {% endif %}
+    - install_recommends: False
     - fire_event: True


supriti:
@smithfarm Were you able to test the patch already and see if install_recommends is working for zypper+salt?

smithfarm:
@supriti No, not yet. I'll report back (and remove "needs-qa") when I get around to testing.

jschmid1:
@smithfarm I briefly checked the code and could not find the implementation for zypper.. there is one for apt as described in the docs.

smithfarm:
I assume this patch does not work. It's unfortunate that deploying nfs-ganesha causes so many unnecessary packages to be installed. @jschmid1 @swiftgist Would you be open to calling zypper directly in this case?

kshtsk:
WARNING. This PR is not tested with teuthology, please run the tests again if required.

PR #916, opened 2018-01-30
fix updated node exporter

jan--f:
I have not yet tested the this for lack of a suitable test environment but will do so. If anyone wants to test drive this...feel free.


tserong:
I'm not sure about the smartmon collector changes (how do I test those?) but the update to `--` style ARGS in /etc/sysconfig/prometheus-node_exporter works fine when tested on openSUSE 42.3 with golang-github-prometheus-node_exporter 0.15.2.

One small kink - the commit messages in this PR seem to be for the wrong commits, i.e. 
93f2874 fixes the args, and ca495d9 updates the smartmon collector :-)

jan--f:
Now how did that happen? Fix coming up.

jan--f:
The smartmon collector change only adds a few fields to the device_info metric...so more metadata per disk. The other changes are only internal...improving a few regular expressions. In essence this is just an update to what the node_exporter currently ships.

Issue #910, opened 2018-01-26
rgw.py: Don't call s3connect for manually created users. 


supriti:
Manually created users don't have keys in json file at /srv/salt/ceph/rgw/cache. We are calling s3connect with keys passed as parameters. This result in s3connect throwing error for manually created user: " 'Check your credentials' % (len(names), str(names))) NoAuthHandlerFound: No handler was ready to authenticate. 1 handlers were checked. ['HmacAuthV1Handler'] Check your credentials".

PR #919, opened 2018-01-30
Check if valid access key/secret key before calling s3connect

supriti:

### DIFF
--- a/srv/salt/_modules/rgw.py
+++ b/srv/salt/_modules/rgw.py
@@ -262,6 +262,8 @@ def s3connect(user):
     """
     Return an S3 connection
     """
+    if access_key(user) is None or secret_key(user) is None:
+        return
     endpoint = endpoints()[0]
 
     s3conn = boto.connect_s3(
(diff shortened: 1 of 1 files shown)


supriti:
https://github.com/SUSE/DeepSea/issues/910

srv/salt/_modules/rgw.py
```diff
@@ -262,6 +262,8 @@ def s3connect(user):
     """
     Return an S3 connection
     """
+    if access_key(user) is None or secret_key(user) is None:
```
theanalyst: There is a specific case where this might be not true, if LDAP is used for RGW authentication; then secret_key might be empty. Does boto module throw the same exception for empty secret keys?

supriti: I tried with empty access_key and secret_key. Surprisingly, boto did not complain in that case. It seems like the issue we saw was because we returned None here (https://github.com/SUSE/DeepSea/blob/master/srv/salt/_modules/rgw.py#L157) If I change the return just to empty string it works. Boto does not complain with empty keys.

supriti: I think this PR will work for LDAP where we have empty keys. In that case, https://github.com/SUSE/DeepSea/blob/master/srv/salt/_modules/rgw.py#L155 will return empty string, and boto s3connect will not complain. create_bucket/lookup will handle the exception. In case there is no user json file, then it make sense to not call s3connect at all, and return.

PR #928, opened 2018-02-07
Check if valid access key/secret key before calling

supriti:
s3connect

### DIFF
--- a/srv/salt/_modules/rgw.py
+++ b/srv/salt/_modules/rgw.py
@@ -230,6 +230,11 @@ def endpoints(cluster='ceph'):
 
 
 def s3connect(user):
+    """
+    Return an S3 connection
+    """
+    if access_key(user) is None or secret_key(user) is None:
+        return
     endpoint = endpoints()[0]
 
     s3conn = boto.connect_s3(
@@ -244,6 +249,8 @@ def s3connect(user):
 
 def create_bucket(**kwargs):
     s3conn = s3connect(kwargs['user'])
+    if s3conn is None:
+        return False
     try:
         s3conn.create_bucket(kwargs['bucket_name'])
     except boto.exception.S3CreateError:
@@ -252,6 +259,8 @@ def create_bucket(**kwargs):
 
 def lookup_bucket(user, bucket):
     s3conn = s3connect(user)
+    if s3conn is None:
+        return False
     if s3conn.lookup(bucket, validate=True) is None:
         return False
 


srv/salt/_modules/rgw.py
```diff
@@ -230,6 +230,11 @@ def endpoints(cluster='ceph'):
 
 
 def s3connect(user):
+    """
+    Return an S3 connection
+    """
+    if access_key(user) is None or secret_key(user) is None:
```
jschmid1: more pythonic:

``` python
if not smth:
  do_smth
```
You can omit the `is None` unless you specifically expect `None`. `if not` looks for `False`, `None` and empty strings.

srv/salt/_modules/rgw.py
```diff
@@ -230,6 +230,11 @@ def endpoints(cluster='ceph'):
...
+        return
     endpoint = endpoints()[0]
 
     s3conn = boto.connect_s3(
```
jschmid1: @theanalyst That's the boto(3) call we have to adapt, right?

srv/salt/_modules/rgw.py
supriti: I am specifically looking for None here. This function (https://github.com/SUSE/DeepSea/blob/master/srv/salt/_modules/rgw.py#L146) can return None and empty strings. Empty strings are ok, but None is not.

srv/salt/_modules/rgw.py
supriti: @jschmid1connect_s3 complaints only when None is passed for access_key/secret_key(https://github.com/SUSE/DeepSea/issues/910). So exit early in that case.

theanalyst: yes, for newer versions we can move this to boto3

srv/salt/_modules/rgw.py
jschmid1: ok, thanks. Haven't looked that far. No objections then.

supriti:
@theanalyst @jschmid1 @swiftgist can this pr be merged?

jschmid1:
please run rgw suite

jschmid1:
please run ganesha suite

jschmid1:
@smithfarm @kshtsk 

both testsuites report success although they timed out.

smithfarm:
@jschmid1 Can you post a link to the log, so I can look at the timeout?

jschmid1:
@smithfarm 
http://storage-ci.suse.de:8080/job/deepsea-teuthology-ganesha-trigger/6/default/artifact/teuth-deepsea-b6.log
http://storage-ci.suse.de:8080/job/deepsea-teuthology-rgw-trigger/12/default/artifact/teuth-deepsea-b12.log

kshtsk:
@jschmid1 
>> both testsuites report success although they timed out.
We need to add a check to the jenkins job, that if it didn't find any jobs, i.e. PASS+FAIL+DEAD=0 it should return false. Or handle an exit code.

kshtsk:
Pulpito at http://158.69.67.200:8081/ is down, this is the reason of this failure.

kshtsk:
please run rgw suite

kshtsk:
please run ganesha suite

Issue #929, opened 2018-02-09
maintenance.upgrade, remove time_init: ntp


swiftgist:
The time_init setting has been hardcoded into the default global.yml.  With the switch to chrony, the entry should be removed and rely on the expectation of the .../ceph/time/default.sls file.

This only impacts existing clusters.  The upgrade can remove the "time_init: ntp" line from /srv/salt/ceph/proposals/config/global.yml.  For those that wish to keep using ntp, add the same entry to /srv/salt/ceph/stack/global.yml.

For those wishing to switch to chrony before this is implemented, add "time_init: chrony" to /srv/salt/ceph/stack/global.yml.

PR #930, opened 2018-02-09
Support chrony, set as default, add rescind

swiftgist:
Same strategy as ntp setup.  DeepSea only handles client configuration.  Available commands are

```
salt '*' state.apply ceph.time
salt '*' state.apply ceph.time.chrony
salt '*' state.apply ceph.time.ntp
```

```
salt '*' state.apply ceph.rescind.time
salt '*' state.apply ceph.rescind.time.chrony
salt '*' state.apply ceph.rescind.time.ntp
```

The default for a fresh install is chrony.  This will work with ntpd or chronyd as a server.  (Be sure to add the 'allow' directive for chronyd to listen on port 123.)

Chrony supports immediately sync'ing, so no separate step needed.  The Salt commands are plain enough that I am hopeful these will work on other distros without exceptions.  The rescind functionality cannot be part of Stage 5 or the purge orchestration.  

Existing installations can follow #929 for now.

### DIFF
--- a/srv/salt/ceph/time/README
+++ b/srv/salt/ceph/time/README
@@ -1,12 +1,24 @@
-Set up NTP to manage system time
+Set up Chrony or ntp to manage system time.  Chrony is preferred.
 
-Ceph relies on loosely synchronized clocks. DeepSea will by default start an
-NTP client on each cluster node and point it to the NTP server running on the
-Salt Master if and only if the ntpd.service systemd unit is not running.
+Ceph relies on synchronized clocks. DeepSea will by default start a chrony 
+client on each minion.  If chrony is not available, then ntp must be selected
+in the pillar.  In either case, the default time server is the Salt master 
+unless overridden.  The Salt master must be configured manually.  (Salt can be
+used for setting up a time server, but admins have strong opinions about the
+correct configuration for time.  The Salt master is suitable for virtual
+environments.)
 
-If you want DeepSea to always overwrite an existing NTP configuration set
-"time_init: ntp" in the pillar.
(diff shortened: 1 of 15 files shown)


srv/salt/ceph/time/README
```diff
@@ -1,12 +1,24 @@
-Set up NTP to manage system time
+Set up Chrony or ntp to manage system time.  Chrony is preferred.
 
-Ceph relies on loosely synchronized clocks. DeepSea will by default start an
-NTP client on each cluster node and point it to the NTP server running on the
-Salt Master if and only if the ntpd.service systemd unit is not running.
+Ceph relies on synchronized clocks. DeepSea will by default start a chrony 
+client on each minion.  If chrony is not available, then ntp must be selected
+in the pillar.  In either case, the default time server is the Salt master 
+unless overridden.  The Salt master must be configured manually.  (Salt can be
```
mogeb: Should we add quick instructions on how to setup a chrony server? Eric suggests either adding something in the wiki, or a prewritten configuration file under `contrib/`.

Martin-Weiss: We should also add the info that the default setup that deepsea creates is creating a single point of failure (admin down = no timesync) and due to that must not be used that way in production.

swiftgist: I am fine with adding sample configuration files for time servers under contrib and putting a README there or adding something to the wiki to reference it.  

The default setup is just that.  It's the default which is suitable for a Proof of Concept.  I expect admins have a time server setup already and that the admin node would be a client as well.

jan--f: @swiftgist's suggestion sounds good to me.

PR #933, opened 2018-02-12
benchmarks/fs: add generic filesystem benchmark

ddiss:
This functionality triggers fio, using a local path as destination
(work_dir), with the assumption that work_dir has already been mounted.

In future, this functionality will be inherited by CephFS and SMB/CIFS
benchmark backends.

Signed-off-by: David Disseldorp <ddiss@suse.de>

PR #945, opened 2018-02-19
benchmarks/fs: add generic filesystem benchmark

ddiss:
changes since v1 (https://github.com/SUSE/DeepSea/pull/933):
- add rescind stubs

This functionality triggers fio, using a local path as destination
(work_dir), with the assumption that work_dir has already been mounted.

In future, this functionality will be inherited by CephFS and SMB/CIFS
benchmark backends.

Signed-off-by: David Disseldorp <ddiss@suse.de>


ddiss:
This was also missing the Makefile changes for the rescind stubs. Fixed in this new version.

jschmid1:
please run basic suite

PR #946, opened 2018-02-19
Create a pullrequest template file

jschmid1:
Create a pullrequest template file


smithfarm:
Nice, but I'm missing information on how to view the test results?

jschmid1:
@smithfarm which is not intuitive at all.. even with a proper explanation this is quite hard. I'll give it a try

jschmid1:
@smithfarm We probably have to pre-parse the teuth-log and extract the links for easier inspection. The current approach is quite tedious :)

smithfarm:
@jschmid1 OK, and the "health-ok check" appears only when someone explicitly asks for this check? (I don't see it in this PR, for example)

jschmid1:
@smithfarm I took this info for granted. I mean, there is an entire paragraph where certain 'phrases' are described.

What I'd also like to avoid is bloating the text too much. That's a template you will be presented with every time you create a PR. If you have to scroll down to see the information you'll probably ignore it..

What do others think? @jan--f @swiftgist @tserong @mogeb

smithfarm:
@jschmid1 What do you mean "take this info for granted"? I am just asking because I don't know this info, not that you necessarily need to add it to the document.

smithfarm:
(I thought all 7 teuthology CI tests were triggered automatically when the PR was opened)

jschmid1:
@smithfarm 

> (I thought all 7 teuthology CI tests were triggered automatically when the PR was opened)

That's not the case. You have to individually trigger the tests.

That's what I tried to describe in the file:

>  If you want to run tests against your PR you can simply trigger these by leaving comments.
>  Depending on your needs, we currently have implemented tests for:



> What do you mean "take this info for granted"? I am just asking because I don't know this info, not that you necessarily need to add it to the document.

I thought this clear enough. We just added yet another check to 'Github's Status Checks' which are known, I think. If that's not the case I can add more information, but I fear that this will go beyond the scope of this document.

.github/pull_request_template.md
```diff
@@ -0,0 +1,49 @@
+Fixes #
+
+
+Please sign your PullRequest
```
tserong: Is it worth putting "Signed-off-by: ..." here instead of "Please sign your PR"?

tserong:
Once this is accepted, I assume when opening a PR, the description box will be pre-filled with this, so then the person opening the PR can see all those details, be informed about the testing (which is helpful), and delete all those bits before _actually_ opening the PR.

What will happen with single-commit PRs, where the PR description would usually be pre-filled with the commit's description? Will the template take precedence, or will both things appear, or something else?

.github/pull_request_template.md
smithfarm: Are we now requiring "Signed-off-by:" in DeepSea? Unless everyone starts signing their commits, this should just be dropped?

jschmid1: good point @tserong 

@smithfarm I think this is best practice and we should at least start remind people of it.

jschmid1:
@tserong 
> What will happen with single-commit PRs, where the PR description would usually be pre-filled with the commit's description? Will the template take precedence, or will both things appear, or something else?

This, I actually don't know.. We might try it out and when it appears to be annoying we can just move the information to the wiki.

tserong:
Works for me :-)

smithfarm:
Shouldn't the core developers be signing their commits, though, if signing PRs is to be declared a best practice that contributors are asked to follow?

jschmid1:
> Shouldn't the core developers be signing their commits, though, if signing is to be declared a best practice that contributors are asked to follow?

sure, they should. I personally enabled GPG-Signing for all my commits and additionally add a 'Signed-off-by' tag when using the commandline(this PR was created from the GUI). I personally think that the 'Signed-off-by' line is not needed anymore when the commit is gpg-signed already. 
Thoughts?

LenzGr:
_I personally think that the 'Signed-off-by' line is not needed anymore when the commit is gpg-signed already._

Actually, I have to disagree - the `Signed-off-by` line serves a different purpose than cryptographically signing your commit. By adding this line, you usally express your agreement to the [Developer Certificate of Origin (DCO)](https://developercertificate.org/), a concept that originates from the Linux Kernel community and has since then been adopted by many OSS projects like Ceph (or [openATTIC](http://docs.openattic.org/en/latest/developer_docs/contributing_guidelines.html#signing-your-patch-contribution), FWIW).

jschmid1:
> Actually, I have to disagree - the Signed-off-by line serves a different purpose than cryptographically signing your commit. By adding this line, you usally express your agreement to the Developer Certificate of Origin (DCO), a concept that originates from the Linux Kernel community and has since then been adopted by many OSS projects like Ceph (or openATTIC, FWIW).

Interesting, thanks @LenzGr for the insight. I'd be also fine with enforcing the 'Signed-off-by:' line and have your commit cryptograhically signed, optional.

Let's elaborate on that in our upcomming meeting.

jschmid1:
I moved all the infromation about how to trigger a test to the wiki and pasted a link into the template.

I figured that you'd actually need the information _after_ you created the PR. So you can either just leave the link there for later inspection or just delete it after opening it in a second tab.

Additionally I added a issue template which is inspired by saltstack/salt's template.

Thoughts?

jan--f:
Sorry to draw out this discussion...

- I think a short text might be good. ```Short description of the issue or feature  is addressed in this PR``` for example.
- Checking for signed commits can be added as a check, can it not? Wouldn't that be better then providing a line in the PR template (that could be left empty)?

jschmid1:
> I think a short text might be good. Short description of the issue or feature is addressed in this PR for example.

maybe similar to the ISSUE_TEMPLATE ?



> Checking for signed commits can be added as a check, can it not? Wouldn't that be better then providing a line in the PR template (that could be left empty)?

Probably. ceph/ceph does that iirc.. Maybe @kshtsk can help with that.

jschmid1:
I postponed the "Signed-off-by" enforcement until we have an automated checker for this.

I also added a 'Description' paragraph in the pullrequest template file.

Any more remarks/thoughts/suggestions from anyone?

Issue #582, opened 2017-08-26
Ceph allows public and cluster networks to be specified as multiple subnets, DeepSea did too - but no longer (REGRESSION)


smithfarm:
In ceph.conf, `public_network` and `cluster_network` might be assumed to be a single subnet like 192.168.0.1/24, but Ceph actually allows multiple subnets as the value. See http://docs.ceph.com/docs/master/rados/configuration/network-config-ref/#id1 where it says: "You may specify comma-delimited subnets." `{ip-address}/{netmask} [, {ip-address}/{netmask}]`

DeepSea was handling the "multiple subnets" case nicely for some time, but some recent changes seem to have broken this. For details, see #580.

Here are two examples of code that breaks when `public_network` is either a string like "149.202.170.163/32, 149.202.170.164/32" or a list like `['149.202.170.163/32, 149.202.170.164/32']`:

https://github.com/SUSE/DeepSea/blob/f259476910ce5e5875a4f448a184e01860dc2d16/srv/modules/runners/populate.py#L933

https://github.com/SUSE/DeepSea/blob/e9ffafc41fe77f38557cafb9b2403e75677ae7f1/srv/salt/_modules/public.py#L40

Presumably there might be more.

swiftgist:
I submitted #583 but I don't know why we are storing the multiple networks as a string in the pillar.  No other data structure does this IIRC.  I expected multiple networks to be an array originally.  For instance, look at roles for each of the minions.

In other words, when did the concatenated string in Salt become a requirement?

smithfarm:
It would certainly make sense if Salt stored the subnets as an array. First, all the code that uses the subnets would be forced to take an array (even if it had just one element). Second, support for multiple subnets would be ensured.

smithfarm:
As far as I can tell, the concatenated string was introduced by https://github.com/SUSE/DeepSea/pull/96 - @rjfd what do you think about changing this to an array?

rjfd:
When I made that PR the only place where the public_network was used was to create the ceph.conf configuration using jinja, and therefore it was easier to use a string.

Now that the public_network is used elsewhere I also think it should be an array. People just need to make sure that update all places where this info is used.

PR #951, opened 2018-02-20
populate: support multiple public, cluster networks in engulf (bsc#1076010)

tserong:
This removes the assumption in the engulfer that there can be no more
than one public or cluster network.  Without this change, running engulf
on a cluster with (e.g.) public_network = 10.27.21.0/24, 10.27.20.0/24
will result in the following error:

  [ERROR   ] Failed to determine cluster's public_network.
  [ERROR   ] Failed to replace derived public_network with public_network
             of existing cluster.

Note that this still leaves the public_network and cluster_network settings
as strings in the pillar data (changing the pillar data to be an array would
have been somewhat more invasive).

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1076010
Fixes: https://github.com/SUSE/DeepSea/issues/582

### DIFF
--- a/srv/modules/runners/populate.py
+++ b/srv/modules/runners/populate.py
@@ -996,11 +996,11 @@ def _replace_key_in_cluster_yml(key, val):
     return True
 
 
-def _get_existing_cluster_network(addrs, public_network=None):
+def _get_existing_cluster_networks(addrs, public_networks=[]):
     """
     Based on the addrs dictionary { minion: [ipaddress] }, this function
-    returns an address consisting of network prefix followed by the cidr
-    prefix (ie. 10.0.0.0/24), or None.
+    returns a list of addresses consisting of network prefix followed by the
+    cidr prefix (e.g. [ "10.0.0.0/24" ]).  It may return an empty list.
     """
     target = deepsea_minions.DeepseaMinions()
     search = target.deepsea_minions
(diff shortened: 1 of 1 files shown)


srv/modules/runners/populate.py
```diff
@@ -1009,30 +1006,31 @@ def _get_existing_cluster_network(addrs, public_network=None):
...
-        # We have multiple possible networks.  This is liable to happen with OSDs
-        # when there is a private cluster network.  Let's try to remove the public
-        # network.
-        minion_networks = [n for n in minion_networks if n != public_network]
-        network = minion_networks[0] if len(set(minion_networks)) == 1 else None
+    # Collapse minion_networks back to unique items.  Usually there'll
+    # be only one network after this, unless the cluster is using a
+    # separate public and private network, or is using multiple public
+    # and/or private networks
+    minion_networks = list(set(minion_networks))
```
jan--f: Any reason you're not using ```minion_networks = set()``` right away?

tserong: Just that when I wrote it I was thinking "minion_networks" needs to always be a list. I now realise that the `[n for n in minion_networks]` business on the next line would work just fine against a set, and would return a list, so `list()` in the above line is redundant.

tserong:
I took the liberty of adding a 'TODO' comment to investigate simplifying `_get_existing_cluster_networks`. I've also gotten rid of that redundant `list()`

tserong:
please run basic suite

tserong:
`2018-02-21 23:00:13,534.534 INFO:teuthology.openstack:2018-02-22 04:00:13,527.527 ERROR:teuthology.lock:Could not contact lock server: http://localhost:8080/`

That can't possibly be a result of this commit, can it?

jan--f:
@tserong that wouldn't be my first guess either. @smithfarm what do you think?

smithfarm:
My guess is teuthology is hosed and will need to be redeployed.

kshtsk:
please run health-ok suite

kshtsk:
please run openattic suite

kshtsk:
please run rgw suite

kshtsk:
please run health-ok suite

kshtsk:
please run openattic suite

tserong:
Anyone up for reviewing this?

tserong:
Rebased

Issue #962, opened 2018-02-23
disengage.safety not long enough in case of not-responding minions


Martin-Weiss:
Just ran into this issue today:

admin-p:~ # salt-run disengage.safety; salt-run remove.osd 45
[WARNING ] Although 'dmidecode' was found in path, the current user cannot execute it. Grains output might not be accurate.
safety is now disabled for cluster ceph
[WARNING ] Although 'dmidecode' was found in path, the current user cannot execute it. Grains output might not be accurate.
[WARNING ] Although 'dmidecode' was found in path, the current user cannot execute it. Grains output might not be accurate.
False
[ERROR   ] ('Safety is not disengaged...refusing to remove OSD', ' run "salt-run disengage.safety" first THIS WILL CAUSE DATA LOSS.')
False

--> root cause seems to be that we have some minions registered to the salt-master on the admin that are not up and running and due to that do not respond..

admin-p:~ # salt '*' test-ping
admin-p.ses.intern.thomas-krenn.com:
    'test-ping' is not available.
osd04-p.ses.intern.thomas-krenn.com:
    'test-ping' is not available.
osd03-p.ses.intern.thomas-krenn.com:
    'test-ping' is not available.
osd01-p.ses.intern.thomas-krenn.com:
    'test-ping' is not available.
osd02-p.ses.intern.thomas-krenn.com:
    'test-ping' is not available.
sles12sp3-test3.ses.intern.thomas-krenn.com:
    Minion did not return. [No response]
sles12sp3-test1.ses.intern.thomas-krenn.com:
    Minion did not return. [No response]
sles12sp3-test4.ses.intern.thomas-krenn.com:
    Minion did not return. [No response]
sles12sp3-test2.ses.intern.thomas-krenn.com:
    Minion did not return. [No response]
ERROR: Minions returned with non-zero exit code

--> it seems that the "disengage.safety" and also the other jobs take too long in case one or the other minion is not reachable..

Is it possible to adjust the time the safety disengagement allows us to do "bad" things? 
Or could the processes somehow be configured to ignore the not available clients during these processes?

jschmid1:
> admin-p:~ # salt '*' test-ping

it's `test.ping`

note the `.` instead of the `-`

> Is it possible to adjust the time the safety disengagement allows us to do "bad" things?

Currently it's hardcoded to 60 seconds, but can be changed if needed.

Martin-Weiss:
[...]

PR #963, opened 2018-02-23
Add disengage_safety_duration parameter.

jschmid1:
Fixes: #962

That parameter allows to set the disengaged safety time from the
default(300) seconds to a variable amount.

Invocation with:
``` bash

salt-run disengage.check timeout=42

salt-run remove.osd timeout=42
(the _timeout_ arg is passed through)

```

For orchestrations you can obviously can't utilize kwarg passing. For those cases there is now a 

*disengage_safety_duration* 

variable which can be set in the pillar.

Note that passing args via the CLI has higher precendence.

### DIFF
--- a/srv/modules/runners/disengage.py
+++ b/srv/modules/runners/disengage.py
@@ -10,6 +10,9 @@ from __future__ import absolute_import
 from __future__ import print_function
 import os
 import time
+import logging
+
+log = logging.getLogger(__name__)
 
 
 class SafetyFile(object):
@@ -29,7 +32,7 @@ def help_():
              '    Touches a file to signify imminent dangerous operations\n'
              '\n\n'
              'salt-run disengage.check:\n\n'
-             '    Check whether the timestamp is less than a minute old\n'
+             '    Check whether the timestamp is less than 300 seconds old\n'
              '\n\n')
     print(usage)
     return ""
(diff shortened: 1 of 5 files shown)


jschmid1:
please run basic suite

swiftgist:
This timeout is not for the software, but for the human.  One side effect is that we are not controlling which admin ran the command and therefore, any destructive command can be run within the timeout.  If one admin sets 'timeout=3600' and another is searching their history and fat fingers the enter key, what was the point of preventing accidents?

To my knowledge, the only complaint we have had is the one from Martin with a broken Salt cluster (i.e. minions downed and not responding)

Martin-Weiss:
> To my knowledge, the only complaint we have had is the one from Martin with a broken Salt cluster (i.e. minions downed and not responding)

I believe in every bigger system we have to expect that not all minions are responding or are reachable.
So if you have a cluster with 100 or more nodes - it is normal that some hosts are broken.. (the bigger the more likely).

Once we manage clients this will be even more likely.

Any idea why a client that is not reachable but not part of the cluster increases the time required for calling disengage.safety?

jschmid1:
> If one admin sets 'timeout=3600' and another is searching their history and fat fingers the enter key, what was the point of preventing accidents?

That's ofc not ideal. Maybe there should be a max_value to which you can set the duration.



> To my knowledge, the only complaint we have had is the one from Martin with a broken Salt cluster (i.e. minions downed and not responding)

It's still the case that when you have a minion that is currently not online but is expected to come back within the next couple of hours (think hardware replacement), the default duration we had before is not sufficient enough to survive the initial up-time-check that is perfomed by salt-run.

If checks for other minions to be up for about a minute. Then it starts to invoke the i.e. `remove.osd` command which internally checks for the disengaged safety file. By that time your safety is already disengaged, which means that you can never execute the command although you are aware of the risk.
Again this is a discussion of "Do you really know what you are doing" and "Should Deepsea patronize you"

swiftgist:
@Martin-Weiss I will have to disagree with your assertion that a working Salt cluster can contain minions that fail to respond.  If a minion is broken and will not be corrected, then removing the minion from Salt or removing the minion from deepsea_minions would be the correct action.  Either that minion is important or it's not.

@jschmid1 If a minion is down for hours, I don't think 

```
# salt-key -d minion.domain
```
followed by 
```
# salt-key -A 
```
when the minion is booted is a huge burden.  It's not like whatever we do specifically for `remove.osd` is going to fix all the Stages and other non-DeepSea Salt related tasks which will timeout on the one minion as well.  Give them a ten minute window and Stage 0/3 will still take forever with a broken minion.

kshtsk:
retest this please

jschmid1:
obsolete with #1216

PR #970, opened 2018-02-27
Include /dev/disk/by-id/scsi*

### LINKED ISSUE #969: Include scsi device names
The find command in _uuid_device only searches for ata* and nvme*, but some servers have neither.  As a result, we fallback to the short name (e.g. /dev/sdb).  This affects the grains that get saved for removing an OSD later.

swiftgist:
fix #969 


jschmid1:
I assume we need this to be backported aswell. Adding labels accordingly

PR #982, opened 2018-03-05
validate: ceph_version improvements

smithfarm:
There are some issues with `ceph_version`, will try to address them in this PR.

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -752,12 +752,12 @@ class Validate(object):
         target = deepsea_minions.DeepseaMinions()
         search = target.deepsea_minions
         local = salt.client.LocalClient()
-        contents = local.cmd(search, 'pkg.latest_version', ['ceph'], tgt_type="compound")
+        contents = local.cmd(search, 'pkg.latest_version', ['ceph-common'], tgt_type="compound")
         for minion, version in contents.items():
             if not version:
-                info = local.cmd(minion, 'pkg.info_installed', ['ceph'])
-                if info and 'version' in info[minion]['ceph']:
-                    version = info[minion]['ceph']['version']
+                info = local.cmd(minion, 'pkg.info_installed', ['ceph-common'])
+                if info and 'version' in info[minion]['ceph-common']:
+                    version = info[minion]['ceph-common']['version']
                 else:
                     self.errors.setdefault('ceph_version', []).append(
                         "No Ceph version is available for installation in {}".format(minion))
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -754,14 +754,24 @@ class Validate(object):
         local = salt.client.LocalClient()
         contents = local.cmd(search, 'pkg.latest_version', ['ceph-common'], tgt_type="compound")
         for minion, version in contents.items():
+            # sometimes, version contains a string value
+            # other times, it contains an empty dict
+            log.debug("VALIDATE ceph_version: minion ->{}<- latest_version version ->{}<-"
+                      .format(minion, version))
             if not version:
                 info = local.cmd(minion, 'pkg.info_installed', ['ceph-common'])
                 if info and 'version' in info[minion]['ceph-common']:
                     version = info[minion]['ceph-common']['version']
+                    log.debug("VALIDATE ceph_version: minion ->{}<- info_installed version ->{}<-"
+                              .format(minion, version))
                 else:
-                    self.errors.setdefault('ceph_version', []).append(
-                        "No Ceph version is available for installation in {}".format(minion))
-                    continue
+                    failmsg = ("No Ceph version is available for installation on minion {}"
+                               .format(minion))
+                    if self.in_dev_env:
+                        log.warning('VALIDATE ceph_version: ' + failmsg)
+                    else:
+                        self.errors.setdefault('ceph_version', []).append(failmsg)
+                        continue
 
             colon_idx = version.find(':')
             if colon_idx != -1:
@@ -769,10 +779,15 @@ class Validate(object):
             dash_idx = version.rfind('-')
             if dash_idx != -1:
                 version = version[:dash_idx]
+            log.debug("VALIDATE ceph_version: minion ->{}<- final munged version ->{}<-"
+                      .format(minion, version))
+            assert isinstance(version, str), "version value is not a string"
+            # FIXME: "10.10" < "10.2" in Python terms, but not in terms of
+            # version numbering semantics
             if version < JEWEL_VERSION:
                 self.errors.setdefault('ceph_version', []).append(
-                    "The Ceph version available in {} is older than 'jewel' (10.2)"
-                    .format(minion))
+                    "The Ceph version available on minion {} ({}) is older than 'jewel' ({})"
+                    .format(minion, version, JEWEL_VERSION))
 
         self._set_pass_status('ceph_version')
 
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -25,7 +25,7 @@ import sys
 import glob
 from subprocess import Popen, PIPE
 from collections import OrderedDict
-from distutils.version import LooseVersion
+from distutils.version import LooseVersion  # pylint: disable=no-name-in-module,import-error,blacklisted-module
 import yaml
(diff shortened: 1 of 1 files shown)


smithfarm:
This code sometimes leads to:

```
2018-03-15T14:10:14.423 INFO:teuthology.orchestra.run.target149202166039.stdout:ceph_version             : [1m[91m["The Ceph version available on minion target149202166039.teuthology ( Zypper command failure: Repository 'SLE-Module-Server-Applications15-Updates' is invalid.\n[Server_Applications_Module_15_x86_64:SLE-Module-Server-Applications15-Updates|https://updates.suse.com/SUSE/Updates/SLE-Module-Server-Applications/15/x86_64/update?12G6EXPLytchxzDtQtrGRUsbLPHhWuDTz1o4dmfmH1MWnPq9oaPxPfuF1GDhPkRpznZvhxzOOI1Lt_LbJdQUy4yGwg75T_cW0WxOycIFoXCObDXyFHhwG7kr7AC27DwZV2gpKwz-ezerz3GbqCboZbpywl-NnXRSKHoi] Valid metadata not found at specified URL\nPlease check if the URIs defined for this repository are pointing to a valid repository.\nSkipping repository 'SLE-Module-Server-Applications15) is older than 'jewel' (10.2)"][0m
```

smithfarm:
sigh

```
[WARNING ] VALIDATE ceph_version: Minion target149202187048.teuthology reports unparseable Ceph version 13.0.1
```

jan--f:
@smithfarm did you consider using using ```distutils.version```? This is sadly undocumented, but nevertheless part if pythons std lib. The [docstrings in the source code](https://hg.python.org/cpython/file/tip/Lib/distutils/version.py) explain things pretty well and stackoverflow has an [example](https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python)

smithfarm:
@jan--f Given the distinct possibility that the version string might be anything (see the commit message of the last commit), and we're only interested in the first two version number components ("X.Y"), I don't think the external module will reduce the complexity here. And indeed the complexity is not high in the first place.

smithfarm:
Note that Salt's `pkg` module uses "zypper" to get the RPM version, instead of "rpm -q". Zypper, in turn, tries to refresh its repos first, and if that fails, `pkg` (IMO stupidly) populates the version string with the error message zypper returns.

This forces us to treat the version number returned by `pkg` with extreme distrust.

srv/modules/runners/validate.py
```diff
@@ -752,27 +752,71 @@ def ceph_version(self):
...
+                failmsg = ("Minion {} reports unparseable Ceph version {}"
+                           .format(minion, version))
+                if self.in_dev_env:
+                    log.warning('VALIDATE ceph_version: ' + failmsg)
+                else:
+                    self.errors.setdefault('ceph_version', []).append(failmsg)
+                    continue
+
+            older = False
+            if int_version_components[0] < LUMINOUS_VERSION[0]:
```
kshtsk: Taking into account that _int_version_compnents and LUMINOUSE_VERSION are integer arrays, you can drop the code which compares each elements one by one, and just use `int_version_components < LUMINOUS_VERSION`. That works well for array with different sizes.

smithfarm: The arrays are guaranteed to be of the same size (2 elements). Your solution to the comparison is elegant, thanks - implementing.

srv/modules/runners/validate.py
```diff
@@ -752,27 +752,71 @@ def ceph_version(self):
...
+            #
+            # We can assume that Ceph version numbers will always begin with X.Y.Z
+            # where X, Y, and Z are integers. Here, we are only interested in X and Y.
+            #
+            # In other words, there must be at least two version components and
+            # both must be convertible into integers.
+
+            try:
+                version_components = version.split('.')
+                int_version_components = [0, 0]
```
kshtsk: It is better to use shorter name for local scope variables, and use array comprehensions. Check the one-liner:
```ver = [int(x) for x in version.split('.')```

**Comment by smithfarm** [comment]
There is another tricky part to this. The original semantics only look at the first two components of the version number (X.Y). So as long as the version is >= 11.2, the validation test should pass.

But the trickiest thing is validating that the components are indeed integers.

**Comment by theanalyst** [comment]
@smithfarm there is something called `distutils.version` in python which should support most standard versioning schemes

So something like
```py

from distutils.version import LooseVersion

LooseVersion(myversion) > LooseVersion('11.2') 

```
should fix without the need to do a lot of manual checks

smithfarm:
@theanalyst What does `distutils.version` say when `myversion` is set to the value " Zypper command failure: Repository 'SLE-Module-Server-Applications15-Updates' is invalid.\n[Server_Applications_Module_15_x86_64:SLE-Module-Server-Applications15-Updates|https://updates.suse.com/SUSE/Updates/SLE-Module-Server-Applications/15/x86_64/update?12G6EXPLytchxzDtQtrGRUsbLPHhWuDTz1o4dmfmH1MWnPq9oaPxPfuF1GDhPkRpznZvhxzOOI1Lt_LbJdQUy4yGwg75T_cW0WxOycIFoXCObDXyFHhwG7kr7AC27DwZV2gpKwz-ezerz3GbqCboZbpywl-NnXRSKHoi] Valid metadata not found at specified URL\nPlease check if the URIs defined for this repository are pointing to a valid repository.\nSkipping repository 'SLE-Module-Server-Applications15" ? (see comments, above)

smithfarm:
Since we have to validate the output from `pkg` anyway, the external module does nothing for us IMO.

kshtsk:
@smithfarm The best way to validate particular version is to set an agreement on how a version should look like and use a regex for initial validation. For parsing a version I think it is okay to use standard libraries, if only it is not changed dramatically in next version of python.

theanalyst:
> But the trickiest thing is validating that the components are indeed integers, before embarking on the comparison.

regular expression or something like 
`all(s.isdigit() for s in version_str.split(".")[0:2])`

smithfarm:
Changelog:

* keep version number as a string and use `isdigit()` and `split()` to validate contents
* use `LooseVersion` from distutils for comparing version strings

smithfarm:
Does anyone understand the lint error?

```
E: 28, 0: No name 'version' in module 'distutils' (no-name-in-module)
E: 28, 0: Unable to import 'distutils.version' (import-error)
W: 28, 0: Uses of a blacklisted module 'distutils.version': Please use 'from salt.utils.versions import LooseVersion' (blacklisted-module)
```

smithfarm:
Changing `distutils.version` to `salt.utils.version` doesn't help:

```
E: 28, 0: No name 'version' in module 'salt.utils' (no-name-in-module)
E: 28, 0: Unable to import 'salt.utils.version' (import-error)
C: 31, 0: Imports from package salt are not grouped (ungrouped-imports)
```

smithfarm:
At least one shop is saying to not use distutils.version anymore: https://github.com/postmarketOS/pmbootstrap/issues/342

smithfarm:
Note to self: saved the old (`try...except`) version in `wip-ceph-version-orig`

smithfarm:
According to https://github.com/PyCQA/pylint/issues/73 the Travis CI implementation of pylint is running in a virtualenv, and that (apparently) causes pylint to report bogus errors.

smithfarm:
Tested; working:

```
2018-03-19T21:56:11.128 INFO:teuthology.orchestra.run.target137074026022.stdout:[DEBUG   ] VALIDATE ceph_version: minion ->target137074026022.teuthology<- info_installed version ->13.0.1<-
2018-03-19T21:56:11.129 INFO:teuthology.orchestra.run.target137074026022.stdout:[DEBUG   ] VALIDATE ceph_version: minion ->target137074026022.teuthology<- final munged version ->13.0.1<-
2018-03-19T21:56:11.129 INFO:teuthology.orchestra.run.target137074026022.stdout:[INFO    ] VALIDATE PASSED  deepsea_minions          : [1m[92mvalid[0m
2018-03-19T21:56:11.129 INFO:teuthology.orchestra.run.target137074026022.stdout:[INFO    ] VALIDATE PASSED  master_minion            : [1m[92mvalid[0m
2018-03-19T21:56:11.129 INFO:teuthology.orchestra.run.target137074026022.stdout:[INFO    ] VALIDATE PASSED  ceph_version             : [1m[92mvalid[0m
```

PR #983, opened 2018-03-05
QA test  - migration from filestore to bluestore

markostanojlovic:
# QA test  - migration from filestore to bluestore 

New QA test case for testing migration from filestore to bluestore:
- Disabled restart in stage 0
- Cluster deployed with filestore OSDs
- Then migration run with DS command ceph.migrate.osds
- No services deployed, not special settings for OSDs

New functions added to common.sh :
- configure_OSD_to_filestore
- check_OSD_type
- migrate_to_bluestore
- disable_restart_in_stage_0

### DIFF
--- a/qa/common/common.sh
+++ b/qa/common/common.sh
@@ -727,15 +727,29 @@ EOF
     _run_test_script_on_node $TESTSCRIPT $STORAGENODE
 }
 
-function configure_OSD_to_filestore {
+function configure_all_OSDs_to_filestore {
 	salt-run proposal.populate format=filestore name=filestore 
 	chown salt:salt /srv/pillar/ceph/proposals/policy.cfg
 	sed -i 's/profile-default/profile-filestore/g' /srv/pillar/ceph/proposals/policy.cfg
 }
 
-function check_OSD_type {
-	salt \* pillar.get ceph
-	salt \* cmd.run "for i in \$(mount|grep ceph|awk '{print \$3}');do cat \${i}/type;done"
+function verify_OSD_type {
+    # checking with 'ceph osd metadata' command
+    # 1st input argument: type 'filestore' or 'bluestore'
+    # 2nd input argument: OSD ID 
+    osd_type=$(ceph osd metadata $2 -f json-pretty | jq '.osd_objectstore')
+    if [[ $osd_type != \"$1\" ]]
+        then 
+        echo "Error: Object store type is not $1 for OSD.ID : $2"
+        exit 1
+    else
+        echo OSD.${2} $osd_type
+    fi
+}
+
+function check_OSD_type {  
+    # expecting as argument 'filestore' or 'bluestore' 
+    for i in $(ceph osd ls);do verify_OSD_type $1 $i;done
 }
 
 function migrate_to_bluestore {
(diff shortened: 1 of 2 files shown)


jschmid1:
@kshtsk could you please add @markostanojlovic to the list of contributors?

kshtsk:
@jschmid1 anyone in SUSE should be whitelisted, but I can't see him there. Could you help and add Marko to SUSE organization?

smithfarm:
@kshtsk @jschmid1 Marko should write to github-owners@suse.de and ask them to be added.

jschmid1:
And would you please change the title to something meaningful :)? @markostanojlovic

swiftgist:
@smithfarm Are you okay with this PR?

qa/common/common.sh
```diff
@@ -726,3 +726,29 @@ echo "Result: OK"
 EOF
     _run_test_script_on_node $TESTSCRIPT $STORAGENODE
 }
+
+function configure_OSD_to_filestore {
```
smithfarm: this function appears to change *all* the OSDs to filestore? Maybe the function name could be changed to reflect that, e.g `configure_all_OSDs_to_filestore`

qa/common/common.sh
```diff
@@ -726,3 +726,29 @@ echo "Result: OK"
...
+	salt-run proposal.populate format=filestore name=filestore 
+	chown salt:salt /srv/pillar/ceph/proposals/policy.cfg
+	sed -i 's/profile-default/profile-filestore/g' /srv/pillar/ceph/proposals/policy.cfg
+}
+
+function check_OSD_type {
+	salt \* pillar.get ceph
+	salt \* cmd.run "for i in \$(mount|grep ceph|awk '{print \$3}');do cat \${i}/type;done"
```
jschmid1: This function does not verity the expected OSD_TYPE. That effectively means that you would need to inspect the logs in order to find out if the test passed or not.

Adding a 'expected' param which can be set to either filestore or bluestore will help.

qa/suites/ceph-test/fs_to_bs_migration.sh
```diff
@@ -0,0 +1,40 @@
...
+policy_cfg_storage # no node will be a "client"
+configure_OSD_to_filestore
+cat_policy_cfg
+run_stage_2
+ceph_conf_small_cluster
+ceph_conf_mon_allow_pool_delete
+run_stage_3
+ceph_cluster_status
+ceph_health_test
+check_OSD_type
```
jschmid1: .. continue from comment#1

passing 'filestore'

qa/suites/ceph-test/fs_to_bs_migration.sh
```diff
@@ -0,0 +1,40 @@
...
+migrate_to_bluestore
+ceph_health_test
+check_OSD_type
```
jschmid1: passing 'bluestore'

qa/common/common.sh
smithfarm: note that salt returns 0 regardless of whether the remote command succeeded or not, so you have to grep the results. In this case, it's a simple case of grepping for any instances of the "other" possibility. For example, if all the OSDs are expected to be filestore, grep the output for bluestore. And vice versa.

The mere absence of a string in the output is not sufficient as success criteria, however. I would also count the number of instances of the string that is expected and check that it is the same as the number of OSDs in the cluster. (There is a function `json_total_osds` for determining this.)

PR #988, opened 2018-03-07
Makefile: reorganize target deps

jan--f:
Just some reorganization to make the target dependencies a bit more explicit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,24 +21,6 @@ endif
 endif
 endif
 
-
-usage:
-	@echo "Usage:"
-	@echo -e "\tmake install\tInstall DeepSea on this host"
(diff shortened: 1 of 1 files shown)


jschmid1:
That will break `make rpm` I fear..

`make rpm` calls `rpmbuild -bb deepsea.spec` which calls `pyc` in `%build` and `copy-files` in `%install`.

Either we keep `setup.py` a prerequisite of `pyc` or add a second step to `%build` in the specfile which calls `make setup.py`

What's the reason to change this in the first place?

jan--f:
@jschmid1 I don't even remember my though process anymore. This is probably wrong...closing.

Issue #993, opened 2018-03-08
python abi


swiftgist:
### Description of Issue/Question

With the latest master, a locally built rpm gains the requirement

python(abi) = 3.4

The result is that the rpm installation will fail on a SLE15 system unless --nodeps is used.

The requirement is tied to the deepsea cli.  The python 3.4 comes from the host system, but the cli should not really need to be hardcoded to a specific abi.

rjfd:
>  The python 3.4 comes from the host system

Are you referring to the OBS host that builds the RPM?

swiftgist:
No, I was building locally.  But even in the OBS case, if I as a user had been using DeepSea and my python version changes from 3.X to 3.Y, I would expect something as portable as python to keep working.  I know this isn't an api >= 3.4 requirement, but it seems like it should be.

Also, pysetup seemed to support the --no-autoreq at one point.

lyzenghq:
I also encountered this problem, and my deepsea does not work properly.

jschmid1:
Can be circumvented by building with the correct python abi

PR #994, opened 2018-03-08
Create cli subpackage

swiftgist:

Partly addresses #993 by moving the abi requirement.  Also, I think the requires communicate a bit better to anyone wanting to understand the dependency differences between the Salt portions and the cli.

Also, I do not know if I have the best wording for the cli description.


tserong:
Is it worth adding a Recommends: deepsea-cli to the base deepsea package, to encourage everything to be installed by default?

jan--f:
Other then @tserong's comment lgtm. Should make things much cleaner.

swiftgist:
Added the recommends.

PR #1005, opened 2018-03-13
Tuning Chrony

swiftgist:
The 1 second default is too large for Ceph.  Trying 1/10th.

While experimenting with the qa suite, Ceph complained about a clock skew.  Restarting chrony  had no effect because the time difference was 120ms.

### DIFF
--- a/srv/salt/ceph/time/chrony/files/chrony.conf.j2
+++ b/srv/salt/ceph/time/chrony/files/chrony.conf.j2
@@ -12,8 +12,8 @@ server {{ server }} iburst
 driftfile /var/lib/chrony/drift
 
 # In first three updates step the system clock instead of slew
-# if the adjustment is larger than 1 second.
-makestep 1.0 3
+# if the adjustment is larger than 0.1 second.
+makestep 0.1 3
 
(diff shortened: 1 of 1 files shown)


tserong:
Likewise, sorry. The change appears syntactically correct, so I have no objections on that front, at least :-) but I have no idea whether or not this setting will have any undesirable side effects.

Martin-Weiss:
Hm - I have not much idea about Chrony - but with NTPD one should not use burst against public sources and it is a best practice (also default in SLES 12) to use "iburst". 

So could you try out iburst instead of burst with chrony , too?

swiftgist:
I stuck with the default configuration for chrony.  The line I changed means it will try to step the time up to 3 times on startup *if* the time is off by greater than 1/10th of a second.  So, it will not do this endlessly.

@Martin-Weiss The config is using iburst.

Martin-Weiss:
Just to understand this right - we do NOT use burst 2/10?

--> see also http://www.bircd.org/ntpserver.html

> you must not use BURST, except for temporary testing purposes. you may use IBURST.

If I remember right the problem was that BURST is generating load on the NTP server.. which is bad for public sources - but fine for internal sources most often ;-).

PR #1012, opened 2018-03-15
[backport][SES5]populate: support multiple public, cluster networks in engulf (bsc#1076010)

tserong:
This removes the assumption in the engulfer that there can be no more
than one public or cluster network.  Without this change, running engulf
on a cluster with (e.g.) public_network = 10.27.21.0/24, 10.27.20.0/24
will result in the following error:

  [ERROR   ] Failed to determine cluster's public_network.
  [ERROR   ] Failed to replace derived public_network with public_network
             of existing cluster.

Note that this still leaves the public_network and cluster_network settings
as strings in the pillar data (changing the pillar data to be an array would
have been somewhat more invasive).

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1076010
Fixes: https://github.com/SUSE/DeepSea/issues/582
(cherry picked from commit 205ed7c10f38e49db0400516e816241a6f5204c6)

Fixes #

Description:

-----------------

### DIFF
--- a/srv/modules/runners/populate.py
+++ b/srv/modules/runners/populate.py
@@ -972,11 +972,12 @@ def _replace_key_in_cluster_yml(key, val):
 
     return True
 
-def _get_existing_cluster_network(addrs, public_network=None):
+
+def _get_existing_cluster_networks(addrs, public_networks=[]):
     """
-    Based on the addrs dictionary { minion: [ ipaddress ] }, this function
-    returns an address consisting of network prefix followed by the cidr
-    prefix (ie. 10.0.0.0/24), or None.
+    Based on the addrs dictionary { minion: [ipaddress] }, this function
+    returns a list of addresses consisting of network prefix followed by the
+    cidr prefix (e.g. [ "10.0.0.0/24" ]).  It may return an empty list.
     """
     target = deepsea_minions.DeepseaMinions()
     search = target.deepsea_minions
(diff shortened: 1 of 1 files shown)


tserong:
This is a backport of #951.

Note: Because I'm paranoid, I still want to do at least one manual test of an engulf with this on a SES5 system, so please don't merge this yet until I've done that.

tserong:
Manual test went fine, safe to merge.

For the record, with vagrant-ceph, tweak this:

````
--- a/lib/settings.rb
+++ b/lib/settings.rb
@@ -10,8 +10,8 @@ def common_settings(node, config, name)
       # Ceph has three networks
       networks = config[CONFIGURATION]['nodes'][name]
       node.vm.network :private_network, ip: networks['management']
-      node.vm.network :private_network, ip: networks['public']
-      node.vm.network :private_network, ip: networks['cluster']
+      node.vm.network :private_network, ip: networks['public'], :libvirt__forward_mode => 'route'
+      node.vm.network :private_network, ip: networks['cluster'], :libvirt__forward_mode => 'route'
 end
````

Then you can put different cluster and public networks on different nodes in config.yml, and they'll all talk to each other.

jschmid1:
> This is a backport of #951 (please let me know if I'm doing the right thing with the labels and whatnot here).

That information might go into the body of the PR. I also use the title to 'flag' it as a [backport] which goes to [branch] -> SES5 in that case.

besides that.

:+1:

Issue #992, opened 2018-03-08
Stage 2 (non-CLI) fails because node_exporter not present


smithfarm:
### Description of Issue/Question

Stage 2 fails on openSUSE Leap 15

Command used to run Stage 2: `salt-run --no-color state.orch ceph.stage.2`

```
2018-03-08T16:34:43.957 INFO:teuthology.orchestra.run.target149202180192.stdout:yaml_syntax              : [1m[92mvalid[0m
2018-03-08T16:34:43.957 INFO:teuthology.orchestra.run.target149202180192.stdout:public network           : [1m149.202.180.192/32[0m
2018-03-08T16:34:43.957 INFO:teuthology.orchestra.run.target149202180192.stdout:cluster network          : [1m149.202.180.192/32[0m
2018-03-08T16:34:43.957 INFO:teuthology.orchestra.run.target149202180192.stdout:target149202180192.teuthology_master:
2018-03-08T16:34:43.957 INFO:teuthology.orchestra.run.target149202180192.stdout:  Name: push.proposal - Function: salt.runner - Result: Changed Started: - 16:32:30.178597 Duration: 1008.777 ms
2018-03-08T16:34:43.957 INFO:teuthology.orchestra.run.target149202180192.stdout:  Name: refresh_pillar1 - Function: salt.state - Result: Changed Started: - 16:32:31.187668 Duration: 6810.168 ms
2018-03-08T16:34:43.957 INFO:teuthology.orchestra.run.target149202180192.stdout:  Name: advise.networks - Function: salt.runner - Result: Clean Started: - 16:32:37.998071 Duration: 1269.634 ms
2018-03-08T16:34:43.957 INFO:teuthology.orchestra.run.target149202180192.stdout:  Name: admin key - Function: salt.state - Result: Changed Started: - 16:32:39.268778 Duration: 742.155 ms
2018-03-08T16:34:43.958 INFO:teuthology.orchestra.run.target149202180192.stdout:  Name: mon key - Function: salt.state - Result: Changed Started: - 16:32:40.011353 Duration: 543.103 ms
2018-03-08T16:34:43.958 INFO:teuthology.orchestra.run.target149202180192.stdout:  Name: mgr key - Function: salt.state - Result: Changed Started: - 16:32:40.554686 Duration: 1800.376 ms
[...]

PR #1013, opened 2018-03-15
monitoring: remove conditional for node_exporter config

jan--f:
The jinja conditional does not reliably evaluated after the package is
installed. Until a better solution is found, assume node_exporter >= 0.15.
Fixes: #992

Fixes #

Description:

-----------------


smithfarm:
With this PR, `health-ok.sh` passes for me in OVH in a single-minion configuration.

smithfarm:
This makes other test scripts pass on SLE15, too - not just `health-ok.sh`.

jan--f:
@swiftgist nay objections to this? Should make it into the next package.

Issue #1009, opened 2018-03-14
rgw_exporter requires python3-prometheus-client


jan--f:
But deepsea attempts to install python-prometheus-client .

Issue #998, opened 2018-03-09
distro friendly init.sls


swiftgist:
Currently, we have 12 files that have specific hardcoded distro names

```
./repo/default.sls
./monitoring/grafana/init.sls
./monitoring/prometheus/exporters/rbd_exporter.sls
./monitoring/prometheus/exporters/ceph_rgw_exporter/default.sls
./monitoring/prometheus/exporters/node_exporter.sls
./monitoring/prometheus/exporters/node_exporter_package.sls
./monitoring/prometheus/alertmanager.sls
./monitoring/prometheus/init.sls
./time/ntp/default.sls
./ganesha/install/default.sls
./ganesha/service/default.sls
./packages/common/default.sls
```

Honestly, I do not like the conditionals.  From a development perspective, there's always a chance - no matter how small - of breaking some other distro behavior.  From a user perspective, the file requires some mental gymnastics to understand that Jinja is running Salt commands as conditionals.  When exceptions are added for specific versions of distros, then even more debugging effort is required if the sls file is not running as expected.

I experimented with a fictional component in /srv/salt/ceph/ex.  Here's a different approach for the init.sls:

```
{% set osfinger = grains.get('osfinger') %}
{% set os = grains.get('os') %}
{% set osrelease = os + "-" + grains.get('osrelease') %}

{% set abspath = "/srv/salt/" + slspath %}
{% set custom = salt['pillar.get']('ex_init', 'not a file') %}

include:
{% if salt['file.directory_exists'](abspath + "/" +  custom) %}
  - .{{ custom }}
{% elif salt['file.directory_exists'](abspath + "/" +  osfinger) %}
  - .{{ osfinger }}
{% elif salt['file.directory_exists'](abspath + "/" + osrelease) %}
  - .{{ osrelease }}
{% elif salt['file.directory_exists'](abspath + "/" + os) %}
  - .{{ os }}
{% else %}
  - .default
{% endif %}
```

Understand that this would not be throughout DeepSea and that the init.sls would only be for the directories of files listed above.  If no distro specific directory exists, then the behavior would be exactly the same as it is currently.  That is,  a custom file specified in the Salt pillar would have precedence, but the default.sls would be called normally.

If any of `osfinger`, `osrelease` or `os` directories exist, then that would be included.  If multiple exist, then the most specific is preferred.

***  Why have osrelease if Salt provides osfinger?
[...]

PR #1016, opened 2018-03-15
refactor ceph_rgw_exporter

swiftgist:

Description:
SLE-15 changed package name, implement distro init.sls
This addresses the same issue as #1010, but using the #998 approach.  I ran into the same problem #1009 but thought this would work as an ideal example since the original change is minor.

The default is still python-prometheus-client, but CentOS and SLES-15 have specific package names.  All three share default-exporter logic, so no duplicate code.  I think this is about as ideal as it gets for refactoring a configuration into DRY setup that is fairly obvious.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -289,6 +289,10 @@ copy-files:
 	install -m 644 srv/salt/ceph/monitoring/prometheus/exporters/files/* $(DESTDIR)/srv/salt/ceph/monitoring/prometheus/exporters/files
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter
 	install -m 644 srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter/*.sls $(DESTDIR)/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter/CentOS
+	install -m 644 srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter/CentOS/*.sls $(DESTDIR)/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter/CentOS
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter/SLES-15
+	install -m 644 srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter/SLES-15/*.sls $(DESTDIR)/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter/SLES-15
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/monitoring/prometheus/exporters/ceph_rgw_exporter/cron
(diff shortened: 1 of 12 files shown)


smithfarm:
@swiftgist Does this support openSUSE 15? (It should be treated the same as SLE15 in most - if not all - cases.) I would like to get to a point where both SLE15 and openSUSE 15 are tested. Actually, after today I think we are nearly there.

rjfd:
@swiftgist as we discussed previously having an sls file per OS looks like a nice idea, but after seeing the changes that are made in this PR and compare to the current solution of a single file with jinja conditionals, I'm a bit skeptical that this is the best solution.

Adding another jinja conditional to the already existing sls, seems more easier to read, understand, and maintain.

jan--f:
I think this looks pretty alright. I see a some advantage of this over only using conditionals in the default.sls.

1. New distro support can be added by simply dropping in the respective directory (if necessary) instead of adding a new if clause (I find jinja conditionals are somewhat error prone to write).
2. It's straight forward to add specialisation (say if Debian 8 and 9 would need different states) or generalisation.

While all this is possible with a state file and a bunch of if statements, if find this is a nicer abstraction. If I'm not mistaken we could even abstract most of the init.sls (and the distro choosing code) out and include it from the salt root.

Granted this abstraction adds some complexity, but I think the advantages outweigh. A bit of documentation can easily lower the hurdle for understanding this.

tl;dr: +1

jan--f:
@tserong I think that is what the ```os``` grains does.

tserong:
> @tserong I think that is what the os grains does.

OK, apparently I'm confused. I'd _swear_ I had at some time in the past seen the `os` grain with something weird in it (like `openSUSE Tumbleweed`), but I've just tested here on SLE 12 SP3, Tumbleweed and Leap 42.3, and the `os` grain in all cases just gives me a nice, simple "SUSE". So I think we can safely ignore my earlier comment.

swiftgist:
Trying to answer all items... if I miss something, bring it up.

@smithfarm Yes about openSUSE 15, but I have not tested so I didn't create it.  If it turns out that Leap, Tumbleweed and SLE are identical, then renaming SLE-15 to SUSE (that's the Salt grain name) would be the right thing to do.  If Leap and Tumbleweed are identical, but SLE is the odd one out, then creating SUSE with what's needed and keeping SLE-15 would behave correctly.

I expect this is where we are flying close to the sun with some distros and that many small PRs will keep adjusting this as needed.  And it's perfectly okay to flip everything around.  If SLE-15 seems to match more distros than the default, we update the default, remove SLE-15 and create any exceptions for what we know.

@rjfd I think my fear is that changes over time will diverge from one another across several files.  One dev will use 'osfinger', another will just use 'os'.  As a user, it's really a pain to have to decipher code to guess if the right command is running.

And I did take this example to an extreme of including a common file, but I thought it was fair to show a worse case example and not a contrived one.

@jan--f I think I was starting small here in the "well, it's only ~16 files and the init.sls is more complicated than the two liners".  I think I would still like to go with an incremental approach to see if something horrid is really lurking in the shadows. 

@tserong I didn't check osfamily, but I'm fairly certain os is single words with no spaces.  With respect to Salt and central things, I find some of these difficult.  I have started avoiding using `include` in any file unless it's by itself.  The consequence of explaining that the `include` runs first regardless of position is not worth the effort.

Note that the custom attribute still matches the directory like the original init.sls.  As for generalizing, I suppose we could always look at Jinja macros. ;)  That may make Jan ill though  (me too probably).

And I just read your last comment.  

***
Does this init.sls need a paragraph explaining what to do?  Is that the best place?  Or do we keep the same README in the areas that need it?  My instinct says that somebody wanting to help shouldn't have to go searching for the right way to submit a PR.

jschmid1:
I agree with @jan--f here. It might look intimidating at a first glance but the PROs definitively prevail the CONs here.

+1 from me.

PR #1023, opened 2018-03-19
Support kwargs in orchestrations

swiftgist:
The proposal runner handles kwargs, but Salt is passing the kwargs dict
as a keyword 'kwargs'.  This change allows Stage 1 to be part of the
process rather than requiring manual discovery.

### DIFF
--- a/srv/modules/runners/proposal.py
+++ b/srv/modules/runners/proposal.py
@@ -121,6 +121,11 @@ def _parse_args(kwargs):
     """
     args = STD_ARGS.copy()
     args.update(kwargs)
+    if 'kwargs' in kwargs:
+        args.update(kwargs['kwargs'])
+        args.pop('kwargs')
+    log.info("args: {}".format(args))
+
     if args.get('name') == 'import':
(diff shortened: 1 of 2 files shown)


srv/modules/runners/proposal.py
```diff
@@ -121,6 +121,10 @@ def _parse_args(kwargs):
     """
     args = STD_ARGS.copy()
     args.update(kwargs)
+    if 'kwargs' in kwargs:
+        for key in kwargs['kwargs']:
```
jan--f: Wouldn't it be more efficient to use ```dict.update()``` here? Something like
```
if 'kwargs' in kwargs:
    args.update(args.pop('kwargs')
```

swiftgist: Will be circling back soon :)

swiftgist:
Tested and behaves.  I found no trivial method for creating a functional test.

jschmid1:
are we ready to merge here?

swiftgist:
Yeah, I made the change.

swiftgist:
SUSE internal: bsc#1078508

PR #1026, opened 2018-03-21
Add active+clean wait for migrations

swiftgist:
Before removing an OSD, ensure that all PGs are active+clean.
Includes unit and functional tests.


swiftgist:
I am aware of the "osd safe-to-destroy" command now, but this is not working in python 3 for me.  

I believe logging the states will help in the long run since if a timeout occurs, the admin will have the information on the state of the Ceph cluster at the time of the migration.

I have put the tests and smoketests in the qa package.  A separate PR can handle calling this from teuthology.

smithfarm:
I think we have one integration test case that exercises this code: `suites/ceph-test/fs_to_bs_migration.sh` - correct? This test does `salt-run state.orch ceph.migrate.osds` to migrate all the OSDs.

How would we integrate the functional tests into that script?

swiftgist:
Keeping with the theme of the other salt-run calls, I would think an additional function in common.sh

```
function smoketests_quiescent {
  salt-run ceph.smoketests.quiescent
}
```
would run both tests.  Adding this to  `fs_to_bs_migration.sh` at the point that PGs actually exist seems reasonable.  

I am avoiding going into tangents about other migrations and whether those would be separate and if so, is this test added to all of them?  

Now, I cannot test that change in Teuthology and I really do not want every PR to be blocked on merging until we have a Teuthology run.  Also, this was mainly a single example that I'm hoping to use as a reference for what may be expected for future PRs.

If we want to banter about what comments I should have put and where, I think that's fine.  I am undecided on more expressive Salt IDs vs. comments that only somebody looking at the file will find.

smithfarm:
I studied the functional test code, and here is what I think it does. There are two test cases - positive and negative. The positive test case chooses a storage node at random, and checks that the PGs on the first OSD it finds on that node are in "active+clean" state. The negative test uses systemctl to stop the OSD and then checks that the same PGs are not in "active+clean" state.

So this test could be added to the set of "basic" functional tests that can be run on any cluster deployed by DeepSea, regardless of which roles are used, since there will always be at least one storage node.

smithfarm:
For `fs_to_bs_migration.sh`, since it migrates all the OSDs, I think we need to:

1. before starting the migration:
a. populate the cluster with data
b. check that all the PGs are in active+clean state
2. run the migration
3. after the migration completes, check that
a. all the OSDs are bluestore
b. all the PGs are (still) in active+clean state

But I'm certainly not suggesting that this improvement of `fs_to_bs_migration.sh` should block the merging of this PR.

smithfarm:
Correction, I see now that the positive test checks that *all* PGs are in "active+clean" state. So we can just add (correct me if I'm wrong)

```
salt-run ceph.tests.quiescent
```

before and after the migration.

smithfarm:
One last question before I sign off: What happens when scrubbing/deep-scrubbing starts on a PG while the migration is happening? Does it cause the PG state to change to something other than "active+clean"? If so, that might interrupt the running migration.

smithfarm:
I played a little today with the "scrubbing" and "deep" PG states. Under normal circumstances (small cluster, small amount of data) scrubbing/deep-scrubbing does not take long enough to cause the PG states reported by "ceph pg stat" to change from "active+clean".

I suspect it would need a PG with a lot more data in it, or some other not-so-easily-reproducible condition, for the scrubbing to take long enough to produce a visible change in state.

The other thing I noticed is that this code seems to be checking the state of all PGs in the cluster, instead of just those PGs that are on the OSD being migrated. But maybe it's better to demand a completely quiescent cluster before migrating any OSD.

swiftgist:
Adding safe-to-destroy check in the wait of OSDWeight.  This impacts redeploy and remove.osd.

While having both active+clean and safe-to-destroy checks seem like overkill, the original problem was https://github.com/SUSE/DeepSea/blob/master/srv/salt/_modules/osd.py#L365.  Turns out that is called with an int from remove.osd, but as a string from redeploy.  Fun, fun.  This PR is somewhat an overreaction.

I will say that safe-to-destroy is somewhat newish in that it's less than a year old while the `pg stat` call has been around.  I think the logging will be more helpful for those debugging what is happening.  Having the migration pause when Ceph gets busy is probably a good thing.  Also, the timeout/delays are different between the two (i.e. 60/6 for remove, 120/12 for redeploy).

tserong:
I don't understand how srv/salt/ceph/tests/quiescent/timeout/init.sls works. I mean, I know it's stopping an OSD then checking that it's PGs aren't active+clean, but shouldn't `osd.ceph_quiescent` fail, thus failing the state? Or are the two `test.nop`s and `onfail: test: passes` thingy making this magic happen?

srv/salt/_modules/osd.py
```diff
@@ -382,28 +392,104 @@ def wait(self):
...
+    def quiescent(self):
+        """
+        Wait until PGs are active+clean or timeout is reached.  Default is a
+        2 minute sliding window.
+        """
+        i = 0
+        last = []
+        while i < self.settings['timeout']/self.settings['delay']:
+            current = self.pg_states()
+            if len(current) == 1:
```
jan--f: I think I'd be happier if we'd actually test for active+clean as well. Maybe 
```
if len(current) == 1 && current[0] == 'active+clean':
```

srv/salt/_modules/osd.py
```diff
@@ -382,28 +392,104 @@ def wait(self):
...
+            raise RuntimeError("connection error: {}".format(error))
+
+    def quiescent(self):
+        """
+        Wait until PGs are active+clean or timeout is reached.  Default is a
+        2 minute sliding window.
+        """
+        i = 0
+        last = []
+        while i < self.settings['timeout']/self.settings['delay']:
```
jan--f: Only because we use it in a division here and its a config, maybe check that delay is not 0? Pretty sure some of our usual suspects will try to set delay to 0 at some point...to make things go faster.

swiftgist: I thought of that for a second and wondered exactly what somebody would expect the correct behavior is?  That we would try to busy wait a cpu?  

Do you think this is just deserving of a specific exception?  That is, RuntimeError("Delay cannot be 0")?

srv/salt/_modules/osd.py
swiftgist: I suppose the short circuit will prevent additional checks each time through the loop.  That would be award winning if somebody could get their cluster into a state where all PGs are in 'active+' something to have only one entry.

Would it be bad to add a log message that says "What did you do?" :)

It's fine, I'll add this.

srv/salt/_modules/osd.py
jan--f: I think the built-in ValueError would be appropriate.

srv/salt/_modules/osd.py
jan--f: Well if I understand the code correctly this checks if the ```num_pg_by_state``` list has the length 1, regardless of what that state is. And yes the condition is evaluated lazily.

swiftgist:
@tserong Right, the module fails by calling the exception, but that's what we wanted to test.  The onfail prevents the test state from failing and calls another state.  I didn't need anything to happen, so I picked test.nop.

I added a failure in this last change to catch if the timeout doesn't fail as expected.  Basically, we are back to goto's. :)

PR #1044, opened 2018-03-28
Correct conditional

swiftgist:
If the pool and image exist, generate a full lrbd.conf.  When neither
exists, use the minimal configuration.


rjfd:
@swiftgist wasn't that state file working before? is this change required because the return value semantics of `salt['cmd.run']` changed in a newer salt version?

swiftgist:
@rjfd I do not believe it ever worked for me, but was too low a priority to investigate further at the time.

PR #1047, opened 2018-03-30
Add keyring smoketest

swiftgist:
Initial verification that keyrings work.  Some check restrictions.

### DIFF
--- a/srv/salt/ceph/tests/keyrings/README
+++ b/srv/salt/ceph/tests/keyrings/README
@@ -1,0 +1,8 @@
+The general idea is that keyrings work.  Over time, keyrings will be restricted to only the permissions necessary.  All sls files are checking that a ceph query using the respective keyring returns successfully.  Some sls files check that a ceph query fails.
+
+These should never fail, but accidental modification would find an issue immediately.
+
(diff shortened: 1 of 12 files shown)


srv/salt/ceph/tests/keyrings/README
```diff
@@ -0,0 +1,8 @@
+The general idea is that keyrings work.  Over time, keyrings will be restricted to only the permissions necessary.  All sls files are checking that a ceph query using the respective keyring returns successfully.  Some sls files check that a ceph query fails.
+
+These should never fail, but accidental modification would find an issue immediately.
+
+Extend these sls files if specific ceph commands must always work for a specific keyring and could be disabled by an incorrect capabilities setting.  Tests of Ceph are unnecessary here.
+
+The init.sls includes all other files.  The individual files allow checking specific keyrings.
```
smithfarm: Would be nice to insert some newlines into this file, so the lines don't disappear beyond the right edge of the screen/window/whatever.

jan--f: Isn't this just github's formatting? My editor breaks lines fine.

PR #1049, opened 2018-04-03
Return a single device

swiftgist:
The find command may return multiple symlinks, but only one is needed.

SUSE internal: bnc#1080102


jschmid1:
can't easily review without digging deeper into the code.
Maybe @tserong could review?

tserong:
LGTM, although the rstrip() is probably no longer necessary. If `find` returns a single line with '\n' at the end, split() will give an array with one member _without_ the '\n', and `[-1]` will give that member. If there's multiples, then `[-1]` just gives the last member, again without the '\n'.

swiftgist:
Removed rstrip().

PR #1051, opened 2018-04-04
Add salt2018 support

### LINKED ISSUE #995: trackerissue: salt2018 changes
As of today salt 2018 landed in the TEST repos which requires some changes again.

1) Utility modules:

       https://docs.saltstack.com/en/latest/topics/utils/index.html

2) 'saltutil' __virtual__ returned False: module 'salt.utils.files' has no attribute 'fopen'
      falling back to `open` but with a context manager

3)  [WARNING ] /srv/modules/runners/__pycache__/filequeue.cpython-36.py:336: DeprecationWarning: Use of 'salt.utils.alias_function' detected. This function has been moved to 'salt.utils.functools.alias_function' as of Salt 2018.3.0. This warning will be removed in Salt Neon.
      moved to correct path.

4) importing changed from `imp` to `importlib`
    switching to proper salt style importing.

jschmid1:

Fixes #995 

Description:

* Remove filequeue.add/pop aliases
   - salt moved the alias function paths in [salt2018 ](https://docs.saltstack.com/en/latest/topics/releases/2018.3.0.html) 
   - we almost nowhere used the aliases so I dropped them to avoid
      unnessasary if/else branching when supporting salt2017/18

* Changed deepsea_minions importing
    - to `from deepsea_minions import DeepseaMinions`
    - from `import deepsea_minions`

* Change importing of modules
   - we previously used the pythonic way of importing
   - switched to the salt way  `__salt__['helper.run']`
   - and `__salt__['helper.convert_out']`

* Fix a bug in osd.py:L970
   - wrong partition handling in case of a NVME.
   - NVME partition schemes haunt us since.. ever..
      Maybe it's time for a code change here.

* Fix tests
   - reworked tests:
   - docs can be found in the docstring:
      **https://github.com/SUSE/DeepSea/pull/1051/files#diff-99735a509ed582a25412d9b35e31a592R9**

Caveats:

* There is a bug in salt(presumably) which requires the user to delete the
   `/srv/modules/runners/__pycache__` directory in order to get the runners
    imported properly.

* Skipped tests:
   I had to skip a couple of tests in the test_osd.py::Test_is_incorrect suite due to 
   flickering tests and a missbehaving `pyfakefs` module. More information here https://github.com/SUSE/DeepSea/issues/1052. Has to get reworked.

-----------------

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,7 @@ nosetests.xml
 coverage.xml
 *,cover
 .hypothesis/
+.pytest_cache/
 
 # Translations
 *.mo
--- a/srv/modules/runners/filequeue.py
+++ b/srv/modules/runners/filequeue.py
@@ -240,19 +240,11 @@ def help_():
              '        salt-run filequeue.queues\n'
              '\n\n'
              'filequeue.enqueue:\n'
-             'filequeue.add:\n'
-             'filequeue.push:\n\n'
              '    Add an item on to a queue\n\n'
              '    CLI Example:\n\n'
              '        salt-run filequeue.enqueue abc\n'
              '        salt-run filequeue.enqueue abc queue=prep\n'
              '        salt-run filequeue.enqueue item=abc queue=prep\n'
-             '        salt-run filequeue.add abc\n'
-             '        salt-run filequeue.add abc queue=prep\n'
-             '        salt-run filequeue.add item=abc queue=prep\n'
-             '        salt-run filequeue.push abc\n'
-             '        salt-run filequeue.push abc queue=prep\n'
-             '        salt-run filequeue.push item=abc queue=prep\n'
              '\n\n'
              'filequeue.dequeue:\n\n'
              '    Remove and return oldest item from a queue\n\n'
@@ -333,8 +325,6 @@ def enqueue(queue=None, **kwargs):
 
 
 # pylint: disable=invalid-name
-add = salt.utils.alias_function(enqueue, 'add')
-push = salt.utils.alias_function(enqueue, 'push')
 
 
 def dequeue(**kwargs):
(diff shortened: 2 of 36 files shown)


swiftgist:
I would like to keep the add and push aliases for filequeue.  I did not know about Salt's __func_alias__ at the time.

I looked at line 970 of osd.py and I find _part_probe.  What bug did you find?

I think the rest looks fine.  Everything in helper had to become public.  I also wonder if helper should move to utils now that Salt seems to have a semi official location.  (Not in this PR, just thinking out loud about future organization of code.)

jschmid1:
> I would like to keep the add and push aliases for filequeue. I did not know about Salt's func_alias at the time.

I also tried that.. It doesn't get exported I fear..
We really only use that in one place and 'enqueue' and 'dequeue' are not terrible imho.

> I looked at line 970 of osd.py and I find _part_probe. What bug did you find?

``` python

-                wipe_cmd = ("dd if=/dev/zero of={}{} bs=4096 count=1 "
-                            "oflag=direct".format(device, number))
-                _run(wipe_cmd)
+                prefix = ''
+                if device.startswith('/dev/nvme'):
+                    prefix = 'p'
+                wipe_cmd = ("dd if=/dev/zero of={}{}{} bs=4096 count=1 "
+                            "oflag=direct".format(device, prefix, number))
+                __salt__['helper.run'](wipe_cmd)
```

classic NVME partition name handling..

> I think the rest looks fine. Everything in helper had to become public. I also wonder if helper should move to utils now that Salt seems to have a semi official location. (Not in this PR, just thinking out loud about future organization of code.)

+1

swiftgist:
Removed some trailing spaces, squashed commit and rebased on master since we had a conflict with the one-liner in cephinspector.  Otherwise, this works for me on Salt 2017 too.

Issue #1019, opened 2018-03-16
When Stage 3 fails to deploy ceph-mgr, unrelated error message is displayed


smithfarm:
Due to what appears to be an upstream ceph-mgr bug [1] , Stage 3 fails to deploy MGRs. After running Stage 3 (which fails with the error below), all three MONs are running but no MGRs are running. The error message displayed by Stage 3 says nothing about the MGRs.

```
2018-03-16T01:21:02.990 INFO:teuthology.orchestra.run.target147135132045.stdout:----------
2018-03-16T01:21:02.990 INFO:teuthology.orchestra.run.target147135132045.stdout:          ID: check if osd processes are still running on target147135132045.teuthology after restarting osds
2018-03-16T01:21:02.990 INFO:teuthology.orchestra.run.target147135132045.stdout:    Function: salt.state
2018-03-16T01:21:02.991 INFO:teuthology.orchestra.run.target147135132045.stdout:      Result: False
2018-03-16T01:21:02.991 INFO:teuthology.orchestra.run.target147135132045.stdout:     Comment: Run failed on minions: target147135132045.teuthology
2018-03-16T01:21:02.991 INFO:teuthology.orchestra.run.target147135132045.stdout:              Failures:
2018-03-16T01:21:02.991 INFO:teuthology.orchestra.run.target147135132045.stdout:                  target147135132045.teuthology:
2018-03-16T01:21:02.991 INFO:teuthology.orchestra.run.target147135132045.stdout:                  ----------
2018-03-16T01:21:02.991 INFO:teuthology.orchestra.run.target147135132045.stdout:                            ID: wait for osd processes
2018-03-16T01:21:02.991 INFO:teuthology.orchestra.run.target147135132045.stdout:                      Function: module.run
2018-03-16T01:21:02.991 INFO:teuthology.orchestra.run.target147135132045.stdout:                          Name: cephprocesses.wait
2018-03-16T01:21:02.992 INFO:teuthology.orchestra.run.target147135132045.stdout:                        Result: False
2018-03-16T01:21:02.992 INFO:teuthology.orchestra.run.target147135132045.stdout:                       Comment: Module function cephprocesses.wait executed
2018-03-16T01:21:02.992 INFO:teuthology.orchestra.run.target147135132045.stdout:                       Started: 01:18:47.344292
2018-03-16T01:21:02.992 INFO:teuthology.orchestra.run.target147135132045.stdout:                      Duration: 135494.09 ms
[...]

PR #1053, opened 2018-04-04
Stage 3: fail immediately if MONs or MGRs do not stay up after starting

smithfarm:
`service.running` will start and enable the service, which is nice, but it reports the service as up and running even if it fails, e.g., 0.5 seconds after being started.
    
Add a small bash script that waits a short time and checks the service status, so we don't continue with a failed MGR.

Do the same with the MONs (and migrate them to `service.running`).

Also, restart the MGRs before the OSDs since the OSDs cannot start if MGRs are not running.

Fixes: https://github.com/SUSE/DeepSea/issues/1019

-----------------

### DIFF
--- a/srv/salt/ceph/mon/default.sls
+++ b/srv/salt/ceph/mon/default.sls
@@ -33,17 +33,9 @@ create_mon_fs:
         - file: /var/lib/ceph/tmp/keyring.mon
 
 
-# TODO: we can put this check in a nice exec. module
-start-mon:
-  cmd.run:
-    - name: "systemctl start ceph-mon@{{ grains['host'] }}"
+start mon:
+  service.running:
+    - name: ceph-mon@{{ grains['host'] }}
     - require:
-        - cmd: create_mon_fs
-
-enable-mon:
-  cmd.run:
-    - name: "systemctl enable ceph-mon@{{ grains['host'] }}"
-    - require:
-        - cmd: create_mon_fs
-
-
+      - cmd: create_mon_fs
+    - enable: True
--- a/srv/salt/ceph/mon/default.sls
+++ b/srv/salt/ceph/mon/default.sls
@@ -39,3 +39,15 @@ start mon:
     - require:
       - cmd: create_mon_fs
     - enable: True
+
+
+wait for mon:
+  module.run:
+    - name: cephprocesses.wait
+    - kwargs:
+        'timeout': 6
+        'delay': 2
+        'roles':
+          - mon
+    - fire_event: True
+    - failhard: True
--- a/srv/salt/ceph/mgr/default.sls
+++ b/srv/salt/ceph/mgr/default.sls
@@ -7,4 +7,13 @@ start mgr:
     - name: ceph-mgr@{{ grains['host'] }}
     - enable: True
 
-
+wait for mgr:
+  module.run:
+    - name: cephprocesses.wait
+    - kwargs:
+        'timeout': 6
+        'delay': 2
+        'roles':
+          - mgr
+    - fire_event: True
+    - failhard: True
--- a/srv/salt/ceph/stage/deploy/default.sls
+++ b/srv/salt/ceph/stage/deploy/default.sls
@@ -1,5 +1,5 @@
 include:
   - .core
   - ...restart.mon.lax
-  - ...restart.osd.lax
   - ...restart.mgr.lax
+  - ...restart.osd.lax


smithfarm:
Stage 3 now fails for the right reason:

```
2018-04-04T17:39:05.903 INFO:teuthology.orchestra.run.target149202184174.stdout:          ID: mgrs
2018-04-04T17:39:05.903 INFO:teuthology.orchestra.run.target149202184174.stdout:    Function: salt.state
2018-04-04T17:39:05.903 INFO:teuthology.orchestra.run.target149202184174.stdout:      Result: False
2018-04-04T17:39:05.903 INFO:teuthology.orchestra.run.target149202184174.stdout:     Comment: Run failed on minions: target149202184174.teuthology
2018-04-04T17:39:05.903 INFO:teuthology.orchestra.run.target149202184174.stdout:     Started: 17:38:53.360731
2018-04-04T17:39:05.903 INFO:teuthology.orchestra.run.target149202184174.stdout:    Duration: 12454.749 ms
2018-04-04T17:39:05.904 INFO:teuthology.orchestra.run.target149202184174.stdout:     Changes:
2018-04-04T17:39:05.904 INFO:teuthology.orchestra.run.target149202184174.stdout:              target149202184174.teuthology:
2018-04-04T17:39:05.904 INFO:teuthology.orchestra.run.target149202184174.stdout:                Name: /var/lib/ceph/mgr/ceph-target149202184174/keyring - Function: file.managed - Result: Clean Started: - 17:38:53.799325 Duration: 87.781 ms
2018-04-04T17:39:05.904 INFO:teuthology.orchestra.run.target149202184174.stdout:                Name: ceph-mgr@target149202184174 - Function: service.running - Result: Changed Started: - 17:39:00.377269 Duration: 390.896 ms
2018-04-04T17:39:05.904 INFO:teuthology.orchestra.run.target149202184174.stdout:              ----------
2018-04-04T17:39:05.904 INFO:teuthology.orchestra.run.target149202184174.stdout:                        ID: verify_mgr_running
2018-04-04T17:39:05.904 INFO:teuthology.orchestra.run.target149202184174.stdout:                  Function: cmd.run
2018-04-04T17:39:05.904 INFO:teuthology.orchestra.run.target149202184174.stdout:                      Name: sleep 5
2018-04-04T17:39:05.904 INFO:teuthology.orchestra.run.target149202184174.stdout:              systemctl status ceph-mgr@target149202184174
2018-04-04T17:39:05.905 INFO:teuthology.orchestra.run.target149202184174.stdout:              mgr_status=$?
2018-04-04T17:39:05.905 INFO:teuthology.orchestra.run.target149202184174.stdout:              test $mgr_status -eq 0 || echo "The ceph-mgr@target149202184174 unit failed to start"
2018-04-04T17:39:05.905 INFO:teuthology.orchestra.run.target149202184174.stdout:              test $mgr_status -eq 0
2018-04-04T17:39:05.905 INFO:teuthology.orchestra.run.target149202184174.stdout:
2018-04-04T17:39:05.905 INFO:teuthology.orchestra.run.target149202184174.stdout:                    Result: False
2018-04-04T17:39:05.905 INFO:teuthology.orchestra.run.target149202184174.stdout:                   Comment: Command "sleep 5
2018-04-04T17:39:05.905 INFO:teuthology.orchestra.run.target149202184174.stdout:                            systemctl status ceph-mgr@target149202184174
2018-04-04T17:39:05.905 INFO:teuthology.orchestra.run.target149202184174.stdout:                            mgr_status=$?
2018-04-04T17:39:05.905 INFO:teuthology.orchestra.run.target149202184174.stdout:                            test $mgr_status -eq 0 || echo "The ceph-mgr@target149202184174 unit failed to start"
2018-04-04T17:39:05.905 INFO:teuthology.orchestra.run.target149202184174.stdout:                            test $mgr_status -eq 0
2018-04-04T17:39:05.906 INFO:teuthology.orchestra.run.target149202184174.stdout:                            " run
2018-04-04T17:39:05.906 INFO:teuthology.orchestra.run.target149202184174.stdout:                   Started: 17:39:00.770902
2018-04-04T17:39:05.906 INFO:teuthology.orchestra.run.target149202184174.stdout:                  Duration: 5019.289 ms
2018-04-04T17:39:05.906 INFO:teuthology.orchestra.run.target149202184174.stdout:                   Changes:
2018-04-04T17:39:05.906 INFO:teuthology.orchestra.run.target149202184174.stdout:                            ----------
2018-04-04T17:39:05.906 INFO:teuthology.orchestra.run.target149202184174.stdout:                            pid:
2018-04-04T17:39:05.906 INFO:teuthology.orchestra.run.target149202184174.stdout:                                34312
2018-04-04T17:39:05.906 INFO:teuthology.orchestra.run.target149202184174.stdout:                            retcode:
2018-04-04T17:39:05.906 INFO:teuthology.orchestra.run.target149202184174.stdout:                                1
2018-04-04T17:39:05.906 INFO:teuthology.orchestra.run.target149202184174.stdout:                            stderr:
2018-04-04T17:39:05.907 INFO:teuthology.orchestra.run.target149202184174.stdout:                            stdout:
2018-04-04T17:39:05.907 INFO:teuthology.orchestra.run.target149202184174.stdout:                                * ceph-mgr@target149202184174.service - Ceph cluster manager daemon
2018-04-04T17:39:05.907 INFO:teuthology.orchestra.run.target149202184174.stdout:                                   Loaded: loaded (/usr/lib/systemd/system/ceph-mgr@.service; enabled; vendor preset: disabled)
2018-04-04T17:39:05.907 INFO:teuthology.orchestra.run.target149202184174.stdout:                                   Active: activating (auto-restart) (Result: exit-code) since Wed 2018-04-04 17:39:00 UTC; 5s ago
2018-04-04T17:39:05.907 INFO:teuthology.orchestra.run.target149202184174.stdout:                                 Main PID: 34277 (code=exited, status=1/FAILURE)
2018-04-04T17:39:05.907 INFO:teuthology.orchestra.run.target149202184174.stdout:
2018-04-04T17:39:05.907 INFO:teuthology.orchestra.run.target149202184174.stdout:                                Apr 04 17:39:00 target149202184174.teuthology systemd[1]: ceph-mgr@target149202184174.service: Main process exited, code=exited, status=1/FAILURE
2018-04-04T17:39:05.907 INFO:teuthology.orchestra.run.target149202184174.stdout:                                Apr 04 17:39:00 target149202184174.teuthology systemd[1]: ceph-mgr@target149202184174.service: Unit entered failed state.
2018-04-04T17:39:05.907 INFO:teuthology.orchestra.run.target149202184174.stdout:                                Apr 04 17:39:00 target149202184174.teuthology systemd[1]: ceph-mgr@target149202184174.service: Failed with result 'exit-code'.
2018-04-04T17:39:05.907 INFO:teuthology.orchestra.run.target149202184174.stdout:                                The ceph-mgr@target149202184174 unit failed to start
2018-04-04T17:39:05.908 INFO:teuthology.orchestra.run.target149202184174.stdout:
2018-04-04T17:39:05.908 INFO:teuthology.orchestra.run.target149202184174.stdout:              Summary for target149202184174.teuthology
2018-04-04T17:39:05.908 INFO:teuthology.orchestra.run.target149202184174.stdout:              ------------
2018-04-04T17:39:05.908 INFO:teuthology.orchestra.run.target149202184174.stdout:              Succeeded: 2 (changed=2)
2018-04-04T17:39:05.908 INFO:teuthology.orchestra.run.target149202184174.stdout:              Failed:    1
2018-04-04T17:39:05.908 INFO:teuthology.orchestra.run.target149202184174.stdout:              ------------
2018-04-04T17:39:05.908 INFO:teuthology.orchestra.run.target149202184174.stdout:              Total states run:     3
2018-04-04T17:39:05.908 INFO:teuthology.orchestra.run.target149202184174.stdout:              Total run time:   5.498 s
```

swiftgist:
I would not be fond of adding multiple sleeps that will happen forever more on reruns on Stage 3.  I am already concerned about the serial restart checks that increase the duration of Stages 3 and 4.

If service.running is lying, then I think it's okay to replace it with something that meets the requirements.  If the only concern is the waiting on the initial start that may be easier in python.  (Ultimately, the Salt call would behave as expected, but I think we can focus on meeting our needs until that happens.)  Is this the only place that it matters?  I suspect all "service.running" calls are equally fibbing.

There may be a possibility of using states with onfail, but I think that makes the sls files harder.  I've used it in a smoketest and that caused some confusion.

smithfarm:
On second thought, doesn't Salt start the MONs and MGRs in parallel here? In that case, the wait time added by this PR is as much as 3 times lower than I cited above.

smithfarm:
As for rewriting `service.running`, that might not be so easy, given that the semantics for checking that the service is really running may differ from daemon to daemon. 

Taking as an example "wait and check that the service is really running" code that we already have in master - https://github.com/SUSE/DeepSea/blob/master/srv/salt/ceph/monitoring/grafana/init.sls#L20-L46 - when we start grafana using `service.running`, we follow that with a wait loop that fires curl commands at 5 second intervals for up to 5 * 24 = 120 seconds (if I'm reading the code correctly).

jschmid1:
I'm also not too crazy about adding sleeps to the deployment states but on the other hand it simplifies debugging and points you to the failure more accurately.

If I thought that trough correctly we spent an additional:

(n_mon * 5) + (n_mgr * 5)

however, mons and managers are being deployed in parallel which sets n_mon and n_mgr to 1.

that leads to an extra 10seconds to stage.3 for every invocation independent of the cluster size.

Might have an impact for smaller clusters but will not be noticed in bigger deployments where restarts take up the majority of stage.3 (let alone fresh deployments)

smithfarm:
@tserong Nice idea.

The reason for this PR is that `systemctl start` returns success when it successfully starts the MGR process. But, at some point after starting, the MGR process aborts for whatever reason, so **effectively** the process failed to start, even though **technically** it (initially) succeeded. In the case when the MGR aborts before becoming fully operational, this usually happens within the first *n* seconds, where *n* is some small number, but 5 seconds is just arbitrary.

So I agree it would be better to test for "full operational-ness" in a wait loop, but I don't have that test ATM.

swiftgist:
Is the actual problem that Salt tries to start the mgr processes too soon after starting the monitor on a fresh deployment?

smithfarm:
I would say the actual problem is that DeepSea continues as if nothing happened when the MONs or MGRs fail to come up.

smithfarm:
Regarding what @tserong wrote, for the MONs there is a "ceph mon ping" command which could be used.

And I don't think it will be difficult to come up with an equivalent for the MGRs. That will be better than just waiting 5 seconds, which may or may not be enough.

srv/salt/ceph/mon/default.sls
```diff
@@ -33,17 +33,23 @@ create_mon_fs:
...
-    - name: "systemctl enable ceph-mon@{{ grains['host'] }}"
     - require:
-        - cmd: create_mon_fs
+      - service: mon-start
+    - name: |
+        sleep 5
+        systemctl status ceph-mon@{{ grains['host'] }}
+        mon_status=$?
+        test $mon_status -eq 0 || echo "The ceph-mon@{{ grains['host'] }} unit failed to start"
+        test $mon_status -eq 0
```
swiftgist: Rather than the hardcoded sleep, I think one of these will be better:

```
check mon:
  module.run:
    - name: cephprocesses.check
    - kwargs:
        'roles':
          - mon
    - fire_event: True
    - failhard: True
```

```

wait for mon:
  module.run:
    - name: cephprocesses.wait
    - kwargs:
        'timeout': 6
        'delay': 2
        'roles':
          - mon
    - fire_event: True
    - failhard: True
```

The only reason to  use the wait version is if we think these processes are slow to start.  But the original problem is saying that these start and fail immediately.  Then, the check will be enough.

swiftgist: This would be the same for mgr as well.  We are using the examples as part of the restart logic.  Take a look at /srv/salt/ceph/processes/`mon|mgr`/default.sls.

smithfarm: @swiftgist I made the change you suggested.

smithfarm:
Passed integration test at http://147.135.132.231:8081/ubuntu-2018-04-25_06:52:22-deepsea:basic:health-ok-mini-ses6---basic-openstack/

jschmid1:
looks good

smithfarm:
@swiftgist How is this looking to you?

swiftgist:
I didn't see the change from sleep to cephprocesses.wait.

smithfarm:
We got a report that this is missing in SES5 - opened https://github.com/SUSE/DeepSea/pull/1564 with the backport.

PR #1054, opened 2018-04-04
Support restart smoketests

swiftgist:
Initially working for mon, mds, mgr and rgw.  Each service is checked
for the following:

1) Forced restart must restart
2) No config changes must not restart
3) Config changes must restart
4) Config removal must restart

The global.conf, igw and ganesha are not addressed.  Test 2 will fail if your cluster does have pending changes.  I an undecided on further initializing or requiring that pending changes be applied.

### DIFF
--- a/srv/salt/ceph/tests/restart/setup.sls
+++ b/srv/salt/ceph/tests/restart/setup.sls
@@ -1,0 +1,6 @@
+
+update {{ service }}.conf:
+  cmd.run:
+    - name: "date +\\#\\ %c > /srv/salt/ceph/configuration/files/ceph.conf.d/{{ service }}.conf"
+
+
--- a/srv/salt/ceph/tests/restart/teardown.sls
+++ b/srv/salt/ceph/tests/restart/teardown.sls
@@ -1,0 +1,4 @@
+
+/srv/salt/ceph/configuration/files/ceph.conf.d/{{ service }}.conf:
+  file.absent
+
--- a/Makefile
+++ b/Makefile
@@ -63,11 +63,23 @@ copy-files:
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/tests/quiescent/timeout
 	install -m 644 srv/salt/ceph/tests/quiescent/*.sls $(DESTDIR)/srv/salt/ceph/tests/quiescent
 	install -m 644 srv/salt/ceph/tests/quiescent/timeout/*.sls $(DESTDIR)/srv/salt/ceph/tests/quiescent/timeout
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/tests/restart
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/tests/restart/mon
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/tests/restart/mds
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/tests/restart/mgr
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/tests/restart/rgw
+	install -m 644 srv/salt/ceph/tests/restart/*.sls $(DESTDIR)/srv/salt/ceph/tests/restart
+	install -m 644 srv/salt/ceph/tests/restart/mon/*.sls $(DESTDIR)/srv/salt/ceph/tests/restart/mon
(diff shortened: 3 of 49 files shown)


smithfarm:
I'm a little concerned about the number of files (dozens?) required for each new test case. Purely from a software maintenance perspective, it seems unwieldy.

swiftgist:
Part of the issue was trying to factor out the common steps.  Salt and Jinja includes cannot be mixed since Jinja evaluations will happen in the wrong order.  The original goal was to keep the three states so that these commands would work

```
salt 'nodeX' state.apply ceph.tests.restart.mds.force
salt 'nodeX' state.apply ceph.tests.restart.mds.nochange
salt 'nodeX' state.apply ceph.tests.restart.mds.change
```

This allows debugging at the smallest level.  The second issue is that ceph process names are inconsistent.  The rgw service needs its own versions of the sls files.  (This gets worse with ganesha and igw.  I did not complete these yet, but I can add them without having to change common.sls.)

For a given service, we have 9 files.  Three are related to the states so that an admin can easily call them.  The six others (save_pid, same_pid, changed_pid, setup, teardown, reset) are all an attempt to keep things DRY.  The only other choice is to save on six files at the top level .../tests/restart, but have duplicate commands in separate directories.  Passing Jinja variables via context is coming, but I do not know that it will simplify much.

Lastly, I do not think Ceph will be adding a large number of other services.  Adding a new service is `rsync -a mds/ new_service` and `sed -i 's!mds!new_service!' *.sls */init.sls`.  I plan on finishing ganesha and igw soon-ish.

srv/salt/ceph/tests/restart/setup.sls
```diff
@@ -0,0 +1,6 @@
+
+update {{ service }}.conf:
+  cmd.run:
+    - name: "date +\\#\\ %c > /srv/salt/ceph/configuration/files/ceph.conf.d/{{ service }}.conf"
```
tserong: This is destructive of any custom config that's already there, which is fine of course for automated testing, but might be irritating when running smoketests in random dev environments. Should we document somewhere what tests are destructive of?

srv/salt/ceph/tests/restart/teardown.sls
```diff
@@ -0,0 +1,4 @@
+
+/srv/salt/ceph/configuration/files/ceph.conf.d/{{ service }}.conf:
+  file.absent
```
tserong: Likewise.

jschmid1:
Am I using it wrong?

``` 
admin:~/deepsea # salt-run state.orch ceph.smoketests.restart.mon
[WARNING ] The function "module.run" is using its deprecated version and will expire in version "Sodium".
[ERROR   ] Unable to write to grains file at /etc/salt/grains. Check permissions.
[ERROR   ] {'out': 'highstate', 'ret': {'mon1.ceph': ["No matching sls found for 'ceph.tests.restart.mon.forced' in env 'base'"]}}
admin.ceph_master:
  Name: reset systemctl initially - Function: salt.state - Result: Changed Started: - 11:34:12.545333 Duration: 462.484 ms
  Name: grains.setval - Function: module.run - Result: Changed Started: - 11:34:13.008803 Duration: 5.496 ms
----------
          ID: check mon forced restart
    Function: salt.state
      Result: False
     Comment: Run failed on minions: mon1.ceph
     Started: 11:34:13.014500
    Duration: 383.742 ms
     Changes:   
              mon1.ceph:
                  Data failed to compile:
              ----------
                  No matching sls found for 'ceph.tests.restart.mon.forced' in env 'base'

Summary for admin.ceph_master
------------
Succeeded: 2 (changed=2)
Failed:    1
------------
Total states run:     3
Total run time: 851.722 ms
```

srv/salt/ceph/tests/restart/setup.sls
swiftgist: I have not decided where to put this officially.  Maybe in the README of the smoketests?  I think it's fair to state that these tests can be destructive, but will still try to restore the environment to some degree.  The target audience is developers and automated testing.

tserong: README of the smoketests sounds fine to me.

jschmid1:
> Is this where we need that try-restart of the Salt master?

Not sure, trying it again today

swiftgist:
rebased

PR #1056, opened 2018-04-04
Rescind ceph.conf from minions without roles

### LINKED ISSUE #559: rescind ceph.conf missing
It turns out we do not have a rescind for ceph.conf.  If a minion no longer belongs to the assigned cluster, Stage 5 should remove the configuration file.

swiftgist:

Fixes #559


jschmid1:
lgtm

PR #1061, opened 2018-04-06
Prefer underscore device names

### LINKED ISSUE #952: pillar data and proposals should use "user-friendlier" names for disks
When we have to deal with disks in a server for OSDs there are three different names available in /dev/disk/by-id - i.e.:

```
lrwxrwxrwx 1 root root    9 Feb 20 13:16 scsi-35000cca2443c7c80 -> ../../sdf
lrwxrwxrwx 1 root root    9 Feb 20 13:16 scsi-SHGST_HUS726040AL5210_N8H2870Z -> ../../sdf
lrwxrwxrwx 1 root root    9 Feb 20 13:16 wwn-0x5000cca2443c7c80 -> ../../sdf

```

At the moment it seems that the first one - in the above example "scsi-35000cca2443c7c80" is used - but that number is most often not very user friendly as it does not allow to identify the vendor or the disk serial number directly.

The other name "scsi-SHGST_HUS726040AL5210_N8H2870Z" would be much better as this has the disk type "HUS726040AL5210" and the serial number "N8H2870Z".

Would it be possible to adjust DeepSea to use the name "scsi-SHGST_HUS726040AL5210_N8H2870Z"?

swiftgist:
Many symlinks in /dev/disk/by-id match a single device, but the most
useful for the administrator has the best description.

Intentional code duplication to not conflate issues.

Fixes #952

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -1747,7 +1747,7 @@ class OSDDevices(object):
     # pylint: disable=no-self-use
     def _uuid_device(self, device, pathname="/dev/disk/by-id"):
         """
-        Return the uuid device, last one if multiple are matched
+        Return the uuid device, prefer the most descriptive
         """
         if os.path.exists(device):
             if os.path.exists(pathname):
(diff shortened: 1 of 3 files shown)


swiftgist:
See #1062 about the code duplication.  I expect that this PR would be backported as is to SES5, but reorganization would not.

swiftgist:
@jschmid1 We would be moving twice though and that  was part of my reason for leaving it as a separate issue?  Or are you thinking for SES5, it can live in helper.py?

jschmid1:
> @jschmid1 We would be moving twice though and that was part of my reason for leaving it as a separate issue? Or are you thinking for SES5, it can live in helper.py?

Well, there is no helper.py in SES5. But I'd reason with fixing things properly in master and _then_ care about backporting.

jschmid1:
Approving under the premise of a follow-up fix with /_utils etc.

jschmid1:
@swiftgist 
```
17:39 <jschmid> I think there is an issue with #1061
17:40 <jschmid> def _device in proposal should be under the scope of the Proposal class
17:40 <jschmid> but it's not globally avalable
17:40 <jschmid> but the class is referencing it to be in the class
17:40 <jschmid> by accessing it with self._device(arg)
```

the diff proves it

``` python
 22 <<<<<<< HEAD
 21     def _device(self, drive):
 20         """
 19         Default to Device File value.  Use by-id if available.
 18         """
 17         if 'Device Files' in drive:
 16             for path in drive['Device Files'].split(', '):
 15                 if 'by-id' in path:
 14                     return path
 13         # fallback to Device File if no by-id path was found
 12         return drive['Device File']
 11 =======
 10 
  9 def _device(drive):
  8     """
  7     Default to Device File value.  Prefer most descriptive.
  6     """
  5     if 'Device Files' in drive:
  4         devices = drive['Device Files'].split(', ')
  3         index = _prefer_underscores(devices)
  2         # Prefer 'Device File' over last item
  1         if index > -1:
217             return devices[index]
  1     return drive['Device File']
  2 >>>>>>> a0168a100... Prefer underscore device names


```

PR #1074, opened 2018-04-11
Validate salt version

### LINKED ISSUE #839: Validate salt version when running stages
DeepSea is known to work with only certain versions (or just one certain version??) of Salt. As e.g. #798 demonstrates, running DeepSea with a "not-known-to-work" version of Salt can lead to weird errors, and without special knowledge of Salt/DeepSea it's impossible to know the root cause.

DeepSea has a mechanism for defining validation checks [1] that run before a stage's orchestration files are triggered.

If a "Salt version" validation check were added, it would be immediately apparent in such cases what the problem is.

[1] https://github.com/SUSE/DeepSea/blob/master/srv/modules/runners/validate.py

alexandergraul:
Fixes #839 

Description:
Add a validation for salt versions, failing the check results in a warning only.

Additionally, I refactored Validate.py to move collection of grain and pillar data inside Validate(). As a result, only passing a bool is required when instantiating Validate().


swiftgist:
I like the refactor and better variable names.  (The `grains` variable came a while after the `data` variable, but bugged me.)  During backports, we found the incompatibility between 2018 and 2016.  If it really is only the `tgt_type` vs. `expr_form` in local.cmd, I would like to keep 2016 as the minimal Salt version.  (IIRC, I think there's enough differences with 2015.8 that the effort isn't worth it for that version.)  I would like DeepSea to not require all users to be bleeding edge.  It's generally annoying when DeepSea is simply an application of Salt.

As far as having a top limit, I do not know that we should.  We had issues since we had python imports vs. Salt module calls, but I think the expectation is that we would work with the latest Salt.  Is the general idea that the admin may be ahead of the developers in trying the latest Salt?  The warning for this says "Sure, you can go ahead, but we haven't yet".  That's probably not too bad, but it will be something we would have to keep updated.

jschmid1:
also a LGTM from me..

> I would like to keep 2016 as the minimal Salt version.

We need to verify first if that still works, though :) I know that 2017 works, but haven't tried 2016..

> That's probably not too bad, but it will be something we would have to keep updated.

The initial idea is to keep a 'matix-text-file' where we maintain a map of version we support. That might be easier to approach from the outside aswell.

Issue #1004, opened 2018-03-13
salt jobs, jinja and root


swiftgist:
### Description of Issue/Question

The Jinja conditionals are executed as root within a runner.  The side effect is creating a job in /var/cache/salt/master/jobs/... as root.

The second part is that Salt pretends to be cron and run a "clean_old_jobs" function.  SUSE runs the salt-master as user 'suse'.  Permission denied causes python exception.

The second side effect is that clean_old_jobs never succeeds since it will die every time it is run.  This results in the complaints I have heard from the field about /var filling.

The long solution will likely involve rewriting these runners differently, but I do not know how at the moment.

One short workaround is to use cron.  This will not remove the exception, but will address the problem in the field.  One example would be

15 20 * * * /usr/bin/find /var/cache/salt/master/jobs/ -type f -mtime +1 | xargs rm 2>&1 | logger -t salt-jobs

This would remove all files older than one day every day at 8:15 pm.  Output would be submitted to syslog.

swiftgist:
Related to #935

swiftgist:
Closing with #1078

PR #1078, opened 2018-04-11
Correct permissions on job cache

### LINKED ISSUE #935: Runners, root and jobs
Since SUSE runs the salt-master as the user 'salt', we have a unique situation.  Any jinja that executes a runner on the Salt master, runs as root since the minion does.  (I was surprised.)

The unfortunate side effect is that the jobs cache ends up with files owned by root.

The Salt master runs a clean up job that raises a permission denied exception and stops.  The jobs begin to accumulate until the /var filesystem is filled.

While a cron job will workaround this for now, rethinking runners and Jinja is the likely a better path.

swiftgist:
Any state file including Jinja templates that call a salt runner
indirectly creates a Salt job owned by root.  Dropping privileges in the
runner will break the state.  This kludge will keep the Salt cron
equivalent from crashing with a permissions exception.

Fixes #935, #1004

### DIFF
--- a/srv/salt/ceph/configuration/create/default.sls
+++ b/srv/salt/ceph/configuration/create/default.sls
@@ -12,6 +12,11 @@ removing minion cache:
     - makedirs: True
     - fire_event: True
 
-
-
+/var/cache/salt/master/jobs:
+  file.directory:
+    - user: {{ salt['deepsea.user']() }}
+    - group: {{ salt['deepsea.group']() }}
+    - recurse:
+      - user
+      - group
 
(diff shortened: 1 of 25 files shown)


swiftgist:
Right, the only guaranteed way to get an included file to run at the end (without resorting to requires or order numbers which is just as cumbersome) is to turn all of these existing states into subfiles which get included as well.

I'm hoping that one of us will come up with some brilliant alternative to querying the pillar from the master without the side effect of running as root.

smithfarm:
Looks like this solution was replaced by https://github.com/SUSE/DeepSea/pull/1175, which has already been backported, so removing "needs backport".

PR #1083, opened 2018-04-13
[backport][SES5] qa: health-ok: fix --mini option

smithfarm:
Commit ea7af31f0 added a --mini option to health-ok.sh, but it was
not possible to use because it was not added to the getopt command.


jschmid1:
backport of #1082

PR #1084, opened 2018-04-16
runners/populate: don't create public_address proposals for mon/igw

### LINKED ISSUE #772: populate.proposal creates unnecessary files
For roles mon an igw the populate runner still creates yml files with public_address settings in them.

jan--f:

Fixes #772


swiftgist:
For anyone finding this, the original method before the 0.7 series added static addresses for the mon and igw to the Salt pillar.  Overriding these values caused many headaches.  These addresses are now generated dynamically via Salt module `public.address` and that static entries are obsolete.

PR #1086, opened 2018-04-16
[RFC] [DNM] Shrink MDS cluster and restart on updates

jan--f:
This is a basic implementation of the [update routine recommended by upstream](http://docs.ceph.com/docs/master/cephfs/upgrading/).
This is not integrated into any stages yet. We need to add a conditional to decide if a sequential restart is ok (e.g. on config changes) or if the shrink-restart routine is needed (any time the mds package is updated).

The code can be tried by starting a cluster, updating the binaries manually and then run ```salt-run state.orch ceph.restart.mds.default-parallel```.

The naming is also not final yet. This RFC PR is against SES5 since it originally way meant to be for the update to or over 12.2.4. It has since turned out that this procedure is needed for all MDS updates until further notice. I.e. master will need a PR as well.

### DIFF
--- a/srv/modules/runners/cmd.py
+++ b/srv/modules/runners/cmd.py
@@ -1,0 +1,15 @@
+import salt.client
+
+from subprocess import check_output
+
+def run(**kwargs):
+    args = {'cmd': 'echo no command supplied'}
+    args.update(kwargs)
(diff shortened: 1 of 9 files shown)


jan--f:
Closing since this is superseded by #1117. Once this is merged I will backport.

Issue #1093, opened 2018-04-17
deepsea can not create OSD but does not throw an error


Martin-Weiss:
### Description of Issue/Question
We deploy a new cluster and we get 7 out of 8 configured OSDs created.
One fails continuously to be created but stage.3 does not show any error!
We see that with every rerun of stage.3 two additional partitions are created on the NVMe (one for the WAL and one for the RocksDB of the OSD that fails to be created). 

### Setup
4 OSD node + 1 admin cluster , super micro hardware, NVMe and spinner Bluestore setup.
32 OSDs - each OSD node has 2 NVMe - ratio 1:4

### Steps to Reproduce Issue
run / re-run stage.3

### Versions Report
(Provided by running:
`salt-run deepsea.version` 

0.8.2+git.0.6b39c2648

`rpm -qi salt-minion`

dd-ses-adm:~ # rpm -qi salt-minion
Name        : salt-minion
Version     : 2016.11.4
Release     : 46.17.1
Architecture: x86_64
Install Date: Tue Apr 17 13:59:37 2018
Group       : System/Management
Size        : 38128
License     : Apache-2.0
Signature   : RSA/SHA256, Fri Feb  9 15:06:01 2018, Key ID 70af9e8139db7c82
Source RPM  : salt-2016.11.4-46.17.1.src.rpm
Build Date  : Fri Feb  9 15:03:24 2018
Build Host  : sheep26
Relocations : (not relocatable)
Packager    : https://www.suse.com/
Vendor      : SUSE LLC <https://www.suse.com/>
URL         : http://saltstack.org/
Summary     : The client component for Saltstack
Description :
Salt minion is queried and controlled from the master.
Listens to the salt master and execute the commands.
Distribution: SUSE Linux Enterprise 12

`rpm -qi salt-master`

dd-ses-adm:~ # rpm -qi salt-master
Name        : salt-master
Version     : 2016.11.4
Release     : 46.17.1
Architecture: x86_64
Install Date: Tue Apr 17 13:57:40 2018
Group       : System/Management
Size        : 1662872
License     : Apache-2.0
Signature   : RSA/SHA256, Fri Feb  9 15:06:01 2018, Key ID 70af9e8139db7c82
Source RPM  : salt-2016.11.4-46.17.1.src.rpm
Build Date  : Fri Feb  9 15:03:24 2018
Build Host  : sheep26
Relocations : (not relocatable)
Packager    : https://www.suse.com/
[...]

PR #1096, opened 2018-04-18
Fail OSD deployment early in case of error

### LINKED ISSUE #984: set_bytes exception
With a misconfigured cluster (i.e. drives configured in the pillar that do not exist on the minion), the following error

Exception: '<' not supported between instances of 'NoneType' and 'int'

will be the result.  Not terribly helpful.  

The simple fix is adding a 'return 0'

```
            for disk in disks[__grains__['id']]:
                log.debug("Device File: {}".format(disk['Device File']))
                log.debug("device: {}".format(self.device))
                if disk['Device File'] == self.device:
                    return int(disk['Bytes'])
            return 0
```

This may just run into another error though.  We should skip these gracefully and log that they do not exist.

### LINKED ISSUE #1076: In Stage 3, error in "ceph-disk activate" passes unnoticed

When re-deploying a cluster with unzapped bluestore OSD disks, Stage 3 fails with an obscure error ("cephprocesses.wait"). Upon closer investigation, it is found that "ceph-disk activate" returned status code 1 on all disks.

It would be much nicer if the error message reflected the real cause of the failure. I'm marking it as a bug because the error is reported incorrectly.

### Setup

DeepSea master branch

jschmid1:

Fixes #1076 
Fixes #984 

Description:

If a OSD fails to deploy you will not get past stage.3 as the restart logic will stop you. This patch stops the deployment earlier and therefore saves the user time ($timeout is in the restarts is omitted)

Example error message:

```

                        ID: deploy OSDs
                  Function: module.run
                      Name: osd.deploy
                    Result: False
                   Comment: Module function osd.deploy executed
                   Started: 17:12:07.961774
                  Duration: 3284.045 ms
                   Changes:   
                            ----------
                            ret:
                                ----------
                                msg:
                                    One or more OSDs failed to deploy. Please see the minion logs
                                result:
                                    False

              Summary for data4.ceph
              ------------
              Succeeded: 2 (changed=1)
              Failed:    1
              ------------
              Total states run:     3
              Total run time:   3.349 s
```

-----------------

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -583,6 +583,7 @@ class OSDConfig(object):
             for disk in disks[__grains__['id']]:
                 if disk['Device File'] == self.device:
                     return int(disk['Bytes'])
+            return 0
         else:
             error = "Mine on {} for cephdisks.list".format(__grains__['id'])
             log.error(error)
@@ -1884,15 +1885,36 @@ def deploy():
     The last idea is converting all of this into a state module that returns
     all the commands in the comment.
     """
+    ret_data = {}
+    failed_run = False
     for device in configured():
         if not is_prepared(device):
+            ret_data[device] = {}
             config = OSDConfig(device)
             osdp = OSDPartitions(config)
             osdp.clean()
             osdp.partition()
             osdc = OSDCommands(config)
-            __salt__['helper.run'](osdc.prepare())
-            __salt__['helper.run'](osdc.activate())
+            ret_data[device]['prepare'] = __salt__['helper.run'](osdc.prepare())
+            ret_data[device]['activate'] = __salt__['helper.run'](osdc.activate())
+    for device, return_map in ret_data.items():
+        for func_name, run_data in return_map.items():
+            if run_data[0] != 0:
+                failed_run = True
+                msg = ("""
+                       Error encountered in {} function for device {}.
+                       Return_code: {}
+                       Stdout: {}
+                       Stderr: {}
+                       """.format(func_name,
+                                  device,
+                                  run_data[0],
+                                  run_data[1],
+                                  run_data[2]))
+                log.error(msg)
+    if failed_run:
+        return {'msg': 'One or more OSDs failed to deploy. Please see the minion logs',
+                'result': False}
 
 
 def redeploy(simultaneous=False, **kwargs):


srv/salt/_modules/osd.py
```diff
@@ -1884,15 +1885,36 @@ def deploy():
...
+                       Stdout: {}
+                       Stderr: {}
+                       """.format(func_name,
+                                  device,
+                                  run_data[0],
+                                  run_data[1],
+                                  run_data[2]))
+                log.error(msg)
+    if failed_run:
+        return {'msg': 'One or more OSDs failed to deploy. Please see the minion logs',
```
jschmid1: I probably could be precise in the counting :>

swiftgist:
If I have 24 drives and the second one fails, does this not force the admin to deal with the second drive before continuing?  In the current behavior, all could succeed except the second drive.

jschmid1:
> If I have 24 drives and the second one fails, does this not force the admin to deal with the second drive before continuing? In the current behavior, all could succeed except the second drive.

Yes, that's on purpose. I think it's more convenient for the user to get all the drives configured instead of being stopped at 2/24, fixing the error and running the stages over and over again. 
With this 'error-collecting' approach we basically do what we can and then bail out if an error occured.

Martin-Weiss:
> I think it's more convenient for the user to get all the drives configured instead of being stopped at 2/24,

I believe this really depends... 

In small clusters where it is unlikely to have "dead on arrival" disks this is the right approach IMO, too - especially due to issue #1093 where during each re-run of stage.3 unused and orphaned partitions will be created that need to be removed manually..

However - in large clusters - think of 100 nodes with 2400 OSDs - and where we have maybe 50 disks "dead on arrival" - stopping and restarting stage.3 50 times might not be the best approach.

swiftgist:
I think Martin's comment is fair and that "it depends".  We don't have trivial way to allow an admin to pick one strategy or the other.  I think the bigger issue is we still don't have a nice way of presenting failures back up the stack.  I think some of this starts to hit on how Salt works and that we have "terse" as the default setting.  And not to get carried away, but allowing one site to use both strategies for different minions would be the most flexible.  (Think of one troublesome server where any working is okay vs. the rest of the cluster that the admin wants strict failures.)

jschmid1:
ok, let's close this for now. We can still revive it later.

Issue #1062, opened 2018-04-06
Using __utils__ 


swiftgist:
We have various Salt modules that do similar things.  PR #1061 has intentional code duplication because the resolution could use a discussion.

DeepSea originally started against Salt 2015.8.  The `__utils__` did not exist.  In the past year, helper.py came into existence, but we were using the python import methods and not Salt.  We are changing this with #1051.

My inclination is to use the Salt way or naming.  

***
The next issue is that the function _prefer_underscores seems like an obvious choice since it is needed by osd.py and proposal.py.  However, this gets a little more fun.  The _uuid_device in osd.py and the get_uuid in fs.py need to be the same.  (The _uuid_device calls the _prefer_underscores function.)

The `__utils__` works the same as `__salt__` except looking in /srv/salt/_utils instead of /srv/salt/_modules.  This means that all calls will have two part names: filename and function.  Now, we cannot use the same filenames as /usr/lib/python3.6/site-packages/salt/utils.  (I have not verified, but I expect the behavior to match _modules.)  

I am undecided on naming and organizing.  I prefer filenames with meaning (e.g. fs.py is better than misc.py).  We also need to worry about future name collisions should Saltstack ever accept a fs.py.  These utilities are intended for developers writing other Salt modules.  We intentionally use _deepsea_ sparingly.  (It's a package and there's *deepsea_minions*.)  One could argue that *ceph* is overused.

Should we worry about /srv/salt/_utils becoming as cluttered as /srv/salt/_modules?  Is there a great argument to divide functions into several files?  What if a developer is reading code a year from now, is `deepsea.readlink` better than `fs.readlink` because the latter could be confused with a regular Salt module?

I will follow up #1061 with another PR once we have had some consensus.

PR #1101, opened 2018-04-18
Make master_minion optional

### LINKED ISSUE #547: Setting master_minion
From #538, we should be able to get rid of setting the master_minion altogether.  One of those "should've done this from the start" things.

The original problem is that some runners and the orchestration need to call a module on the master.  I do not have a way to do a salt-call from an orchestration.  That would be another way.

Relying on master_minion in the pillar has been solid.  No surprises there.  Getting it into the pillar from `__opt__` is doable too.  We need to think about the best answer here.  Do we do an external pillar approach or use a runner to update the pillar and refresh?

swiftgist:
The master_minion setting was originally set statically in the Salt
pillar.  By using a master module (not a minion module), this value will
use /etc/salt/minion_id and stay in sync.  Continue to support
overriding with a master_minion pillar variable.

Fixes #547

### DIFF
--- a/srv/modules/runners/populate.py
+++ b/srv/modules/runners/populate.py
@@ -28,11 +28,11 @@ import salt.key
 import salt.config
 import salt.utils
 import salt.utils.minions
+import salt.loader
 # pylint: disable=relative-import
 import ready
 
 import re
-import pprint
 import string
 import random
 import yaml
@@ -49,6 +49,7 @@ import logging
 # pylint: disable=relative-import
 from deepsea_minions import DeepseaMinions
 import operator
+import pprint
 
 import sys
 import six
@@ -697,17 +698,10 @@ class CephRoles(object):
 
         # second step, find cluster network by checking which network salt-master does not belong
         master_addrs = []
-        master_minion = None
-        cmd_result = local.cmd(self.search, 'pillar.get', ['master_minion'], tgt_type="compound")
-        for _, value in cmd_result.items():
-            master_minion = value
-            break
-        if not master_minion:
-            raise Exception("No master_minion found in pillar")
-        cmd_result = local.cmd(master_minion, 'grains.get', ['ipv4'], tgt_type="compound")
-        for _, addr_list in cmd_result.items():
-            master_addrs.extend([ipaddress.ip_address(u'{}'.format(addr))
-                                for addr in addr_list if not addr.startswith('127.')])
+        __opts__ = salt.config.minion_config('/etc/salt/minion')
+        __grains__ = salt.loader.grains(__opts__)
+        master_addrs.extend([ipaddress.ip_address(u'{}'.format(addr))
+                            for addr in __grains__['ipv4'] if not addr.startswith('127.')])
         for _, network in priorities:
             if network not in networks:
                 continue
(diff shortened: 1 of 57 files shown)


swiftgist:
When looking at the change in calls to the dunder set, note that the two `__opts__`  are not the same.  It matters.  I'm keeping the Salt convention until we have time to discuss and refactor.  The populate.py runner already needs to be refactored and we have talks of using __utils__ #1062.

Otherwise, the one moving part of a *master* module will give you a headache. :)  From Jinja, modules in /srv/salt/_modules are not available.  Only modules in /srv/modules/modules are.  The side effect is that `salt 'admin*' master.minion` or `salt-call master.minion` will return 'not available'.

While we could create another in /srv/salt/_modules or even symlink it, I believe that would just add to the confusion.  Also, I purposely removed a check when I realized that only the master can load these.

A side note is that this may lead to a better solution for #1078 by turning runners into master modules.

srv/modules/runners/populate.py
```diff
@@ -49,6 +50,7 @@
 # pylint: disable=relative-import
 from deepsea_minions import DeepseaMinions
 import operator
+import pprint
```
jan--f: debug leftover?

swiftgist: Removed. :)

swiftgist:
Removed master_minion.sls and entry from /srv/pillar/ceph/init.sls.

srv/modules/runners/populate.py
```diff
@@ -697,17 +698,10 @@ def public_cluster(self, networks):
...
-        for _, value in cmd_result.items():
-            master_minion = value
-            break
-        if not master_minion:
-            raise Exception("No master_minion found in pillar")
-        cmd_result = local.cmd(master_minion, 'grains.get', ['ipv4'], tgt_type="compound")
-        for _, addr_list in cmd_result.items():
-            master_addrs.extend([ipaddress.ip_address(u'{}'.format(addr))
-                                for addr in addr_list if not addr.startswith('127.')])
+        __opts__ = salt.config.minion_config('/etc/salt/minion')
```
tserong: This is reading `/etc/salt/minion`...

srv/modules/runners/populate.py
```diff
@@ -828,8 +822,14 @@ def _global(self):
         if not os.path.isdir(stack_dir):
             _create_dirs(stack_dir, self.root_dir)
         filename = "{}/global.yml".format(stack_dir)
+
+        __opts__ = salt.config.client_config('/etc/salt/master')
```
tserong: ...and this is reading `/etc/salt/master`

srv/modules/runners/populate.py
swiftgist: This is the assignment you normally expect when you need grains or modules on the minions alone.  I think this could be considered the default for the number of examples out there.

srv/modules/runners/populate.py
swiftgist: Check the following lines which I pulled directly from https://docs.saltstack.com/en/latest/ref/clients/.  The short answer is that if I need access to modules from /srv/modules/modules, those are only available when using the master config.  I had to figure this out backwards after getting Jinja to work in normal state files.

That's why I think this different `__opts__` is worthy of discussion and another refactor.  It's unfortunate that Salt examples use `__opts__` for two different contexts.  If we are centralizing pieces like this in `__utils__`, I think we need precise names.

The other reason is the switch to the dunder calls.  Originally, I used the simplest examples of local.cmd.  IIRC, I found the Salt scopes confusing and likely could not get queries to work.

swiftgist:
Updated validate check.

PR #1106, opened 2018-04-23
add 'rbd' kernel module validation

### LINKED ISSUE #723: Add Stage 4 kernel validation check
As https://bugzilla.suse.com/show_bug.cgi?id=1062120 illustrates, Stage 0 will complete successfully on a system that has "kernel-default-base" instead of "kernel-default", even though the former does not include the Ceph kernel client modules.

The kernel client modules are required for successful deployment, at least of the `igw` role. Not sure about other roles. Stage 4 fails for the `igw` role because lrbd tries to do `modprobe rbd` when it starts. 

One option would be to add a kernel validation check to Stage 4, which would fail if running on SUSE and the running kernel package is not `kernel-default`, or, more generally, if "modprobe rbd" fails.

(Though `kernel-default-base` is not expected to be used on production systems, the scenario in question *can* happen because it's shipped on the SLES and Leap "JeOS" images.)

alexandergraul:
Fixes #723 

Description:
call `modprobe` in stage 4 (services) to check if rbd is available as a kernel module.

-----------------

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -953,6 +953,23 @@ class Validate(Preparation):
                    "See `/srv/pillar/ceph/deepsea_minions.sls` for details")
             self.errors['deepsea_minions'] = [msg]
 
+    def kernel(self):
+        """
+        Verify that target_core_rbd kernel module is available on iSCSI Gateways and admin node
+        """
+        targets = [node for node in self.data if node == 'admin' or
+                   ('roles' in self.data[node] and 'igw' in self.data[node]['roles'])]
+        check = self.local.cmd(targets, 'kmod.check_available', ['target_core_rbd'],
+                               tgt_type='list')
+        for node, passed in check.items():
+            if not passed:
+                self.errors.setdefault('kernel_module', []).append(
+                        "{}: kernel module not active".format(node))
+
+        if 'kernel_module' not in self.errors:
+            self.passed['kernel_module'] = 'valid'
+        self._set_pass_status('kernel_module')
+
     def report(self):
         """
         Print the validation report
@@ -1092,6 +1109,7 @@ def deploy(**kwargs):
     valid = Validate("deploy", search_pillar=True, search_grains=True,
                      printer=printer)
     valid.openattic()
+    valid.kernel()
     valid.report()
 
     if valid.errors:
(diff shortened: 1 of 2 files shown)


swiftgist:
Are you checking for the rbd kernel module for block devices in general or specifically iSCSI?  This check will likely be more complicated since the validate runner is only checking on the master.  It's possible that the master and other minions are not on the same patch levels.  The command could succeed on the master, give a positive validation message and then fail on the minion.

If this is intended for a role (e.g. igw), then the check could be a state file calling a module and added to ceph.stage.iscsi. (While we could call this from Stage 0, we may be doing an unnecessary check for those sites not using iSCSI.)  If this is for a client, then the purpose is similar but we do not have a well defined role yet.  (We have started the idea, so the right role needs to be created. That's fine too.)

swiftgist:
@alexandergraul which direction did you want to go with this PR?

kshtsk:
retest this please

alexandergraul:
Back in April when you first pointed out that it would only run on master and hence not be conclusive, I've talked to Joshua and if I recall correctly we decided not to pursue this further because more fundamental changes to iSCSI handling were/are planned. I might remember wrong.

I just took a look at checking for iSCSI specifically and could not find anything (most documentation that I could find expects the kernel to just have support). If anyone knows a good way to do this check I can work on it, otherwise I would just close the PR.

ddiss:
For our current iSCSI implementation it's just a matter of checking for the target_core_rbd kernel module, in addition to the rbd.ko check that you already have here.

smithfarm:
I think checking for those modules on the igw node(s) is a good idea.

swiftgist:
Following @ddiss' comment, checking for the presence of the two files can be done in a Salt file.  If you need help with that, let me know.

alexandergraul:
This implementation adds checks if the kernel module is available with salt's `kmod.check_available` execution module and is part of `validate.deploy`

jschmid1:
jenkins says:

`Found coredumps on ubuntu@target054037030087.teuthology` 

That's kinda weird :) otherwise it seems to be fine.

kshtsk:
@jschmid1 This issue happened sometime, and there are coredump occur periodically in ganesha tests. Schedule another run, it can disappear.

jschmid1:
@susebot run teuthology

jschmid1:
@suesbot infrastructure failure again. @kshtsk Can we get those failures added to the list of failures that would restart a job?

Implementation seems to work fine. We just need to keep a close eye on what will be the status quo for iscsi.

Issue #1107, opened 2018-04-23
Stage 0 fails ("repo" and "common packages")


smithfarm:
### Description of Issue/Question

In a wip branch based on master 543b06a754a1f2832b552685af0dbf591aabe133 Stage 0 fails when run via the health-ok.sh integration test script:

```
2018-04-23T16:08:11.682 INFO:teuthology.orchestra.run.target147135134182.stdout:No minions matched the target. No command was sent, no jid was assigned.
2018-04-23T16:08:11.682 INFO:teuthology.orchestra.run.target147135134182.stdout:deepsea_minions          : [1m[92mvalid[0m
2018-04-23T16:08:11.683 INFO:teuthology.orchestra.run.target147135134182.stdout:master_minion            : [1m[92mvalid[0m
2018-04-23T16:08:11.683 INFO:teuthology.orchestra.run.target147135134182.stdout:ceph_version             : [1m[92mvalid[0m
2018-04-23T16:08:11.683 INFO:teuthology.orchestra.run.target147135134182.stdout:salt_version             : [1m[92mvalid[0m
2018-04-23T16:08:11.683 INFO:teuthology.orchestra.run.target147135134182.stdout:No minions matched the target. No command was sent, no jid was assigned.
2018-04-23T16:08:11.683 INFO:teuthology.orchestra.run.target147135134182.stdout:No minions matched the target. No command was sent, no jid was assigned.
2018-04-23T16:08:11.684 INFO:teuthology.orchestra.run.target147135134182.stdout:target147135134182.teuthology_master:
2018-04-23T16:08:11.684 INFO:teuthology.orchestra.run.target147135134182.stdout:  Name: sync master - Function: salt.state - Result: Changed Started: - 16:03:50.619388 Duration: 1465.724 ms
2018-04-23T16:08:11.684 INFO:teuthology.orchestra.run.target147135134182.stdout:  Name: salt-api - Function: salt.state - Result: Changed Started: - 16:03:52.085354 Duration: 11370.346 ms
2018-04-23T16:08:11.684 INFO:teuthology.orchestra.run.target147135134182.stdout:  Name: repo master - Function: salt.state - Result: Clean Started: - 16:04:03.456110 Duration: 461.272 ms
2018-04-23T16:08:11.684 INFO:teuthology.orchestra.run.target147135134182.stdout:  Name: prepare master - Function: salt.state - Result: Changed Started: - 16:04:03.917698 Duration: 238137.924 ms
[...]

PR #1108, opened 2018-04-23
Correct master_minion check

swiftgist:
Calling master modules in runners evaluated in Jinja causes states to
fail in later orchestrations.


smithfarm:
This fixes https://github.com/SUSE/DeepSea/issues/1107 for me.

PR #1109, opened 2018-04-23
Fix ceph_version

### LINKED ISSUE #1104: ceph_version takes 56 seconds
While working on the #1101, I found that `salt-run validate.setup` took about 64 seconds consistently.  I tracked this down the the ceph_version call.  

We call this validation for both Stage 0 and Stage 1.  The wait time increases with the number of minions.

swiftgist:
The pkg.latest_version does not exist and pkg.info_installed was called
serially resulting in 64 second setup runs.  With pkg.info_installed and
pkg.info_available called once, the time drops to 18 seconds for
existing clusters and 38 seconds for a fresh install.  All tests were on
a sixteen node cluster.

Fixes #1104

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -235,6 +235,10 @@ class Validate(Preparation):
             self.search = search
         # self._minion_check()
 
+        # Ceph version
+        self.package = 'ceph-common'
+        self.uninstalled = []
+
     def __get_items(self, enabled, target):
         """
         Look up [pillar|grains].items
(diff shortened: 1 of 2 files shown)


tests/unit/runners/test_validate.py
```diff
@@ -241,6 +241,131 @@ def test_storage_too_few(self, mock_localclient, mock_deepsea):
...
+        fake_data = { 'admin.ceph': '2018.1.99',
+                      'data.ceph': '2018.1.99'}
+
+        local = mock_localclient.return_value
+        local.cmd.return_value = fake_data
+        validator = validate.Validate("setup", search_grains=True)
+
+        assert len(validator.passed) == 0
+        validator.salt_version()
+        assert validator.passed['salt_version'] == 'valid'
```
smithfarm: could we add some tests for real-life version strings? For example, as per the commit message in 97bd8dd0765d854e1505648da1b1de36260a5f8e the version string might be " Zypper command failure: Repository 'SLE-Module-Server-Applications15-Updates' is invalid.\n[Server_Applications_Module_15_x86_64:SLE-Module-Server-Applications15-Updates|https://updates.suse.com/SUSE/Updates/SLE-Module-Server-Applications/15/x86_64/update?12G6EXPLytchxzDtQtrGRUsbLPHhWuDTz1o4dmfmH1MWnPq9oaPxPfuF1GDhPkRpznZvhxzOOI1Lt_LbJdQUy4yGwg75T_cW0WxOycIFoXCObDXyFHhwG7kr7AC27DwZV2gpKwz-ezerz3GbqCboZbpywl-NnXRSKHoi] Valid metadata not found at specified URL\nPlease check if the URIs defined for this repository are pointing to a valid repository.\nSkipping repository 'SLE-Module-Server-Applications15"

Downstream ceph RPM packages have version strings like "13.0.2.989+g371e07763c-lp150"

srv/modules/runners/validate.py
```diff
@@ -768,65 +772,67 @@ def ceph_version(self):
...
+        # When search matches no minions, salt prints to stdout.  Suppress stdout.
+        _stdout = sys.stdout
+        sys.stdout = open(os.devnull, 'w')
+        results = self.local.cmd(search, 'pkg.info_available', [self.package],
+                                 tgt_type="compound")
+        sys.stdout = _stdout
+        for minion in results:
+            if isinstance(results[minion], dict) and self.package in results[minion]:
+                if 'version' in results[minion][self.package]:
+                    version = results[minion][self.package]['version']
```
smithfarm: and here, could we have `log.debug("VALIDATE ceph_version: minion ->{}<- info_available version ->{}<-".format(minion, version))` please

tests/unit/runners/test_validate.py
swiftgist: The two tests for strings are test_check_installed_is_not_installed and test_check_available_has_no_repo.  The contents of the string do not change the test.  The test is for the missing key.  Neither the tests nor the checks do anything with the strings themselves.

I can update the second so that it's similar to the first.

swiftgist: Correction, I just check pkg.info_available and it returns an empty string for no package in a repo.

smithfarm: In my version (which is being replaced here), I have code that sanity-checks the version string returned by pkg.info_installed/pkg.info_available. 

The problem is that these Salt modules appear to assume that zypper will return the output they expect - i.e. a version-like string. If zypper returns an error, they return the text of the error as the version string. That's why I put that sanity-checking code in.

swiftgist:
Besides addressing the issues above, I will be adding some additional logic concerning environments.  I will open a new PR when that is ready.

PR #1110, opened 2018-04-24
force python2 virtualenv for pylint

alexandergraul:
Signed-off-by: Alexander Graul <agraul@suse.com>

`lint` virtualenv defaults to python2 on travis but can default to python3 (e.g. on Tumbleweed). That is a problem for two reasons: the first problem is the lack of consistency between travis and dev setups, the second is that `make lint` does not work with python3 properly.

This change forces a python2 virtualenv to ensure the same `pylint` is used locally and on travis.

I also fixed the broken tox documentation URL.
-----------------


alexandergraul:
I am working on moving it to python3, this is more of a band-aid than anything else.

PR #1113, opened 2018-04-24
openATTIC AppArmor

callithea:
Adds initial openATTIC AppArmor profiles

Signed-off-by: Laura Paduano lpaduano@suse.com


jschmid1:
I'd suggest a rebase to get rid of the linter issues

PR #1114, opened 2018-04-24
Add iputils dependency

swiftgist:

SUSE internal: 1090393


jan--f:
Do we only need ping on the master?

jschmid1:
> Do we only need ping on the master?

A quick grep suggests that

PR #1115, opened 2018-04-25
Refactor basic integration test scripts for Mimic/SES6

smithfarm:
This PR achieves the following objectives:

* make the integration test scripts pass on Mimic
* clearly separate deploy phase from test phase
* scripts no longer care what the current working directory is

This PR does **not** achieve:

* migration of scripts in `qa/suites/ceph-test/` - these are left for later.

As a side effect, the health-igw.sh and health-openattic.sh scripts were dropped - the components they were testing are being reworked for SES6.

Also I broke down and disabled the Stage 0 reboot to reduce testing noise at this early stage of development. Since it causes teuthology to lose contact with the cluster (failing the test), the only purpose this reboot currently serves is to signal when the kernel-default package in the OpenStack image goes out-of-date. This is not entirely un-useful, so I intend to reinstate the reboot later.


qa/common/common.sh
```diff
@@ -2,50 +2,71 @@
...
+}
+
+function update_salt {
+    # make sure we are running the latest Salt before Stage 0 starts,
+    # otherwise Stage 0 will update Salt and then fail with cryptic
+    # error messages
+    TOTAL_NODES=$(json_total_nodes)
+    salt '*' cmd.run 'zypper -n in -f python3-salt salt salt-api salt-master salt-minion'
+    systemctl restart salt-api.service
+    systemctl restart salt-master.service
```
jschmid1: After restarting the master, there's the chance that some minions take longer to re-connect to the master which means that the succeeding call will not reach all the minions.

To be extra sure me might add an extra `sleep` here?

smithfarm: Yeah, I encountered that in some runs. That's why I have the `ping_minions_until_all_respond` call three lines down. But I'll move the sleep up so it's between master restart and minion restart.

jschmid1:
In order to make it more approachable, I'd like to see a small wiki entry/ document where the process is described.

sidenote:
You mentioned something about the ability to run that test now in an already running cluster. How would one do that?

smithfarm:
Added wiki section: https://github.com/SUSE/DeepSea/wiki#integration-test-scripts

smithfarm:
> You mentioned something about the ability to run that test now in an already running cluster. How would one do that?

@jschmid1 I added a commit that checks for a running cluster and skips the deploy phase if one is detected. For this to work, the user running the script on the master node needs to be able run `ceph` commands.

swiftgist:
Is this fine with everyone?

jschmid1:
> Is this fine with everyone?

fine with me

PR #1116, opened 2018-04-26
Implement __utils__ and fix cached-runners issue

jschmid1:
Description: 

Starts tackling https://github.com/SUSE/DeepSea/issues/1062

Due to the changed behavior in salt's lazy loader we have to accommodate and move our often used code in a separate 'utils' directory that will get injected by the `LazyLoader`.

Explanation for changes in proposal.py

in the mentioned proposal[0] file, a global variable is populated with a value of `__utils__`. By the time importlib.load_module() is called[1], the 'pack'(salt's loader.py speak for all dunders that will be injected into the module) is not present. That happens a bit later in the process [2]

The issue seems that the importlib tries to resolve the `__utils__` dict too early and therefore fails with a syntax error.

By wrapping the `__utils__` call in a class/function the dunder gets expanded later, when the pack is already injected.

Gist:

Instead of using python-style importing, we use salt's importing system and access the needed functions via `__utils__['mod.func']()`

Links:

Keeping the links to `srv/modules/runners` for `ready.py` and `deepsea_minions` as we use at least `ready.check` at the orchestration level.
`deepsea_minions` is convenient to have access to.

SUSE Internal: bsc#1087232

[0] https://github.com/SUSE/DeepSea/blob/wip-fix-cached-runners/srv/modules/runners/proposal.py#L93

[1] https://github.com/saltstack/salt/blob/develop/salt/loader.py#L1454

[2] https://github.com/saltstack/salt/blob/develop/salt/loader.py#L1498

-----------------


srv/modules/runners/populate.py
```diff
@@ -1075,15 +1068,15 @@ def engulf_existing_cluster(**kwargs):
     from . import validate
```
jschmid1: Still not sure if that will make problems or how to solve that. I'm not seeing this in the utils directory.

jan--f:
I'm wondering if we were simply going wrong with importing other salt modules. Iiuc we should have been calling a module from a module using ```__salt__``` instead of ```import```. 

Afaiu there is no clear distinction on when one should use ```__util__``` over ```__salt__```. I'm wondering if we can just leave the modules where they are and call them via ```__salt__``` and not have to carry the symlink (since we want to call those modules from cli). Is ``__util__``` myabe for things we'd want to hide from users?

jschmid1:
> I'm wondering if we were simply going wrong with importing other salt modules. Iiuc we should have been calling a module from a module using __salt__ instead of import.

Yes, that's true for all execution/state modules. 


> Afaiu there is no clear distinction on when one should use __util__ over __salt__. I'm wondering if we can just leave the modules where they are and call them via __salt__ and not have to carry the symlink (since we want to call those modules from cli). Is ``util``` myabe for things we'd want to hide from users?

However the `__salt__` global is not injected in runners.. That means we have to put any modules that have to be reused in this directory. However, if we want to call them from the commandline, they have to be present in the `runners_dir:` hence the symlink.

That is less than ideal, but I haven't found a better solution to this, yet.

jan--f:
hmm I see...thx for clearing that up.

jschmid1:
@alexandergraul 

addressed your requests

PR #1117, opened 2018-04-26
MDS updates

jan--f:
This implements the recommended update routine for MDS dameons (shrink active cluster, stop standbys, restart one remaining MDS and bring cluster back to original state).
At the same time this adds a few new cephprocesses functions that allow to short circuit the restart behaviour, see  https://github.com/SUSE/DeepSea/compare/jan-mds-updates?expand=1#diff-fe392dcb6701c6439127296878fd9eda for an example implementation.
Backport is ready to go.

### DIFF
--- a/srv/salt/ceph/mds/restart/reset-mds-cluster.sls
+++ b/srv/salt/ceph/mds/restart/reset-mds-cluster.sls
@@ -1,8 +1,20 @@
 {% set fs_name = salt['pillar.get']('fs_name') %}
 {% set ranks_in = salt['pillar.get']('ranks_in') %}
 
+{% if fs_name == "" or ranks_in == "" %}
+
+no file system name or rank target supplied:
+  test.fail_without_changes:
+    - name: |
+        Please specify both fs_name and ranks_in pillar vars:
+        salt <target> state.apply ceph.mds.restart.shrink-mds-cluster \
+        pillar='{"fs_name": "<name>", "ranks_in": "<target_ranks>"}'
+
+{% else %}
+
 set max_mds back to {{ ranks_in }}:
   cmd.run:
     - failhard: True
     - name: ceph fs set {{ fs_name }} max_mds {{ ranks_in }}
 
+{% endif %}
--- a/srv/salt/ceph/mds/restart/shrink-mds-cluster.sls
+++ b/srv/salt/ceph/mds/restart/shrink-mds-cluster.sls
@@ -1,6 +1,18 @@
 {% set fs_name = salt['pillar.get']('fs_name') %}
 
+{% if fs_name == "" %}
+
+no file system name supplied:
+  test.fail_without_changes:
+    - name: |
+        Please specify fs_name pillar var:
+        salt <target> state.apply ceph.mds.restart.shrink-mds-cluster pillar='{"fs_name": "<name>"}'
+
+{% else %}
+
 set max_mds to 1:
   cmd.run:
     - name: ceph fs set {{ fs_name }} max_mds 1
     - failhard: True
+
+{% endif %}


srv/salt/_modules/cephprocesses.py
```diff
@@ -246,10 +258,7 @@ def need_restart(role=None):
     """
     assert role
     grain_name = "restart_{}".format(role)
```
jschmid1: `grain_name = "restart_{}".format(role)`

can be removed

srv/salt/_modules/wait.py
```diff
@@ -83,16 +70,76 @@ def wait(self):
...
+            log.debug("status != {}".format(self.settings['status']))
+            return (current != self.settings['status'])
+        else:
+            log.debug("status == {}".format(self.settings['status']))
+            return (current == self.settings['status'])
+
+
+class FsStatusCheck(HealthCheck):
+    """
+    Check the fsmap status of the cep status output. Wait till all active MDS's
```
jschmid1: type in `cep_h_`

srv/modules/runners/cmd.py
```diff
@@ -0,0 +1,13 @@
+import salt.client
+
+from subprocess import check_output
+
+def run(**kwargs):
```
jschmid1: I couldn't find any place where you actually call this, but in the light of the recent changes with `__utils__`[0] and the issues with `__pycache__` and runner importing, I suggest to move this to the `__utils__` directory.
.. Unless it's purpose is solely to run arbitrary commands on the master from the commandline.


[0] https://github.com/SUSE/DeepSea/pull/1116

jschmid1:
This change seems to be kinda hard to test. It would make sense to add a small functional test to it

srv/modules/runners/cmd.py
jan--f: It's called right here https://github.com/SUSE/DeepSea/pull/1117/files#diff-4286400fafff454ffeb90b4624230d9aR1

jan--f: But its certainly meant to be a utility function...so :+1:

srv/salt/_modules/cephprocesses.py
jan--f: How is that?

jan--f: ahh...was confused by the line numbers...my bad.

swiftgist:
with branch jan-mds-updates, I am getting

`install: cannot stat 'srv/salt/ceph/restart/mds/lax/*.sls': No such file or directory`

when trying to build the rpm.

jan--f:
@swiftgist oh right..the usual spec/make file oversoght :grin:

srv/modules/runners/cmd.py
jschmid1: > It's called right here https://github.com/SUSE/DeepSea/pull/1117/files#diff-4286400fafff454ffeb90b4624230d9aR1

woops, my bad.

jan--f:
ok this passes ```salt-run state.orch ceph.smoketests.restart.mds```.

Something is not right with the reset though, a second run won't succeed. Will investigate why.

swiftgist:
I ran into missing directories under /srv/salt/ceph/tests/restart/mds/ with the rpm.  The change, nochange and forced I copied over manually.  

I also needed `jq` on the minions.  Once I did that, the smoketest succeeded.  The second and third attempt failed.  I'm still looking at that.

Would you think of moving the state files, reset-mds-cluster.sls and shrink-mds-cluster.sls, from ceph/restart/mds to ceph/mds since ceph/restart has all orchestrations?  

With regards to the runner for the 'ceph fs dump' command, I cannot think of a way to add a `pkg.installed` since Jinja executes prior to the orchestration executing.  We could add `jq` to packages.common, but are we then conflating smoketest requirements with Ceph requirements?

Then, I saw we need this in a few places and I am thinking maybe this should just be in python.  All the sls files would be more direct.  If there becomes a python equivalent (i.e. mon_command variant some day), we can convert the runner or module then.  We wouldn't need `jq` either since python can parse json directly.  This is just thinking out loud about what to do with the `jq` issue.

The last part is that I cannot run that state files easily.  That is

```
salt 'admin*' state.apply ceph.restart.mds.reset-mds-cluster
salt 'admin*' state.apply ceph.restart.mds.shrink-mds-cluster.sls
```

will fail since I have to pass an obnoxious command line with pillar variables.  I know why you are doing it, but I am thinking, can't we get away with defaults?  Running with the idea of a python runner, the fs_name would just be part of the Jinja call.  With regards to the max rank_in, couldn't we use the roles count for mds?  I believe

```
{% set ranks_in = salt.saltutil.runner('select.minions', cluster='ceph', roles='mds') | length%}
```

will work.  If somehow that didn't work, then `select.count` seems like a reasonable addition.  That may simplify the other orchestrations with the calls in the state files and not needing to pass them as pillar overrides.

jan--f:
> I ran into missing directories under /srv/salt/ceph/tests/restart/mds/ with the rpm. The change, nochange and forced I copied over manually.

This is addressed in #1133.

> I also needed jq on the minions. Once I did that, the smoketest succeeded.

jq is only needed on the master, no minion is running jq. I'll look into the dependency on the master.

> The second and third attempt failed. I'm still looking at that.

As I mentioned earlier, the restart grain setting and unsetting in ```smoketests/restarts/common.sls``` is broken. I have an easy fix for that but thought this should be fixed in a separate PR.

> Would you think of moving the state files, reset-mds-cluster.sls and shrink-mds-cluster.sls, from ceph/restart/mds to ceph/mds since ceph/restart has all orchestrations?

Sure makes sense.

> The last part is that I cannot run that state files easily. That is
>
> salt 'admin*' state.apply ceph.restart.mds.reset-mds-cluster
> salt 'admin*' state.apply ceph.restart.mds.shrink-mds-cluster.sls

>will fail since I have to pass an obnoxious command line with pillar variables. I know why you are doing it, but I am thinking, can't we get away with defaults? 

One could just pass the fs_name and current max_mds in the pillar...no command line necessary (and no...there are no sensible defaults). Something like
```
salt -I 'roles:mds' state.apply ceph.restart.mds.shrink-mds-cluster.sls pillar='{"fs_name": "cephfs"}'
```
Admittedly not pretty, but an admin would simply run ```ceph fs set cephfs max_mds 1``` anyway. I don't see the need to conveniently use these states. 

> Running with the idea of a python runner, the fs_name would just be part of the Jinja call. With regards to the max rank_in, couldn't we use the roles count for mds? I believe

> {% set ranks_in = salt.saltutil.runner('select.minions', cluster='ceph', roles='mds') | length%}

> will work. If somehow that didn't work, then select.count seems like a reasonable addition. That may simplify the other orchestrations with the calls in the state files and not needing to pass them as pillar overrides.

The number of mds daemons will almost never be the same as max_mds, there will always be standbys (in fact Ceph will go into HEALTH_WARN if there are no standbys). I don't think putting this in the select runner makes sense semantically. We have no role for active mds (and we won't...this status is ephemeral and managed by the cluster itself) so we'd have to add some voodoo that runs a ceph command. Imho this is misplaced in the select runner.

swiftgist:
If there's no sensible defaults, would you consider a cheap check that those variables are set and bail out with the explicit command?  I'd rather not have to maintain a list of things admins are not expected to run and answer support issues on those.

jan--f:
How's that @swiftgist?

jschmid1:
A couple of issues still persist:

* jq is not installed
  * on a deeper look, it's actually installed, but the smoketest claims (http://pastebin.suse.de/18807)
  * Did you test that patch with salt2018?
* srv/salt/smoketests/init.sls is not present ( which is probably my bad; missing entry in specfile )

jan--f:
Yes salt 2018.3.0 (Oxygen).

Any objections to adding jq to packages.common?

jschmid1:
> Any objections to adding jq to packages.common?

+1



> Yes salt 2018.3.0 (Oxygen).

That's odd. So the smoketests succeed for you?

swiftgist:
Yeah, so we can get this in.  Everything else looks good

jan--f:
@jschmid1 yeah there is an issue with the somketests...thats for a different PR though. When they fail, some minions have the restart grains set. Its an easy fix...coming up next week. For now make sure that all restart grains are set to False.

jan--f:
Ready to go from my side.

jschmid1:
ok, then let's see how we can fix the tests.

The other topic that comes to my mind is backporting. We probably need to collect a bunch of PRs that this PR is dependent on.

smithfarm:
Please consider backporting https://github.com/SUSE/DeepSea/pull/1219 (follow-on fix) along with this one.

Issue #1087, opened 2018-04-16
Latest ceph ("mimic") MON creates bootstrap keyrings when it starts


smithfarm:
### Description of Issue/Question

Apparently, as of https://github.com/ceph/ceph/pull/21236, the MON is creating the bootstrap keyrings when it starts. This breaks Stage 3:

```
2018-04-14T23:14:01.891 INFO:teuthology.orchestra.run.target147135134061.stdout:----------
2018-04-14T23:14:01.891 INFO:teuthology.orchestra.run.target147135134061.stdout:          ID: osd auth
2018-04-14T23:14:01.891 INFO:teuthology.orchestra.run.target147135134061.stdout:    Function: salt.state
2018-04-14T23:14:01.891 INFO:teuthology.orchestra.run.target147135134061.stdout:      Result: False
2018-04-14T23:14:01.891 INFO:teuthology.orchestra.run.target147135134061.stdout:     Comment: Run failed on minions: target147135134061.teuthology
2018-04-14T23:14:01.891 INFO:teuthology.orchestra.run.target147135134061.stdout:     Started: 23:14:00.701318
2018-04-14T23:14:01.891 INFO:teuthology.orchestra.run.target147135134061.stdout:    Duration: 1082.669 ms
2018-04-14T23:14:01.891 INFO:teuthology.orchestra.run.target147135134061.stdout:     Changes:
2018-04-14T23:14:01.891 INFO:teuthology.orchestra.run.target147135134061.stdout:              target147135134061.teuthology:
2018-04-14T23:14:01.891 INFO:teuthology.orchestra.run.target147135134061.stdout:                Name: skip - Function: test.nop - Result: Clean Started: - 23:14:01.124310 Duration: 1.155 ms
2018-04-14T23:14:01.892 INFO:teuthology.orchestra.run.target147135134061.stdout:              ----------
2018-04-14T23:14:01.892 INFO:teuthology.orchestra.run.target147135134061.stdout:                        ID: auth /srv/salt/ceph/osd/cache/bootstrap.keyring
2018-04-14T23:14:01.892 INFO:teuthology.orchestra.run.target147135134061.stdout:                  Function: cmd.run
2018-04-14T23:14:01.892 INFO:teuthology.orchestra.run.target147135134061.stdout:                      Name: ceph auth add client.bootstrap-osd -i /srv/salt/ceph/osd/cache/bootstrap.keyring
[...]

PR #1118, opened 2018-04-26
Add osd bootstrap keyring to monitor creation keyring

swiftgist:
Mimic monitors generate their own keyring.  Continue to use the DeepSea
generated keyring to allow others to be unblocked.


swiftgist:
It took a while, but I was able to test with 13.0.2 and Stage 3 completed.

swiftgist:
I couldn't find it when I was creating it.

Fixes: #1087

PR #1120, opened 2018-04-27
[DNM] [RFC] use jinja macros for simpler dist switch in init.sls

jan--f:
An attempt to simplify the new dist switch mechnism introduced in #1016.

Any init.sls can do the import from ```ceph.foo.sls``` (name pending) which should greatly simplify the swithching proposed in #1016. The rest works as before...if a distro specific path is present it is taken. fallback is default.
I think this makes it less painfull to roll out throughout deepsea.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,6 +224,9 @@ copy-files:
 	install -m 644 srv/salt/ceph/igw/restart/force/*.sls $(DESTDIR)/srv/salt/ceph/igw/restart/force
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/igw/restart/reload
 	install -m 644 srv/salt/ceph/igw/restart/reload/*.sls $(DESTDIR)/srv/salt/ceph/igw/restart/reload
+	# state files - macros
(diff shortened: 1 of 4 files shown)


swiftgist:
Could abspath be moved into the macro?  I am thinking the slspath must not get evaluated when the macro loads (but then again).  That would take down the function to one argument.

But yes, this would keep the init.sls short and sweet.  As far as words, what do you think of `redirect` (i.e. for foo)?  Short, to the point and easy enough to understand the intent.  Any other ideas?

swiftgist:
Another thought: should we have a macros subdirectory?  /srv/salt/ceph/macros?  While it doesn't have to be that specifically, we do not have any files at the top level of /srv/salt/ceph.

jan--f:
Renamed macro to ```os_switch``` and moved it to macros subdir. Thoughts?

Will try if abspath can be moved to the macro.

swiftgist:
I think that works.

jan--f:
k abspath in the macro works as well. I'm happy with this, last chance for objections I guess.

PR #1127, opened 2018-05-08
macros/os_switch: consider not only dirs but also files, add README

jan--f:

This allows to either add, say a SUSE.sls or a more complex SUSE directory with several more complex sls files.

### DIFF
--- a/srv/salt/ceph/macros/README.md
+++ b/srv/salt/ceph/macros/README.md
@@ -1,0 +1,26 @@
+## OS switch macro
+
+Usage:
+
+```
+{% set custom = salt['pillar.get']('your_customization_pillar', 'not a file') %}
+{% from 'ceph/macros/os_switch.sls' import os_switch with context %}
+
+include:
+  - .{{ os_switch(custom) }}
+```
+
+Use this macro to implement an opportunistic operating system switch in your
+state files. The macro will check the directory of the including state files and
+depending on the content output a single state file or directory to include. The
+rules are as follows:
+- Files are preferred over directories. However the directory include can be forced
+  by explicitly including the directory's init.sls
+- The following names are tested in order and returned if either file or
+  directory is present in the including sls file's directory.
+    - ```custom```
+    - value of the ```osfinger``` grain, for example ```CentOS```
+    - concatenation of both the ```os``` and the ```osrelease``` grains, for example ```SUSE15```
+    - the value of the ```os``` grain, for example ```SUSE```
+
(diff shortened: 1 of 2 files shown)


srv/salt/ceph/macros/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+include:
+  - .{{ os_switch(custom) }}
+```
+
+Use this macro to implement an opportunistic operating system switch in your
+state files. The macro will check the directory of the including state files and
+depending on the content output a single state file or directory to include. The
+rules are as follows:
+- Files are preferred directories. However the directory include can be forced
```
alexandergraul: You forgot the "over" in:
> Files are preferred __over__ directories

srv/salt/ceph/macros/os_switch.sls
```diff
@@ -5,12 +5,20 @@
   {%- set osrelease = os + "-" + grains.get('osrelease') -%}
   {%- if salt['file.directory_exists'](abspath + "/" +  custom) -%}
     {{ custom }}
+  {%- elif salt['file.file_exists'](abspath + "/" +  custom + ".sls") -%}
```
jschmid1: I'm aware that a 'path' should not end with a '/' but in this case we are not using it as a standalone variable which makes all if-clauses a bit redundant:

`abspath + "/" +  custom + ".sls"`

vs
` {% set abspath = "/srv/salt/" + slspath + '/' %}`

and then 

`abspath +  custom + ".sls"`

not sure if I like it either :)

srv/salt/ceph/macros/README.md
```diff
@@ -0,0 +1,26 @@
...
+  by explicitly including the directory's init.sls
+- The following names are tested in order and returned if either file or
+  directory is present in the including sls file's directory.
+- ```custom```
```
jschmid1: nitpicky:

maybe use extra indentation here? Technically it's a enumeration within a enumeration:

- The following names are tested in order and returned if either file or￼ directory is present in the including sls file's directory.
  - value of the ```osfinger``` grain, for example ```CentOS```
  - etc..

srv/salt/ceph/macros/os_switch.sls
jan--f: :+1:

srv/salt/ceph/macros/README.md
jan--f: thx!

PR #1140, opened 2018-05-11
Make salt-api an explicit 'make install' dependency

noelmcloughlin:
Fixes #
```
$ sudo make install 
...
...
systemctl restart salt-master
Failed to restart salt-master.service: Unit salt-master.service not found.
make: *** [Makefile:806: install] Error 5
```

Description:

The service `salt-master` may be absent before `salt-api` package gets installed.  Once `salt-api` is installed the `salt-master.service` will definitely exist.  This PR handles the scenario to ensure `salt-master.service` is running and `make install` completes successfully.

Testing: [junit-py27.out.txt](https://github.com/SUSE/DeepSea/files/1996911/junit-py27.out.txt)

-----------------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -825,16 +825,16 @@ copy-files:
 	-chown $(USER):$(GROUP) $(DESTDIR)/srv/salt/ceph/configuration/files/ceph.conf.checksum || true
 
 install-deps:
-	$(PKG_INSTALL) python3-setuptools python3-click python3-tox
+	$(PKG_INSTALL) python3-setuptools python3-click python3-tox salt-api
 
 install: pyc install-deps copy-files
 	sed -i '/^sharedsecret: /s!{{ shared_secret }}!'`cat /proc/sys/kernel/random/uuid`'!' $(DESTDIR)/etc/salt/master.d/sharedsecret.conf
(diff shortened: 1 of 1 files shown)


swiftgist:
Hmm, I do not use the `make install` so much myself, but the restart could be moved to after the salt-api package is installed.  I think the only reason for the original call was to pickup the `deepsea_minions`.  The other option is to add a `$(PKG_INSTALL) salt-master` since that probably should have been there all along.

noelmcloughlin:
Thanks @swiftgist 
I was testing on salt masterless-minion so `salt-master` is not installed - fortunately the `salt-api` declares a package dependency on `salt-master` so my PR only handled the noise.

But you are correct -  having `$(PKG_INSTALL)`  explicitly declare & enforce the dependency is better assumption handling in this case imho.

noelmcloughlin:
Hi @swiftgist I have updated the PR according to your suggestion.  Tested successfully on Fedora. The PR fixes `make: *** [Makefile:806: install] Error 5`

jschmid1:
@noelmcloughlin looks good, but please squash the commits to one

noelmcloughlin:
Done, thanks!

PR #1147, opened 2018-05-16
macros: fix os_switch macro to work on minions.

jan--f:
Before we were using file.file_exists with an absolute path, that is
only valid on the master. As a result we'd get different results on the
master vs. a minion. This relies on salts caching mechanism and works
reliably on all minions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,8 @@ copy-files:
 	# tests
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/tests/keyrings
 	install -m 644 srv/salt/ceph/tests/keyrings/*.sls $(DESTDIR)/srv/salt/ceph/tests/keyrings
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/tests/os_switch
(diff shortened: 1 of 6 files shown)


jschmid1:
@jan--f  Wouldn't that be a good candidate for a functional test?

You can call those macros with different pillar/grain context and check for the called files/states

jan--f:
I'm not aware of how to pass grain values.

jschmid1:
I've only seen the possibility to pass a custom 'pillar' yet. If we can't handle grain data the same way we handle pillar data, it's probably done in a state file. Just like here https://github.com/SUSE/DeepSea/blob/master/srv/salt/ceph/smoketests/restart/common.sls#L11

jan--f:
> it's probably done in a state file.

Nope...since the os_switch is in jinja, a state won't work. This also means we can only test one case per file and it needs a bit of a hack.

swiftgist:
I did not realize until I ran the smoketest that there were two files, custom.sls and file.sls.

Issue #978, opened 2018-03-03
AttributeError: 'NoneType' object has no attribute 'raw_event' in repeat Stage 0


smithfarm:
The `health-openattic.sh` test script in SES5 branch runs stages 0-4 to deploy openATTIC, and then runs Stage 0 a second time. Sometimes, during this repeat run of Stage 0, the following error appears:

```
2018-03-03T11:27:49.038 INFO:teuthology.orchestra.run.target137074029221.stdout:[72/83] Executing state ceph.wait...
2018-03-03T11:28:23.656 INFO:teuthology.orchestra.run.target137074029221.stdout:         |_  target137074029221.teuthology: wait(wait.out) oException in thread Thread-1:
2018-03-03T11:28:23.657 INFO:teuthology.orchestra.run.target137074029221.stdout:Traceback (most recent call last):
2018-03-03T11:28:23.658 INFO:teuthology.orchestra.run.target137074029221.stdout:  File "/usr/lib64/python2.7/threading.py", line 801, in __bootstrap_inner
2018-03-03T11:28:23.658 INFO:teuthology.orchestra.run.target137074029221.stdout:    self.run()
2018-03-03T11:28:23.658 INFO:teuthology.orchestra.run.target137074029221.stdout:  File "/usr/lib/python2.7/site-packages/deepsea/monitor.py", line 576, in run
2018-03-03T11:28:23.658 INFO:teuthology.orchestra.run.target137074029221.stdout:    event.call()
2018-03-03T11:28:23.658 INFO:teuthology.orchestra.run.target137074029221.stdout:  File "/usr/lib/python2.7/site-packages/deepsea/monitor.py", line 461, in call
2018-03-03T11:28:23.659 INFO:teuthology.orchestra.run.target137074029221.stdout:    getattr(self.monitor, self.func)(self.event)
2018-03-03T11:28:23.659 INFO:teuthology.orchestra.run.target137074029221.stdout:  File "/usr/lib/python2.7/site-packages/deepsea/monitor.py", line 634, in end_stage
2018-03-03T11:28:23.659 INFO:teuthology.orchestra.run.target137074029221.stdout:    self._fire_event('stage_finished', tmp_stage)
2018-03-03T11:28:23.660 INFO:teuthology.orchestra.run.target137074029221.stdout:  File "/usr/lib/python2.7/site-packages/deepsea/monitor.py", line 592, in _fire_event
[...]

Issue #814, opened 2017-11-24
render SLS file fails when `slspath` jinja variable is used


rjfd:
If deepsea-cli is parsing/rendering an SLS file that uses the `{{ slspath }}` jinja variable it fails with the following error:

```
Error description:
SaltRenderError: Jinja variable 'slspath' is undefined
```

The problem is caused by the use of the module `cp.get_template` which renders an SLS file but it does not have all the jinja variables in its context.

The CLI render/parser engine must be able to cope with the presence of these jinja variables.

jschmid1:
fixed by #1149

Issue #803, opened 2017-11-22
cli: error while processing execution failures


rjfd:
The below error was triggered while testing a patch for issuer #793 

```
[20/36]   state.orchestrate(mods=['ceph.restart'])................... ❌ (19s)
            |_ changed.mon........................................... ✓ (0.5s)
            |_ cephprocesses.restart_required(role=mon).............. ✓ (0.3s)
                 |_ cephprocesses.restart_required(role=mon) on
                    mon1.ceph........................................ ✓ (0.1s)
            |_ cephprocesses.restart_required(role=mgr).............. ✓ (0.4s)
                 |_ cephprocesses.restart_required(role=mgr) on
                    mon1.ceph........................................ ✓ (0.2s)
            |_ changed.osd........................................... ✓ (0.4s)
            |_ cephprocesses.restart_required(role=storage).......... ✓ (0.8s)
                 |_ cephprocesses.restart_required(role=storage) on
                    data1.ceph....................................... ✓ (0.6s)
            |_ changed.rgw........................................... ✓ (0.6s)
            |_ cephprocesses.restart_required(role=rgw).............. ✓ (0.1s)
            |_ changed.mds........................................... ✓ (0.4s)
            |_ cephprocesses.restart_required(role=mds).............. ✓ (0.1s)
            |_ cephprocesses.restart_required(role=igw).............. ✓ (0.1s)
            |_ cephprocesses.restart_required(role=ganesha).......... ✓ (0.1s)
            |_ cephprocesses.restart_required(role=openattic)........ ✓ (0.1s)
 
Ended stage: ceph.stage.0 succeeded=19/36 failed=1/36 time=51.6s
 
Failures summary:
 
state.orchestrate:
> /usr/lib/python2.7/site-packages/deepsea/monitors/terminal_outputter.py(760)stage_finished()
-> for line in traceback.split('\n'):
(Pdb) traceback
{'outputter': 'highstate', 'data': {'admin.ceph_master': {}}, 'retcode': 1}
(Pdb) c
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib64/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/site-packages/deepsea/monitor.py", line 576, in run
    event.call()
  File "/usr/lib/python2.7/site-packages/deepsea/monitor.py", line 461, in call
    getattr(self.monitor, self.func)(self.event)
[...]

PR #1149, opened 2018-05-17
cli: fix stage parsing for salt 2018

### LINKED ISSUE #789: deepsea CLI hangs when no Minions have been configured
This is on openSUSE Leap 42.3, using Salt 2017.7.2 and DeepSea 0.7.35.

After accepting all Minion Keys but not configuring `/srv/pillar/ceph/deepsea_minions.sls`, running stage 0 via the `deepsea` CLI just hangs until I abort it with `Ctrl+C`:
```
# deepsea stage run ceph.stage.0
Starting stage: ceph.stage.0
Parsing ceph.stage.0 steps... ⏳
^C
Aborted!
```
Using the `salt` commandline tool reveals my mistake:
```
# salt-run state.orch ceph.stage.0
[WARNING ] /usr/lib/python2.7/site-packages/salt/client/__init__.py:705: DeprecationWarning: The target type should be passed using the 'tgt_type' argument instead of 'expr_form'. Support for using 'expr_form' will be removed in Salt Fluorine.

No minions matched the target. No command was sent, no jid was assigned.
No minions matched the target. No command was sent, no jid was assigned.
deepsea_minions          : ['No minions matched for G@deepsea:* - check /srv/pillar/ceph/deepsea_minions.sls']
master_minion            : ['Missing pillar data']
[ERROR   ] No highstate or sls specified, no execution made
ceph-01.fritz.box_master:
----------
          ID: validate failed
    Function: salt.state
        Name: just.exit
      Result: False
     Comment: No highstate or sls specified, no execution made
     Started: 14:47:01.612702
    Duration: 0.508 ms
     Changes:   

Summary for ceph-01.fritz.box_master
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:   0.508 ms
```
The `deepsea` CLI should have also aborted with a clear error message instead of just hanging here.

rjfd:
Fixes #789 #803 #814 #978

Description:
This PR re-implements the CLI stage parser to work with Salt 2018 versions.

It also improves the `dry-run` functionality by showing a plan of execution with similar aspect of the real execution.


swiftgist:
The `make rpm` fails since it tries to the cli tests.  I changed *tox.ini* from

```
commands =  py.test --tb=line
```
to
```
commands =  py.test --ignore=cli --tb=line
```

Once I installed docker, the `run-unit-tests.sh` worked for me.  I did notice I needed to be in the directory though.  I don't know about adding `(cd cli; ./run-unit-tests.sh)` to the *test* target though.  The 312 unit tests take ~3.9s for me.  The cli unit tests take 1m25s.  I think we should have a *test-cli* target and probably an *all-tests* target.  See what everyone thinks.

That's as far as I've gotten at the moment... I seem to have hung docker.

rjfd:
@jschmid1 @swiftgist added the `--ignore=cli/`

swiftgist:
I am installing the rpm and ran into this

```
# deepsea salt-run state.orch ceph.stage.1
  File "/usr/bin/deepsea", line 5, in <module>
    from pkg_resources import load_entry_point
<snip>
pkg_resources.DistributionNotFound: The 'deepsea==0.9.4+98.gf7a2b376' distribution was not found and is required by the application
```
DeepSea is installed with the rpm as well:
```
# rpm -q deepsea
deepsea-0.9.4+98.gf7a2b376-0.noarch
```

I tried the easy_setup suggestions, but no change.

jschmid1:
@swiftgist IIRC, this is due to a wrong version of setuptools. Please check if you have python3-setuptools and not python2-setuptools installed.

srv/salt/ceph/monitoring/prometheus/update_service_discovery.sls
```diff
@@ -21,8 +21,10 @@
     - dir_mode: 755
     - clean: True
     - makedirs: True
+{% if salt.saltutil.runner('select.minions', cluster='ceph', host=False) %}
```
jan--f: Just out of curiosity: did you actually run into this? This should only happen when there are no minions or what am I missing?

cli/deepsea.py
```diff
@@ -119,80 +122,144 @@ def _validate_stage_file_exists(stage_name):
         sys.exit(1)
 
 
-def _run_show_stage_steps(stage_name, hide_state_steps, only_visible_steps, use_cache):
+def _print_stage_step(step, indent, step_order_map):
+    if isinstance(step, SaltState):
+        PP.print(PP.orange(step.sls))
+        PP.print(PP.dark_green(" ({})".format(step.desc)))
+        PP.println(PP.orange(" on"))
+        if step.on_success_deps or step.on_fail_deps:
```
alexandergraul: Why do we check this twice? Can't we simplify these nested if branches to something like:
```python
if step.on_success_deps:
    # success branch
elif step.on_fail_deps:
   # fail branch
else:
    PP.println()

File: cli/deepsea.py [hunk 3]
```diff
@@ -119,80 +122,144 @@ def _validate_stage_file_exists(stage_name):
...
+                step_order_map[id(s_step)] = str(step_count)
+                _print_stage_step(s_step, indent + 11, step_order_map)
+                step_count += 1
+
+    elif isinstance(step, SaltRunner):
+        PP.print(PP.blue(step.function))
+        PP.println(PP.dark_green(" ({})".format(step.desc)))
+        if step.on_success_deps or step.on_fail_deps:
+            if step.on_success_deps:
+                PP.print("{}".format(" " * indent))
```
alexandergraul: Since we use the same code here as above, does it make sense to pull it out into a method?
I am thinking one method for `step.on_success_deps` and one for `step.on_fail_deps` that takes care of the printing.

cli/deepsea.py
```diff
@@ -119,80 +122,144 @@ def _validate_stage_file_exists(stage_name):
...
+                                         only_visible_steps)
+    except StageRenderingException as ex:
+        PP.println(PF.FAIL)
+        PP.println()
+        PP.println(PP.bold("An error occurred while rendering stage: {}"
+                           .format(ex.stage_name)))
+        for error in ex.error_list:
+            PP.println(PP.red("  - {}".format(error)))
+        PP.println()
+        return
```
alexandergraul: I think it is better to have an explicit `return None` for readability

cli/deepsea.py
```diff
@@ -119,80 +122,144 @@ def _validate_stage_file_exists(stage_name):
...
+    except StateRenderingException as ex:
```
alexandergraul: This looks like it is unintentional :wink:

cli/monitor.py
```diff
@@ -136,20 +129,18 @@ def __init__(self, name, steps, enable_dynamic):
...
             elif isinstance(step, SaltState):
-                wrapper = Stage.TargetedStep(step, step.state, len(self._steps)+1)
-                _curr_state = wrapper
-            elif isinstance(step, SaltModule) or isinstance(step, SaltBuiltIn):
-                assert _curr_state
-                _curr_state.sub_steps.append(Stage.Step(step, step.fun,
-                                                        len(_curr_state.sub_steps)+1))
-                continue
+                wrapper = Stage.TargetedStep(step, step.sls, len(self._steps)+1)
+                for _, s_steps in step.steps.items():
```
alexandergraul: I could not figure out where `step.steps` comes from, is it a dict? If so, I think it is cleaner to just iterate over its values with `for s_steps in step.steps.values()`

cli/tests/start.sh
```diff
@@ -0,0 +1,110 @@
...
+
+    echo "Starting salt-minion minion$i ..."
+    salt-minion -c /etc/salt/minion$i -d
+done
+
+
+echo "Waiting for salt-minions to connect with salt-master..."
+for i in `seq $NUM_MINIONS`; do
+    while true; do
+        FOUND=`grep "^salt/auth" /var/log/salt.event.log | grep "minion$i" | wc -l`
```
kbaikov: Please implement a timeout for this loop

cli/tests/start.sh
```diff
@@ -0,0 +1,110 @@
...
+    done
+    echo "salt-minion minion$i connected."
+done
+
+echo "Accepting minion keys..."
+salt-key -Ay
+
+echo "Waiting for salt-minion to start..."
+for i in `seq $NUM_MINIONS`; do
+    while true; do
```
kbaikov: Also a timeout needed for the loop

srv/salt/ceph/monitoring/prometheus/update_service_discovery.sls
rjfd: Yes, I run into this problem, but it only happens when running the CLI. The CLI parses all state files that are part of a stage. The parsing requires that all jinja statements are evaluated so that we can actually know what is going to execute. This jinja statement assumes that `select.minions` will return something because the configuration was already applied in the beginning of the stage, but the problem is that we are just parsing and not executing the state files, and therefore the configuration was not applied yet.

cli/deepsea.py
rjfd: Not in this case because the `PP.println()` in the `else` branch can only execute if the `step.on_success_deps` evaluates to true, otherwise it should not be executed at all.

cli/deepsea.py
rjfd: but this function does not return anything (void)

cli/monitor.py
rjfd: `step.steps` comes from https://github.com/SUSE/DeepSea/pull/1149/files#diff-6c2e6839d23a2408ea246962365247edR437 and yes it's a dict

rjfd:
@kbaikov @alexandergraul addressed your comments and pushed the changes

jschmid1:
@rjfd  This is what I encounter now: http://paste.opensuse.org/87788864

jschmid1:
can be fixed with

``` python
if hasattr(ex_step, 'targets'):
```

in the monitor's line 238

I also tried to backport the whole thing to SES5, with seems to work just fine with `s/tgt_type/expr_form/`

I'll label it accordingly and backport this with 0.8.5 when this is merged

jschmid1:
What I'd also really like to see ( probably in another PR ) is an evolving 'Parsing Stage.. ' line. It happened several times in the past that the stage parsing hung forever, due to reasons which are hard to debug at glance.  In the logs I saw that you output the current file/state that is being processed. This would be a good candidate to print imho.

TLDR:

Instead of:

`Parsing ceph.stage.0 steps... ⏳`

do:

`Rendering(parsing) state .. <state_name> `

to provide better feedback

rjfd:
@jschmid1 I added a new commit with the implementation of progress feedback of the parsing stage.

PR #1153, opened 2018-05-18
Do not hardcode cluster name

kbaikov:
Signed-off-by: kbaikov 

Fixes #

Description: This fixes a longstanding TODO of hardcoding the name of the cluster

-----------------


srv/modules/runners/proposal.py
```diff
@@ -266,8 +278,8 @@ def _write_proposal(prop, profile_dir):
         # implement merge of existing data
         yaml.dump(content, outfile, default_flow_style=False)
 
-    # TODO do not hardcode cluster name ceph here
-    profile_file = '{}/stack/default/ceph/minions/{}.yml'.format(profile_dir,
+    profile_file = '{}/stack/default/{}/minions/{}.yml'.format(profile_dir,
+                                                                 _cluster_name(),
```
jan--f: Did you run ```make lint```? I suspect the linter will cough on the indentation here.

kbaikov:
@jan--f lint fixed

jan--f:
@jschmid1 this will work though won't it? I'd count it as an improvement...more instances can be fixed in another PR?

jschmid1:
I'd call it inconsistency. It'll become one of these things where you ask yourself after a couple of month why this thing exists in one place and not in another.

Voting for either a complete refactor or none

srv/modules/utils/cluster.py
```diff
@@ -0,0 +1,45 @@
...
+def name():
+    """
+    Return the cluster name from the ceph namespace, original namespace
+    or default to 'ceph'
+    """
+    if ('ceph' in __pillar__ and 'cluster' in __pillar__['ceph']):
+        return __pillar__['ceph']['cluster']
+    elif 'cluster' in __pillar__:
+        return __pillar__['cluster']
+    else:
```
swiftgist: The only issue here is that the proposal runner is driven by the admin rather than the pillar.  The intent is that the cluster name could be set, maybe via parameter.  We can investigate later, but hardcoding this to returning 'ceph' now is okay.

jschmid1:
approving as linting will be fixed in #1163

tserong:
There was a thread on ceph-devel a while ago about removing cluster name support (https://www.spinics.net/lists/ceph-devel/msg37035.html), so we need to be careful with this.

To be clear, the upstream thread is about removing cluster names as an *internal* concept.  For example, where previously you could have `/var/lib/ceph/{mon,osd}/$cluster-$id/` as a path, with an arbitrary name for `$cluster`, going forwards this will always just be `/var/lib/ceph/{mon,osd}/ceph-$id`.

It's certainly still useful for clients to be able to talk to multiple clusters, and to be able to have one instance of DeepSea that can manage multiple ceph clusters, but we probably need to make sure that the cluster names remain an administrative concept (i.e. visible to clients, the administrator, and to DeepSea), and don't bleed into actual deployed cluster settings.

For example, right now the mon deployment code uses the cluster name when it invokes `ceph-mon --mkfs`, and in the path to the mon keyring file (https://github.com/SUSE/DeepSea/blob/master/srv/salt/ceph/mon/default.sls#L25-L33). This, and any other similar code, would need to be changed to just hardcode the string "ceph" instead.

smithfarm:
This seems ill-advised when, for now and the foreseeable future, the cluster name is never going to be anything but "ceph" (as far as Ceph is concerned). Why take the risk? Or, in other words, IMO this is fixing something that isn't broken, for no compelling reason.

It would be different if this (or something like it) were being done as part of a new feature enabling DeepSea to manage multiple Ceph clusters. But just on its own?

PR #1154, opened 2018-05-22
Change log level to info for the ready message

kbaikov:
Description: This is warning was a confusing message for one of the users. Perhaps and info level is more appropriate.

-----------------


jan--f:
Abusing ```log.warning``` is currently our mode of choice to log messages to the user. Using ```log.info``` would require a config change and salt is _very_ noisy at that level drowning out most messages.

Issue #1128, opened 2018-05-08
Travis checks are gone


jschmid1:
### Description of Issue/Question

Github changed the way how checks work.

https://blog.github.com/2018-05-07-introducing-checks-api/

For Travis:

https://blog.travis-ci.com/2018-05-07-announcing-support-for-github-checks-api-on-travis-ci-com

It got extended quite a bit.. probably worth looking at some new features we may use.

alexandergraul:
The new API (currently in public beta) is for Github Applications. As of now, only Travis CI is an interesting GH-App supporting this new feature. To make use of it, we need add the [Travis App](https://github.com/marketplace/travis-ci/plan/MDIyOk1hcmtldHBsYWNlTGlzdGluZ1BsYW43MA==#pricing-and-setupl)

jschmid1:
replaced by internal jenkins

PR #1159, opened 2018-05-22
Add codecov integration to Travis CI

alexandergraul:
Signed-off-by: Alexander Graul <agraul@suse.com>

Description:
Codecov uses code coverage generated by pytest-cov and generates stats
and rich diffs for PRs.
Before we can use this, two things have to happen:
1. We need to discuss the numbers (how much % drop do we allow before the check fails) [[docs](https://docs.codecov.io/docs/commit-status)]
2. Someone with admin rights needs to enable [Travis](https://github.com/marketplace/travis-ci/plan/MDIyOk1hcmtldHBsYWNlTGlzdGluZ1BsYW43MA==#pricing-and-setupl) (see #1128) and [Codecov](https://github.com/marketplace/codecov/plan/MDIyOk1hcmtldHBsYWNlTGlzdGluZ1BsYW4xNg==#pricing-and-setup)

-----------------

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,17 @@
+coverage:
+  precision: 2
+  round: down
+  range: 25..30
+
+  status:
+    project:
+      default:
+        target: null
+    patch:
+      default:
+        target: null
+    changes: no
+
+comment: 
+  layout: "diff"
+  behavior: once
--- a/tox.ini
+++ b/tox.ini
@@ -13,7 +13,11 @@ deps =
 
 [testenv:py27]
 basepython = python2.7
-commands =  py.test --cov=. --cov-report html --cov-report term --tb=line -v --ignore=cli/ --junitxml=junit-{envname}.xml {posargs}
+commands =
+    py.test --cov=. --tb=line -v --ignore=cli/ --junitxml=junit-{envname}.xml {posargs}
+    codecov
+# the environment variables are needed by codecov
+passenv = CI TRAVIS TRAVIS_*
 deps =
     {[base]deps}
     mock
@@ -23,6 +27,8 @@ deps =
     salttesting
     configobj
     boto
+    codecov
+    pytest-cov
 
 [testenv:lint]
 basepython = python3


alexandergraul:
What I forgot to mention is that we should also discuss if we want codecov reports on PRs. Currently they are only available as comments, I hope they integrate the new Github Checks API soon. Codecov can still enforce limits without the comments.

There are different pieces of information available in reports:
1. Reach (displays a graphic made out of boxes, every box is a file. Color of the box depends on change in coverage)
2. Diff stats
3. List of affected files with their coverage delta
4. Flags (these are useful if there is more than one "thing" that is covered, e.g. API and UI - I don't think we can make good use out of them currently)

jschmid1:
> We need to discuss the numbers (how much % drop do we allow before the check fails) [docs]

Voting for a threshold of 25%. This is the current coverage that we have and should not drop below.


> Someone with admin rights needs to enable Travis (see #1128) and Codecov

pinging lars/holgi/denis when the decision stands.


> Reach (displays a graphic made out of boxes, every box is a file. Color of the box depends on change in coverage)

I don't find that box-graph terribly helpful I must say. I think I'd vote for disabling the comment all together. If one is interested in the report, you can still click on the check-report-thing on the bottom of the PR, right?

There are also browser plugins for that purpose, which personally find quite nice and a lot less obtrusive ( thanks alex for finding that out. https://docs.codecov.io/docs/browser-extension )

.travis.yml
```diff
@@ -9,10 +9,10 @@ env:
 matrix:
   fast_finish: true
   include:
-    - env: TOXENV=lint
-      python: '2.7'
-    - env: TOXENV=py27
-      python: '2.7'
+    - python: '2.7'
+      env: TOXENV=lint
```
kbaikov: Perhaps python 3.5 for lint? or soon will be with this PR: https://github.com/SUSE/DeepSea/pull/1122

alexandergraul: I think we should use #1122 for that change since new pylint messages came up after I changed from python2 to python3 which I have fixed in that PR. I propably did not need to change this file but I think the syntax was wrong and should be fixed (even though it works if all python versions are the same)

alexandergraul:
> If one is interested in the report, you can still click on the check-report-thing on the bottom of the PR, right?
Yes, clicking on "Details" takes you to the full report:
screenshot from 2018-06-05 13-46-09

jschmid1:
ping @jan--f @swiftgist

smithfarm:
Backported to SES5 via https://github.com/SUSE/DeepSea/pull/1280

PR #1160, opened 2018-05-23
Install product specific package when on SLES

jschmid1:

Description:

Specific versions of DeepSea are dependent on specific version of salt. Trying to enforce that can be done by either:
Enforcing this via the DeepSea specfile is the first obvious idea, but this solutions fails to lock the salt-minions as DeepSea is only installed on the _master_. To resolve this we have the following options.

* via zypper lock ( not distrowide )
* via a meta package that adds a "Conflicts: salt-minion <= 2016.11.4, salt-minion >= 2016.x.x?"
   which is installed on all minions. ( needs a a separate package and only works with RPM based 
   distros (a separate package needs to be created for *.deb based distros))

Additional caveats:

* If the channel from which you receive new updates has staged an update _before_ you locked down 
   your version, you are in a bad spot.
* This patch adds a intermediate step in _stage.0_ which is there to chime in _before_ the packages are updated ( This is especially true for the master part of stage.0, adding it to the minion part too )

SUSEInternal: ( 323819, 1094311, 1094314)

EDIT: If this is the way we want to go, probably change the name from 'hotfix' to 'metapackage' or something like this.

fixes: #1167 

-----------------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -309,6 +309,10 @@ copy-files:
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/salt-api/files
 	install -m 644 srv/salt/ceph/salt-api/files/*.conf* $(DESTDIR)/srv/salt/ceph/salt-api/files
 
+	# state files - hotfix
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/hotfix
+	install -m 644 srv/salt/ceph/hotfix/*.sls $(DESTDIR)/srv/salt/ceph/hotfix
+
 	# state files - migrate
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/migrate
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/migrate/osds
--- a/Makefile
+++ b/Makefile
@@ -309,9 +309,9 @@ copy-files:
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/salt-api/files
 	install -m 644 srv/salt/ceph/salt-api/files/*.conf* $(DESTDIR)/srv/salt/ceph/salt-api/files
 
-	# state files - hotfix
-	install -d -m 755 $(DESTDIR)/srv/salt/ceph/hotfix
-	install -m 644 srv/salt/ceph/hotfix/*.sls $(DESTDIR)/srv/salt/ceph/hotfix
+	# state files - metapackage
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/metapackage
+	install -m 644 srv/salt/ceph/metapackage/*.sls $(DESTDIR)/srv/salt/ceph/metapackage
 
 	# state files - migrate
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/migrate
--- a/deepsea.spec.in
+++ b/deepsea.spec.in
@@ -105,6 +105,7 @@ systemctl try-restart salt-api > /dev/null 2>&1 || :
 %dir /srv/salt/ceph/cephfs
 %dir /srv/salt/ceph/cephfs/benchmarks
 %dir /srv/salt/ceph/cephfs/benchmarks/files
+%dir /srv/salt/ceph/hotfix
 %dir /srv/salt/ceph/salt-api
 %dir /srv/salt/ceph/salt-api/files
 %dir /srv/salt/ceph/configuration
@@ -424,6 +425,7 @@ systemctl try-restart salt-api > /dev/null 2>&1 || :
 %config /srv/salt/ceph/benchmarks/fs/*.sls
 %config /srv/salt/ceph/cephfs/benchmarks/*.sls
 %config /srv/salt/ceph/cephfs/benchmarks/files/keyring.j2
+%config /srv/salt/ceph/hotfix/*.sls
 %config /srv/salt/ceph/salt-api/*.sls
 %config /srv/salt/ceph/salt-api/files/*.conf*
 %config /srv/salt/ceph/tools/benchmarks/*.sls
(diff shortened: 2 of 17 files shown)


jschmid1:
Staging backport(SES4/SES5) when this is approved.

alexandergraul:
Do I understand this correctly: we will adapt the ses-release.rpm to inlcude the conflicts for salt-minon.rpm and install ses-release.rpm before updating the software stack on all minions? Just for clarification: this is SUSE specific, right? 

I think this is better than using locks since it more verbose. If someone wants to install a newer salt-minion version it will be clear that there is a dependency for SES. In the other case there is only an entry in the locks, which requires additional documentation why it was added.

> If the channel from which you receive new updates has staged an update before you locked down
your version, you are in a bad spot.

Can we use salt to reinstall salt-minion at a specific version? If not, I can't think of a way to deal with this issue after the fact.

jschmid1:
> Do I understand this correctly: we will adapt the ses-release.rpm to inlcude the conflicts for salt-minon.rpm and install ses-release.rpm before updating the software stack on all minions? Just for clarification: this is SUSE specific, right?

Yes, it's suse specific right now. The ses-release.rpm needs to be adapted to any kind of meta-package that will be installed alongside the minions.



> Can we use salt to reinstall salt-minion at a specific version? If not, I can't think of a way to deal with this issue after the fact.

We can and can not:

If a user somehow misses to install the 'meta-rpm', the next time he will use deepsea ( under the circumstance that a new salt is staged in his repos ) salt will be updated to the next version. If that's the case, you will probably end up with a working salt cluster which is operational and can be downgraded to a older version. The 'how' is the more important part I'd say as you now can't use DeepSea anymore, because we possibly have incompatibilities with the newly installed salt version.
Your last resort is to use salt to downgrade salt in order to make deepsea work again .. :disappointed:

alexandergraul:
For the case that meta-rpm was not installed before salt was updated we can use the salt version validation that got introduced in #1074, adjust it to the correct versions and make it an error instead of a warning. Then document the salt command to downgrade the salt version on the minions. It is ugly, but at that point (wrong salt version is present) there are less options...

jschmid1:
Fair point. I think we should add that once we reached consensus on the general approach

jan--f:
Just to make a case for using zypper locks:
1. All major package formats/managers offer a way to pin packages (at least true for zypper, apt and yum)
2. The pins are mostly file managed, so should be easy to pin packages with salt/deepsea.
3. No meta packages required -> less administrative overhead for us (I assume we'd create the meta packages for all "supported" distros?)

I'm not sure what happens if, say salt 2018 is installed and a version pin for 2016 is added after the fact, but I'll investigate.
I find version pinning seems like the more appropriate tool for this kind of thing and has less overhead (no meta packages).

holgis:
Independently of the current salt issue the ses-release package needs to be installed. It tells the system that this is a SES product and which version of it. If someone would install SES manually via SUSE Installer/YaST, this package gets installed automatically.
So at least for SLES/SES product, this is no artificial meta package but one that is around and necessary.
For the salt issue with the conflicts, it would be a reliable solution, that is designed for such cases.

holgis:
For testing purposes I just submitted a new ses-release package with "Conflicts: salt > 2016.12.31" to Devel:Storage:5.0 - should be built soon.

jschmid1:
@jan--f There is a add_lock implementation for zypper at least. https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.zypper.html

jschmid1:
and the 'hold' counterpart for apt
https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.aptpkg.html

jschmid1:
https://github.com/SUSE/DeepSea/pull/1160/commits/5b26f8ad35b0cd7a225f469085585e43b1b0c914 contains the lock/apt implementation. 

The downside to this approach is that we can't really define a range of versions that we support. If user A installs Deepsea in January he'll be locked down on version 201x.01.x while a user that installs later in the year he'll be locked on the version that is released in this timeframe. Not a terrible caveat, but still ..

holgis:
Conflict resolution seem to remove package locks:

#### YaST2 conflicts list - generated 2018-05-24 13:27:00 ####

product:ses-5-0.x86_64 conflicts with salt > 2016.12.31 provided by salt-2018.3.0-4.11.1.x86_64
    [ ] Following actions will be done:
remove lock to allow removal of salt-2018.3.0-4.11.1.x86_64
remove lock to allow removal of salt-master-2018.3.0-4.11.1.x86_64
downgrade of salt-2018.3.0-4.11.1.x86_64 to salt-2016.11.4-46.20.2.x86_64
downgrade of salt-master-2018.3.0-4.11.1.x86_64 to salt-master-2016.11.4-46.20.2.x86_64
    [ ] do not install product:ses-5-0.x86_64

python2-salt-2018.3.0-4.11.1.x86_64 requires salt = 2018.3.0-4.11.1, but this requirement cannot be provided
    [ ] Following actions will be done:
remove lock to allow removal of python2-salt-2018.3.0-4.11.1.x86_64
remove lock to allow removal of salt-api-2018.3.0-4.11.1.x86_64
remove lock to allow removal of salt-master-2018.3.0-4.11.1.x86_64
downgrade of salt-2018.3.0-4.11.1.x86_64 to salt-2016.11.4-46.20.2.x86_64
    [ ] Following actions will be done:
remove lock to allow installation of salt-2018.3.0-4.11.1.x86_64[salt]
remove lock to allow removal of salt-master-2018.3.0-4.11.1.x86_64
    [ ] do not install product:ses-5-0.x86_64
    [ ] break python2-salt-2018.3.0-4.11.1.x86_64 by ignoring some of its dependencies

#### YaST2 conflicts list END ###

holgis:
Installing ses-release (with conflicts salt > 216.12.31) on SES4 and SES5 (and if necessary future versions) is solving the issue for our active products. So let's go ahead with this solution.

I do not see how locking the salt version (so far the patch does not even check the installed version that it will lock) will help us in scenarios like new installation with a higher salt being around or on non SES products were we do not know which salt version is available and installed.

jschmid1:
@swiftgist are you fine with this approach?

PR #1161, opened 2018-05-23
[backport][SES5] MDS restarts & smoketests

jan--f:
This backports the mds restart with shrinking the cluster. This also includes the backported smoketests.

```salt-run state.orch ceph.smoketests.restart.mds``` passes.


jan--f:
Yes I saw this too but chalked it up to salt being salt.

PR #1162, opened 2018-05-25
Fix typos and some lints

kbaikov:
Description: This addresses some lint issues and fixes typos in readme.
Please review, since make lint fails for me, but does not report any errors.

-----------------

### DIFF
--- a/srv/modules/runners/cephprocesses.py
+++ b/srv/modules/runners/cephprocesses.py
@@ -17,7 +17,7 @@ import logging
 import salt.client
 import salt.utils
 import salt.utils.master
-import six
+import salt.ext.six as six
 
 log = logging.getLogger(__name__)
 
@@ -242,6 +242,7 @@ def _timeout(cluster='ceph'):
         return 900
     else:
         return 120
+
 
 __func_alias__ = {
                  'help_': 'help',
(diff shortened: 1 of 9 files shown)


srv/modules/runners/cephprocesses.py
```diff
@@ -17,7 +17,7 @@
 import salt.client
 import salt.utils
 import salt.utils.master
-import six
+import salt.ext.six as six
```
jan--f: I wasn't sure about this change. We import six in other places just fine...ping @jschmid1?

srv/modules/utils/cluster.py
```diff
@@ -0,0 +1,44 @@
...
+def help_():
+    """
+    Usage
+    """
+    usage = ('salt-run cluster.name: \n\n'
+             '    Returns a name of the cluster\n'
+             '\n\n')
+    print(usage)
+    return ""
+
```
swiftgist: Does a utils module work as a pseudo runner?  That seems unexpected.

kbaikov and I were chatting about the name function in #1153.  I think the name function is fine when called from modules since the pillar setting implies membership.  However, for the subset of runners in Stage 0-2 should never use this since the admin would be setting the cluster based on command line argument or environment variable.  (This is for the fictional day in the future where we support multiple Ceph clusters with one Salt master.)

We never solved that issue and I don't know that we have to now.  I do think we need to have enough commentary to explain when the utils should not be used.

kbaikov:
Closing this in favor of the https://github.com/SUSE/DeepSea/pull/1163

PR #1163, opened 2018-05-25
fix lint issues

jan--f:
Fixes #

Description:

-----------------

### DIFF
--- a/srv/salt/_modules/wait.py
+++ b/srv/salt/_modules/wait.py
@@ -15,7 +15,12 @@ import rados
 log = logging.getLogger(__name__)
 
 
+# pylint: disable=too-few-public-methods
 class HealthCheck(object):
+    """
+    This is the base class for health checks and offers the generics of running
+    a command and checking its output until a timeout is reached.
+    """
 
     def __init__(self, **kwargs):
         """
@@ -52,7 +57,7 @@ class HealthCheck(object):
 
         log.debug('wait on condition of command {}'.format(cmd))
         while i < (self.settings['timeout']/self.settings['delay']):
-            ret,output,err = self.cluster.mon_command(cmd, b'', timeout=6)
+            _ret, output, _err = self.cluster.mon_command(cmd, b'', timeout=6)
             json_output = json.loads(output)
 
             if success(json_output):
@@ -74,30 +79,32 @@ class HealthCheck(object):
         """
         Return the "correct" matching status
         """
+        # pylint: disable=no-else-return
         if self.settings['negate']:
             log.debug("status != {}".format(self.settings['status']))
-            return (current != self.settings['status'])
+            return current != self.settings['status']
         else:
             log.debug("status == {}".format(self.settings['status']))
-            return (current == self.settings['status'])
+            return current == self.settings['status']
 
 
+# pylint: disable=too-few-public-methods
 class FsStatusCheck(HealthCheck):
     """
     Check the fsmap status of the ceph status output. Wait till all active MDS's
     daemons have reached up:active status
     """
 
-    def __init__(self, **kwargs):
-        super(FsStatusCheck, self).__init__(**kwargs)
-
     def wait_for_healthy_mds(self):
         """
         Poll until all active MDS' are up:active
         """
-        cmd = json.dumps({"prefix":"status", "format":"json" })
+        cmd = json.dumps({"prefix": "status", "format": "json"})
 
         def success(status):
+            """
+            Success function to be passed to _check_status
+            """
             if 'fsmap' in status:
                 fsmap = status['fsmap']
             else:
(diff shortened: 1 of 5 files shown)


swiftgist:
Hmm, so I hit

```
/local/eric/src/DeepSea/.tox/lint/lib/python3.4/site-packages/saltpylint/minpyver.py:31: RuntimeWarning: Unable to import saltpylint.ext.pyqver. Running on Python 3? Checker skipped.
  RuntimeWarning
************* Module srv.modules.runners.cephprocesses
E: 20, 0: Uses of an external blacklisted import 'six': Please use 'import salt.ext.six as six' (blacklisted-external-import)
```

I'm guessing the `six` import is something new.  I saw your mention of it in #1162.  I haven't checked into pyqver.

srv/salt/_modules/wait.py
```diff
@@ -119,16 +126,16 @@ class HealthStatusCheck(HealthCheck):
...
         """
         Poll until the status "matches" the specificed number of checks.
         """
-        cmd = json.dumps({"prefix":"health", "format":"json" })
+        cmd = json.dumps({"prefix": "health", "format": "json"})
 
         def success(health):
+            """
+            sucess function to passed tp _check_status
+            """
```
kbaikov: 2 typos: success and to pass.
The same as the function on line 106

srv/salt/_modules/wait.py
```diff
@@ -74,30 +79,32 @@ def _check_status(self, current):
...
         """
         Poll until all active MDS' are up:active
         """
-        cmd = json.dumps({"prefix":"status", "format":"json" })
+        cmd = json.dumps({"prefix": "status", "format": "json"})
 
         def success(status):
+            """
+            sucess function to passed tp _check_status
+            """
```
kbaikov: 2 typos: success and to pass.

srv/salt/_modules/wait.py
```diff
@@ -74,30 +79,32 @@ def _check_status(self, current):
...
+# pylint: disable=too-few-public-methods
 class FsStatusCheck(HealthCheck):
     """
     Check the fsmap status of the ceph status output. Wait till all active MDS's
     daemons have reached up:active status
     """
 
-    def __init__(self, **kwargs):
-        super(FsStatusCheck, self).__init__(**kwargs)
-
```
kbaikov: I believe this super had a purpose. E.g. not to inherit stuff from parent class?
Or similar: https://stackoverflow.com/questions/576169/understanding-python-super-with-init-methods

I decided not to touch it in my PR.

@swiftgist can you please elaborate on this super?

jan--f: The code is mine. The purpose of the call to super was to have the constructor of the base class called. There is no reason to explicitly call the super constructor, its just inherited....right?

kbaikov: Right! Just super(FsStatusCheck, self).__init__(**kwargs) looked strange to me, so i assumed it does some mega magic.

srv/salt/_modules/wait.py
```diff
@@ -74,30 +79,32 @@ def _check_status(self, current):
...
         """
         Poll until all active MDS' are up:active
         """
-        cmd = json.dumps({"prefix":"status", "format":"json" })
+        cmd = json.dumps({"prefix": "status", "format": "json"})
 
         def success(status):
+            """
+            success function to passed to _check_status
+            """
```
kbaikov: I believe you mean: "Success function to be passed to _check_status". Also capitalization.

srv/salt/_modules/wait.py
jan--f: The only magic was that this code was supposed to be python 2 and 3 compatible.

srv/salt/_modules/wait.py
jan--f: Absolutely! Thanks.

srv/salt/_modules/wait.py
jschmid1: @jan--f right, that's py2/3 compatibility. That can stay

jschmid1:
Echo eric:

```
************* Module srv.modules.runners.cephprocesses
E: 20, 0: Uses of an external blacklisted import 'six': Please use 'import salt.ext.six as six' (blacklisted-external-import)
```

srv/salt/_modules/wait.py
jan--f: @jschmid1 shouldn't it be fine like it is now? since I don't actually do any work in the child constructor, it can simply be left out, resulting in the base class's constructor being called. Or amI missing something?

jan--f:
> import six as .. needs to be resolved first

Ok I was going to refer to you if thats the best way to tidy this up.

srv/salt/_modules/wait.py
jschmid1: here's the proof: http://paste.opensuse.org/35677684

jan--f: uhm thats not equivalent code. I removed the children's constructors completely. I.e. the base class' constructor will be used without me redirecting things. http://paste.opensuse.org/95356189

jschmid1: yikes, my bad. All good then.

jschmid1:
> Ok I was going to refer to you if thats the best way to tidy this up.

I think this comes from the fact that we are using salt's pylint config. I'm not against adopting their way of importing six.

`import salt.ext.six as six` is used in several other places aswell

alexandergraul:
@jan--f could you please fix the linting issue by replacing

`import six` with `import salt.ext.six as six` so we can merge this?

jan--f:
Done...sorry for the long wait guys.

PR #1166, opened 2018-05-25
Fix bare excepts in fs module and log a complete traceback

kbaikov:
Description:
Handle the concrete exceptions and uses exc_info=True in logging to capture the complete traceback

-----------------

### DIFF
--- a/srv/salt/_modules/fs.py
+++ b/srv/salt/_modules/fs.py
@@ -210,9 +210,8 @@ def _add_fstab_entry(_uuid, path, fstype, subvol):
     try:
         with open('/etc/fstab', 'r') as _fstab:
             fstab_entries = [line.rstrip('\n') for line in _fstab]
-    # pylint: disable=bare-except
-    except:
-        log.error("Failed to read /etc/fstab.")
+    except (IOError, FileNotFoundError):
+        log.error("Failed to read /etc/fstab.", exc_info=True)
         return False
 
     # Process entries.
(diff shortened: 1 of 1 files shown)


jschmid1:
Linting on this branch yields:

```
************* Module srv.salt._modules.fs
C:846, 0: Line too long (105/100) (line-too-long)
C:906, 0: Line too long (109/100) (line-too-long)
C:1105, 0: Line too long (103/100) (line-too-long)
```

kbaikov:
Fixed the long line lint. Please take a look.

srv/salt/_modules/fs.py
```diff
@@ -210,9 +210,8 @@ def _add_fstab_entry(_uuid, path, fstype, subvol):
     try:
         with open('/etc/fstab', 'r') as _fstab:
             fstab_entries = [line.rstrip('\n') for line in _fstab]
-    # pylint: disable=bare-except
-    except:
-        log.error("Failed to read /etc/fstab.")
+    except (IOError, FileNotFoundError):
```
alexandergraul: I went to check what `open` raises since I didn't know it off the top of my head and found out that nowadays it is `OSError` instead of `IOError`. `IOError` is an alias to `OSError`, so that still works, but the new one is `OSError`. I don't know if python will break the backwards compability between the two in the future, I think we are on the safer side if we go with `OSError`

kbaikov: Good point. 
How did you get an OSError?

when i do it in both versions of python i get this: http://paste.opensuse.org/48894588

alexandergraul: I did not try it out, I just read the docs at https://docs.python.org/3.5/library/functions.html#open (I wanted to link this earlier, but forgot it).

After reading your paste I tried to get an OSError in python3, but could not do it either.

kbaikov: Ah, true. In py3 the FileNotFoundError is a subclass of the OSError, but i would rather stay with a more specific exception.
I included the IOError only for py2, bacause back then the exception dependencies were a mess.

PR #1171, opened 2018-05-28
FIX issue that would make benchmark.baseline fail if an osd is removed 

plexorama:
Description:

The following error occurs if an osd is removed and benchmark is run after that:

```
admin-server:/ # salt-run state.orch ceph.benchmarks.baseline

Running osd benchmarks..............[ERROR   ] Exception occurred in runner benchmark.baseline: Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/client/mixins.py", line 395, in _low
    data['return'] = self.functions[fun](*args, **kwargs)
  File "/srv/modules/runners/benchmark.py", line 284, in baseline
    parsed_results = [ast.literal_eval(r[master_minion]) for r in results]
  File "/usr/lib64/python2.7/ast.py", line 49, in literal_eval
    node_or_string = parse(node_or_string, mode='eval')
  File "/usr/lib64/python2.7/ast.py", line 37, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 1
    Error ENOENT: problem getting command descriptions from osd.2
               ^
SyntaxError: invalid syntax

admin-server_master:
----------
          ID: osd bench
    Function: salt.runner
        Name: benchmark.baseline
      Result: False
     Comment: Exception occurred in runner benchmark.baseline: Traceback (most recent call last):
                File "/usr/lib/python2.7/site-packages/salt/client/mixins.py", line 395, in _low
                  data['return'] = self.functions[fun](*args, **kwargs)
                File "/srv/modules/runners/benchmark.py", line 284, in baseline
                  parsed_results = [ast.literal_eval(r[master_minion]) for r in results]
                File "/usr/lib64/python2.7/ast.py", line 49, in literal_eval
                  node_or_string = parse(node_or_string, mode='eval')
                File "/usr/lib64/python2.7/ast.py", line 37, in parse
                  return compile(source, filename, mode, PyCF_ONLY_AST)
                File "<unknown>", line 1
                  Error ENOENT: problem getting command descriptions from osd.2
                             ^
              SyntaxError: invalid syntax
     Started: 16:41:40.856555
    Duration: 25107.797 ms
     Changes:   

Summary for admin-server_master
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:  25.108 s
```

-----------------

### DIFF
--- a/srv/modules/runners/benchmark.py
+++ b/srv/modules/runners/benchmark.py
@@ -263,19 +263,19 @@ def baseline(margin=10, verbose=False, **kwargs):
                              'I@roles:storage and I@cluster:ceph')
     log.info('client glob is {}'.format(client_glob))
 
+    # gotta get the master_minion...not pretty but works
+    master_minion = local_client.cmd(
+        'I@roles:master', 'pillar.get',
+        ['master_minion'], expr_form='compound').items()[0][1]
+
     # get all osd ids for a given client_glob
-    osd_list = local_client.cmd(client_glob,
-        'osd.list', [], tgt_type='compound')
+    osd_list = local_client.cmd(master_minion, 'cmd.shell',
+        ['ceph osd ls'])
 
     if not osd_list:
         raise Exception('No OSDs found for glob {}'.format(client_glob))
 
-    ids = [osd_id for (osd, osd_ids) in osd_list.items() for osd_id in osd_ids]
-
-    # gotta get the master_minion...not pretty but works
-    master_minion = list(local_client.cmd(
-        'I@roles:master', 'pillar.get',
-        ['master_minion'], tgt_type='compound').items())[0][1]
+    ids = osd_list[master_minion].split("\n")
 
     sys.stdout.write('\nRunning osd benchmarks')
     sys.stdout.flush()


jschmid1:
Hey @plexorama thanks for your contribution. Much appreciated.

While 'osd.list' is a mine and should actually reflect the current state of the cluster I don't see why this patch has any downsides.

It'd be still possible to update the mines before calling a baseline benchmark though.. But still, I don't see any advantage in using a mine here. +1 from my side

swiftgist:
Without the `client_glob`, won't the benchmark always run on all minions?

jan--f:
@swiftgist is correct. The intend was to be able to run on a subset of OSDs as well.

I'd rather be in favour of deugging why ```osd.list``` seems to be broken. How can I reproduce this?

plexorama:
@jan--f to reproduce it, create a cluster with a couple osds (for example 5). then simply create a hole in the sequence, ie: remove osd.3 and there you go.

swiftgist:
@plexorama I'll bet the grains aren't updated.  If you still have your setup available, does the following cause the benchmark to succeed?

```
salt-run remove.osd 3
salt 'minionX' osd.retain
```

If that's the case, we should probably open (or maybe re-open) an issue about when to update the grains as part of other utilities.  I don't believe the current `retain` function is smart enough to delete just one entry.

kshtsk:
retest this please

jan--f:
> I don't believe the current retain function is smart enough to delete just one entry.

Agreed. We have seen issues when users execute ceph commands to remove an OSD and DeepSea misses such updates.

In any case we can go ahead with this. Fixing retain is a separate issue.

PR #1181, opened 2018-06-11
Use jq to process json instead of a chain of shell commands

kbaikov:
This fixes a fixme in 2 functions.


smithfarm:
Has anybody tested this?

kbaikov:
@smithfarm I tested it by sourcing helpers.sh directly and using the functions. And using some functions from common.sh like: test_systemd_ceph_osd_target_wants

But not as part of some suit.

PR #1185, opened 2018-06-18
Returning to previous lint version

swiftgist:
This restores master to a clean lint and working tox.ini

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,4 +1,4 @@
-# Tox (https://tox.readthedocs.io/en/latest/) is a tool for running tests
+# Tox (http://tox.testrun.org/) is a tool for running tests
 # in multiple virtualenvs. This configuration file will run the
 # test suite on all supported python versions. To use it, "pip install tox"
 # and then run "tox" from this directory.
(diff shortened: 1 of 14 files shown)


tox.ini
```diff
@@ -9,30 +9,23 @@ envlist = py27,lint
...
     salttesting
     configobj
     boto
-    codecov
-    pytest-cov
 
 [testenv:lint]
-basepython = python3
-commands = pylint --rcfile=.pylintrc  srv/
+commands = pylint --rcfile=.pylintrc --jobs=5 srv/
```
kbaikov: I believe jobs parameter is redundant here since it is already in pylintc

tox.ini
```diff
@@ -9,30 +9,23 @@ envlist = py27,lint
 [base]
 deps =
     ipaddress
-    salt
+    salt<2017.0.0
 
 [testenv:py27]
-basepython = python2.7
```
kbaikov: Why are you removing explicit py versions from here and from lint?

kshtsk:
retest this please

jschmid1:
Running make lint on the current master gives me only one single error.

```
************* Module srv.salt._modules.deepsea
W: 14,20: Unused variable 'state2' (unused-variable)
```

tox.ini
swiftgist: I simply pulled this from SES5 in order to get environments working.  If we have been moving parameters from command line to config file, I am not aware.

tox.ini
swiftgist: Same as above.  It's not deletion as much as restoring to a known working state.

swiftgist:
Since #1186 works for me, this is not needed.

PR #1188, opened 2018-06-19
Remove creation of unused partitions, implement recommendation

swiftgist:
The bluestore partition process has been working from original
assumptions.  The refactor of bluestore args created a disconnect
between the two steps.  Partitions are getting created, but never
used nor removed.

This also implements the recommendation to use a single DB partition
since the WAL will naturally use an existing DB.  This reduces the
number of managed partitions from two to one for each OSD.  Note
that osd.report is intentionally not changed to give the admin an
indication that the configuration can be improved.

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -898,14 +898,16 @@ class OSDPartitions(object):
                     log.warning(("WAL size is unsupported for same device of "
                                  "{}".format(self.osd.device)))
                 else:
-                    # Create wal of wal_size on wal device
-                    self.create(self.osd.wal, [('wal', self.osd.wal_size)])
+                    log.info(("WAL will reside on same device {} as db - "
+                              "recommend removing the WAL entry from the "
+                              "configuration for device "
+                              "{}").format(self.osd.db, self.osd.device))
             else:
                 # pylint: disable=line-too-long
                 log.warning("No size specified for wal {}. Using default sizes.".format(self.osd.wal))
 
             if self.osd.db_size:
-                if self.osd.wal == self.osd.device:
+                if self.osd.db == self.osd.device:
                     # pylint: disable=line-too-long
                     log.warning("DB size is unsupported for same device of {}".format(self.osd.device))
                 else:
@@ -914,25 +916,22 @@ class OSDPartitions(object):
             else:
                 log.warning("No size specified for db {}. Using default sizes".format(self.osd.db))
         else:
-            # This situation seems unintentional - use faster media for
-            # the wal or db but not the other.  Help newbies out by
-            # putting wal and db on same device
+            # Only WAL
             if self.osd.wal:
                 if self.osd.wal_size:
                     if self.osd.wal == self.osd.device:
                         log.warning(("WAL size is unsupported for same device of "
                                      "{}".format(self.osd.device)))
                     else:
-                        log.warning("Setting db to same device {} as wal".format(self.osd.wal))
+                        log.info("DB will reside on device {}".format(self.osd.device))
                         # Create wal of wal_size on wal device
-                        # Create db on wal device
-                        self.create(self.osd.wal, [('wal', self.osd.wal_size),
-                                                   ('db', self._halve(self.osd.wal_size))])
+                        self.create(self.osd.wal, [('wal', self.osd.wal_size)])
             else:
                 if self.osd.wal_size:
                     log.warning(("WAL size is unsupported for same device of "
                                  "{}".format(self.osd.device)))
 
+            # Only DB
             if self.osd.db:
                 if self.osd.db_size:
                     if self.osd.db == self.osd.device:
@@ -941,9 +940,7 @@ class OSDPartitions(object):
                     else:
                         log.warning("Setting wal to same device {} as db".format(self.osd.db))
                         # Create db of db_size on db device
-                        # Create wal on db device
-                        self.create(self.osd.db, [('wal', self._double(self.osd.db_size)),
-                                                  ('db', self.osd.db_size)])
+                        self.create(self.osd.db, [('db', self.osd.db_size)])
             else:
                 if self.osd.db_size:
                     log.warning(("DB size is unsupported for same device of "
@@ -1173,11 +1170,17 @@ class OSDCommands(object):
 
             # WAL cornercase with sizes
             if self.osd.wal and self.osd.wal_size and self.osd.wal != self.osd.device:
-                _partition = self.highest_partition(self.osd.wal, 'wal')
-                if _partition:
-                    args += "--block.wal {}{} ".format(self.osd.wal, _partition)
+                if self.osd.db:
+                    log.warning(("Ignoring WAL setting - "
+                                 "No need for two partitions, "
+                                 "WAL will use same device as DB "
+                                 "{}").format(self.osd.db))
                 else:
-                    args += "--block.wal {} ".format(self.osd.wal)
+                    _partition = self.highest_partition(self.osd.wal, 'wal')
+                    if _partition:
+                        args += "--block.wal {}{} ".format(self.osd.wal, _partition)
+                    else:
+                        args += "--block.wal {} ".format(self.osd.wal)
 
             # DB cornercase with sizes
             if self.osd.db and self.osd.db_size and self.osd.db != self.osd.device:
--- a/tests/unit/_modules/test_osd.py
+++ b/tests/unit/_modules/test_osd.py
@@ -814,8 +814,7 @@ class TestOSDPartitions():
         And I have a wal_size
         And I have a db_size
         And wal is not equivalent to the device
-        Expect to call create('/dev/sddb', [('db', db_size)])
-        Expect to call create('/dev/sdwal', [('wal', wal_size)])
+        Expect to call create('/dev/sddb', [('db', db_size)]) only
         """
         kwargs = {'format': 'bluestore',
                   'wal': '/dev/sdwal',
@@ -826,8 +825,8 @@ class TestOSDPartitions():
         test_module = helper_specs(module=DEFAULT_MODULE)()
         obj = test_module.OSDPartitions(osd_config)
         obj._bluestore_partitions()
-        create_mock.assert_any_call('/dev/sdwal', [('wal', 'walsize')])
         create_mock.assert_any_call('/dev/sddb', [('db', 'dbsize')])
+        assert create_mock.call_count == 1
 
     @mock.patch('srv.salt._modules.osd.OSDPartitions.create')
     @mock.patch('srv.salt._modules.osd.log')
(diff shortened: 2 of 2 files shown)


swiftgist:
This PR is a prerequisite for the wip-migrate branch.  The remove/replace smoketests can fail since the repetitive nature destroys and creates OSDs a few times.  The accumulation of orphaned partitions causes the smoketest to fail.

I will ask Igor to review this change as well since the changes are based on my discussions with him with regards to the preferred method of configuring bluestore on separate devices.

swiftgist:
Does susebot run every time somebody makes a comment?

@tserong You can still have data, separate wal, separate db (i.e. 3 devices), but if only using two devices with wal and db on the same device, we have only been making our lives harder.  There's no performance difference.  If the wal ran out of space, it still spilled over to the db.  And we had two partitions to manage per OSD.

This will give a management aspect similar to filestore.  Is my journal on the same device or different device.  Specifying just the db is apparently enough for the wal to use the db as well.  So, conversations are easier.  One partition per OSD on a shared device should be simpler.

We still keep the wal by itself option on the odd chance somebody has really small SSDs and is willing to keep the db on the main disk for a specific hardware setup.

srv/salt/_modules/osd.py
```diff
@@ -898,8 +898,10 @@ def _bluestore_partitions(self):
                     log.warning(("WAL size is unsupported for same device of "
                                  "{}".format(self.osd.device)))
                 else:
-                    # Create wal of wal_size on wal device
-                    self.create(self.osd.wal, [('wal', self.osd.wal_size)])
+                    log.info(("WAL will reside on same device {} as db - "
```
ifed01: Perhaps I missed something but I don't see the difference in handling following two cases: 
1) wal == db 
2) wal != db
IMO both are getting here but there is no wal creation for case 2)

srv/salt/_modules/osd.py
```diff
@@ -898,8 +898,10 @@ def _bluestore_partitions(self):
...
+                              "recommend removing the WAL entry from the "
+                              "configuration for device "
+                              "{}").format(self.osd.db, self.osd.device))
             else:
                 # pylint: disable=line-too-long
                 log.warning("No size specified for wal {}. Using default sizes.".format(self.osd.wal))
```
ifed01: Unrelated to this PR but just in case - are this line and similar handling for absent db_size relevant? It looks like respective devices wouldn't be created at all rather than "using default sizes".

srv/salt/_modules/osd.py
swiftgist: Right, the wal would not be created in either case, but the reason is different for the admin.  The first message is that ceph-disk does not support specifying a wal on the same device as the OSD.  The second is that it's unnecessary.

srv/salt/_modules/osd.py
swiftgist: If a size is not specified, we cannot create a partition.  However, ceph-disk will accept a device without the partition and create partitions itself.  So, the "default sizes" are for what ceph-disk will use.

kshtsk:
@swiftgist No,  susebot does not do that on each comment, I have triggered the test job for this PR again on jenkins.

kshtsk:
retest this please

jschmid1:
@jan--f @tserong please also review this PR. It's the prerequisite of #1216

tserong:
> You can still have data, separate wal, separate db (i.e. 3 devices), but if only using two devices with wal and db on the same device, we have only been making our lives harder.

It reads to me as if you can have three separate devices, but only if you don't specify wal_size or db_size.

If either wal_size or db_size are specified, it seems to be forcing the wal onto the same device as the db.

Is that correct?

tserong:
Also, see https://github.com/SUSE/DeepSea/blob/2e465f078dc03f5cfa5b5238754bedbdcf80bdaf/srv/salt/_modules/osd.py#L910 -- Judging from the shape of the code around this, I think that's intended to be `if self.osd.db == self.osd.device`, not `if self.osd.wal == self.osd.device`.

swiftgist:
Fixed line 910.  Also, fixed the indent, removed old comments and wal creation when no wal is defined without sizes.

swiftgist:
With respect to no sizes, the `_bluestore_partitions` does nothing.  The `_bluestore_args` will still pass the data, wal and db as three separate arguments and let ceph-disk create the partitions itself.

tserong:
What about if someone wants to specify sizes for both db and wal, but wants to have three separate devices? That seems to be not possible. Or is that not a sensible thing for someone to want? Or, do I still not quite understand the logic? :-)

swiftgist:
No, that would work too.  With sizes, the `_bluestore_partitions` creates a WAL partition and a DB partition on their respective devices and those specific devices are then passed to ceph-disk.  The big difference is whether we pass `/dev/sdb` or `/dev/sdb3` to ceph-disk for a WAL or DB.  When we create the partition (since we have a size), we pass the latter.  

BTW, the `osd.prepare` and `osd.activate` are non-destructive ways to experiment with the configuration generating the ceph-disk commands.

jschmid1:
@tserong are we good to merge?

tserong:
> @tserong are we good to merge?

With the disclaimer that I haven't yet been able to test multi-device OSDs myself, yes, I'm happy this has received enough scrutiny (this one just hurt my brain trying to hold it in my head, for some reason :-/)

jschmid1:
I just tried the following setup:

``` yaml


ceph:
  storage:
    osds:
      /dev/vdb:
        format: bluestore
        wal: /dev/vde
        db: /dev/vdf
        db_size: 1G
        wal_size: 2G
      /dev/vdc:
        format: bluestore
        wal: /dev/vde
        db: /dev/vdf
        db_size: 4G
        wal_size: 3G
      /dev/vdd:
        format: bluestore
        wal: /dev/vde
        db: /dev/vdf

```

which results in:


```
vda    253:0    0   40G  0 disk
└─vda1 253:1    0   40G  0 part /
vdb    253:16   0   20G  0 disk
├─vdb1 253:17   0  100M  0 part
└─vdb2 253:18   0 19.9G  0 part
vdc    253:32   0   20G  0 disk
├─vdc1 253:33   0  100M  0 part /var/lib/ceph/osd/ceph-14
└─vdc2 253:34   0 19.9G  0 part
vdd    253:48   0   20G  0 disk
└─vdd1 253:49   0  100M  0 part
vde    253:64   0   20G  0 disk
vdf    253:80   0   20G  0 disk
├─vdf1 253:81   0    1G  0 part
├─vdf2 253:82   0    4G  0 part
└─vdf3 253:83   0    1G  0 part

```

And the following logs:

```
data1:~ # salt-call osd.deploy
[ERROR   ] Partition type lockbox not found on /dev/vdb
[ERROR   ] Partition type lockbox not found on /dev/vdc
[ERROR   ] Partition type lockbox not found on /dev/vdd
[ERROR   ] Partition type osd not found on /dev/vdd
[WARNING ] No size specified for wal /dev/vde. Using default sizes.
[WARNING ] No size specified for db /dev/vdf. Using default sizes

```

Whilst osd.report still complains:

```
Do NOT use this appliance for production deployments!
data1:~ # salt-call osd.report
local:
    No OSD configured for 
    /dev/vdb
    /dev/vdd
    Different configuration for 
    /dev/vdc

```

I'm not too convinced that this is the right way to go. We more or less silently forbid certain configurations that most users were anticipate to be correct.

jschmid1:
Comment is referencing my last comment above:

I managed to get /dev/vdb and /dev/vdc to deployed as expected ( only one partition gets created on the db: /dev/vdf respectively )

Potential issues:
  1) I set db_size: 1G which is fine for only the DB probably.. but as we are now colocating the WAL and the DB on the same partition 1G might not be enough.  
   _( Assuming the resulting wal actually points to the DB, although I can't find a link to the wal/db in /var/lib/ceph/ceph-osd/$ID )_

  2) I set /dev/vdd to not have specific sizes which resulted in:
    `ceph-disk -v prepare --bluestore --data-dev --journal-dev --cluster ceph --cluster-uuid dd52dd7b-17e7-44f1-9739-c043d15b4481 --block.wal /dev/vde -- block.db /dev/vdf /dev/vdd` 
   **(note the --block-wal )**
   That also failed with http://paste.opensuse.org/70180931 ValueError in ceph-disk

  3) @tserong  was asking:
      > What about if someone wants to specify sizes for both db and wal, but wants to have three separate 
      > devices? That seems to be not possible. Or is that not a sensible thing for someone to want? Or, do I 
      > still not quite understand the logic? :-)
 
       That statement seem to hold if I lock at the osd.prepare outpout of (2)

--------------


This PR is considered to be a prerequisite for #1216 with the argument:

> This PR is a prerequisite for the wip-migrate branch. The remove/replace smoketests can fail since the repetitive nature destroys and creates OSDs a few times. The accumulation of orphaned partitions causes the smoketest to fail.

iirc the smoketests use [remove.osd](https://github.com/SUSE/DeepSea/blob/8ce73c29e1227913cf511894b4223d140c7b8be0/srv/salt/ceph/tests/migrate/remove_osds.sls) which should take care of orphaned partitions? In case we have orphaned partitions there is a bug in the removal.

If we don't have enough space on the partitions in the first place we should think about increasing the size of the disks on our testing machines. ( qcow2 is a sparse disks and does not take up space anyways.. )

---------

I'm still not convinced that we should try to mangle a user-created configuration. Imo we should exactly do what the user told deepsea to do. If he went for it, he shall have it. I'm not against a warning message that tells him that this is not the most efficient configuration to go for, but we do what he asked for.

I'd like to hear your thoughts on that @tserong @swiftgist @smithfarm

swiftgist:
@jschmid1 The bug isn't in the removal.  The removal cannot wipe all partitions on a secondary device and can only remove the ones referenced.  The bug is in the partition creation currently because it creates an unreferenced partition.

The rest was to bring the behavior expected/desired by recent talks on the mailing lists.

I would have to look closer, but the spacing between `--` and `block.db` looks off in the 3 device command with no sizes.  That behavior should happen with master as well then since the no size conditions are the same.

jschmid1:
> The bug isn't in the removal. The removal cannot wipe all partitions on a secondary device and can only remove the ones referenced.

Forgive me my naivety, but isn't that exactly what we need? We need to remove the partitions that the passed OSD is pointing to. Why do we need to remove _all_ partitions on a secondary device?

 

> The bug is in the partition creation currently because it creates an unreferenced partition.

In which circumstances?




> The rest was to bring the behavior expected/desired by recent talks on the mailing lists.

right, I see that. I'd still say we either stick to the old behavior or don't 'advise' anything at all.



> I would have to look closer, but the spacing between -- and block.db looks off in the 3 device command with no sizes.

The space have sneaked in during copy/paste.



> That behavior should happen with master as well then since the no size conditions are the same.

with #1244 we will get a better picture of what is currently works and what doesn't.

jschmid1:
any progress @swiftgist ?

swiftgist:
Closing.  Created https://github.com/SUSE/DeepSea/tree/fix-extra-partition

jschmid1:
@swiftgist is that merged into some exsting/open PR?

PR #1194, opened 2018-06-22
Fix ceph_version

### LINKED ISSUE #1104: ceph_version takes 56 seconds

swiftgist:
Changed serial salt call to a single call.  Refactored as well.  Added
conditional bypass to skip checks on working clusters.

A new ci_validate.py runner still gives the original behavior but is part of
the qa package.

Fixes #1104

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -56,12 +56,20 @@ class PrettyPrinter(object):
     """
 
     # pylint: disable=unused-argument
-    def add(self, name, passed, errors, warnings):
+    def add(self, name, skipped, passed, errors, warnings):
         """
-        Print colored results.  Green is ok, yellow is warning and
-        red is error.
+        Print colored results.  Green is ok, yellow is warning,
+        red is error and blue is skipped.
         """
         # Need to make colors optional, but looks better currently
+        for attr in skipped.keys():
+            format_str = "{:25}: {}{}{}{}".format(attr,
+                                                  Bcolors.BOLD,
+                                                  Bcolors.OKBLUE,
+                                                  skipped[attr],
+                                                  Bcolors.ENDC)
+            log.info("VALIDATE SKIPPED  %s", format_str)
+            print(format_str)
         for attr in passed.keys():
             format_str = "{:25}: {}{}{}{}".format(attr,
                                                   Bcolors.BOLD,
@@ -216,23 +224,29 @@ class Validate(Preparation):
     """
 
     def __init__(self, name, search_pillar=False, search_grains=False,
-                 printer=None, search=None):
+                 printer=None, search=None, skip_init=False):
         """
         Query the cluster assignment and remove unassigned
         """
         # Python2 syntax is only used because the test is running in python2
-        super(Validate, self).__init__()
+        if not skip_init:
+            super(Validate, self).__init__()
         self.name = name
         self.data = self.__get_items(search_pillar, 'pillar')
         self.grains = self.__get_items(search_grains, 'grains')
         self.printer = printer
         self.in_dev_env = self.__dev_env()
+        self.skipped = OrderedDict()
         self.passed = OrderedDict()
         self.errors = OrderedDict()
         self.warnings = OrderedDict()
         if search:
             self.search = search
         # self._minion_check()
+
+        # Ceph version
+        self.package = 'ceph-common'
+        self.uninstalled = []
 
     def __get_items(self, enabled, target):
         """
@@ -272,6 +286,12 @@ class Validate(Preparation):
         """
         if key not in self.errors and key not in self.warnings:
             self.passed[key] = "valid"
+
+    def skip(self, key):
+        """
+        Assign skipped steps
+        """
+        self.skipped[key] = "skipping"
 
     def dev_env(self):
         """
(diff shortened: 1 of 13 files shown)


smithfarm:
I'm unsure why anyone would need a CI-specific runner - I don't think it will be used. The integration/regression tests need to exercise the same code that users use.

swiftgist:
Now that a month has gone by, I had an idea from the wip-migrate branch.  I created a runner specifically for smoketests that now lives in the qa package.  I thought the same strategy may work here with the creation of a `ci_validate`.

I do like this for a few reasons.

1) Continuous Integration requirements will not be conflated with DeepSea requirements.  For example, the validation of the ceph version only needs to see that the versions are equal and present to be useful to the admin.  If something is wrong, then the admin will have to investigate.  This is true regardless of the particular failure.  

2) Parsing zypper error messages is something that must be maintained.  Frankly, I feel that the related Salt packages (i.e. pkg.* for zypper) should be improved instead.  Still, I understand that zypper can fail for many different reasons related to missing repos, repos with incorrect packages, etc. and that in a CI environment, the additional messages are an efficiency for debugging.

3) The side effect of supportable code is less.  We are implementing unit tests and smoketests for all code in the DeepSea package.  Separating requirements means less code to unit test and smoketest.

***
On a separate note, the reason that this issue became a priority is that our Stages are taking longer.  I am sensitive to how many checks and timeouts we have currently.  While testing, I found that my time dropped from 56 seconds to 25 seconds in a fresh cluster for this validation, but the second time should stay roughly consistent regardless of the size of the cluster.  I then realized that we are making those running DeepSea clusters repeatedly run these checks for a working cluster.  I believe that is not necessary.  That is why I added the additional bypass and skipping logic.  The admin is informed, but the check that does nothing takes 4 seconds.  (The cephprocesses.mon takes that long.)

swiftgist:
@smithfarm I was still writing the above post when you commented.  See if you agree.

smithfarm:
@swiftgist Thanks for the explanation, but I don't really agree :-) I will continue to try to write integration/regression tests that exercise as much of the real code as possible.

That said, I don't have any problem with your refactor of validate.py. I would just as soon not have ci_validate.py since I don't think anyone will use or maintain it.

smithfarm:
One more note - having a specific runner for smoke tests makes sense I think. Whatever it takes to make them easy to run? The integration tests, on the other hand, are designed to automate entire deployments always starting from a given, well-defined point like "fresh OS install".

smithfarm:
In conclusion, I don't think there is any "CI testing requirement" that is important enough to warrant writing and maintaining CI-specific code that replaces actual DeepSea code, since going that route has the IMO undesirable consequence that whatever DS code is being so replaced sees no integration testing.

smithfarm:
@swiftgist I can only repeat that the integration testing scripts must use validate.py. Anything that reduces code coverage of the tests tends to defeat the very raison d'etre of these tests.

I do remember arguing vigorously for inclusion, in validate.py, of some of the code you dropped in the refactor, but there is simply no point, at the integration/regression test level, in testing code that is not used in production.

Now, regarding the test scripts you asked about. . . 

All the integration test scripts are testing DeepSea's ability to deploy clusters in various configurations that, if I'm not mistaken, are supported by the SES product - i.e. automated integration/regression testing of these scenarios is needed at the product level. Any time DeepSea is used to deploy a cluster, non-DeepSea code (whether in Salt or other packages in openSUSE/SLE) is exercised along with DeepSea code. (In other words, if the tests identify a regression, the regression could be in any component, not only in DeepSea.) This holds true for all the integration test scripts. 

In short, the purpose of the integration test scripts is to test entire product deployment scenarios from start (defined as "fresh OS install") to finish. All the scripts are tightly integrated to avoid code duplication. So it sounds like we should move all the test scripts to a different code repo, since none of them are testing DeepSea exclusively?

srv/modules/runners/validate.py
```diff
@@ -771,66 +792,78 @@ def ceph_version(self):
...
+        install only happens once.
+        """
+        search = __utils__['deepsea_minions.show']()
+        results = self._silent_search(search, 'pkg.info_installed')
+
+        for minion in results:
+            if isinstance(results[minion], dict) and self.package in results[minion]:
+                if 'version' in results[minion][self.package]:
+                    version = self._check_version(minion, 'pkg.info_installed',
+                                                  results[minion][self.package]['version'])
```
jschmid1: if version is set to `""`, which happens in case of an error in the retrieved data see `check_version`

``` python
if version.split('.')[0].isdigit():
    return version
log.debug("VALIDATE {} on minion {} returns {}".format(func, minion, version))
    return ""
```

The comparison of `LooseVersion`s will fail.

srv/modules/runners/validate.py
```diff
@@ -771,66 +792,78 @@ def ceph_version(self):
...
+        if not self.uninstalled:
+            return
+
+        search = "L@{}".format(",".join(self.uninstalled))
+        results = self._silent_search(search, 'pkg.info_available')
+        for minion in results:
+            if isinstance(results[minion], dict) and self.package in results[minion]:
+                if 'version' in results[minion][self.package]:
+                    version = self._check_version(minion, 'pkg.info_available',
+                                                  results[minion][self.package]['version'])
```
jschmid1: same remark as in https://github.com/SUSE/DeepSea/pull/1194/files#r197745691

srv/modules/runners/validate.py
```diff
@@ -1141,9 +1174,22 @@ def prep(**kwargs):
...
     """
     printer = get_printer(**kwargs)
+    if ('bypass' in kwargs and kwargs['bypass'] and
+        __salt__['cephprocesses.mon']()):
+        # Disable all Salt lookups
+        valid = Validate("setup", search_pillar=False, search_grains=False,
+                         skip_init=True, printer=printer)
+        valid.skip('deepsea_minions')
+        valid.skip('master_minion')
+        valid.skip('ceph_version')
```
jschmid1: If we disable those after the initial deployment, we have no way to identify if an update brought in a ceph version that we don't support. For salt_version we have the RPM that is supposed to conflict with any unsupported version.

jschmid1:
Regarding the topic of 'testing real code', I'm with Nathan here. The tests are supposed to find issues with the code that a user is running and not something we made up. That also means that a test should not rely on a debug message( referring to https://github.com/SUSE/DeepSea/pull/1109#pullrequestreview-114521594 ).

I'd vote for a proper solution where the tests exercise as much 'real' code as possible. Is there any missing requirement that'd make the validate.py unusable for the CI/tests @smithfarm.


--------

+1 for the rework on validate.py. ( see comments )

swiftgist:
Yeah, the `_check_version` should not be returning "".  While a `return "0"` would fix the logic, the message would be misleading.  Undecided at the moment on the cleanest path forward.

swiftgist:
I added a message to `_check_version` and altered the two conditions.  I think *malformed* is a decent catchall.

***
A clarification about the ci_validate.py.  I did not intend for that to be run by the CI instead of the normal validate.py.  My intention is that it can be run "in addition" to give the CI maintainers the logs desired that are not necessarily useful to the admins.

If this is not desired, I can remove the ci_validate.py.

denisok:
I think there might be a misunderstanding on testing types.

It might be that testing @swiftgist is talking about is a functional testing, that could be run on already configured cluster or without a cluster a all some time.

@smithfarm is all about integration testing on the system level, when we want to test all peaces together.

I know nothing about this particular PR though, I don't know how to correctly separate (if that needed at all) two of those.

swiftgist:
After looking over the code, all the errors are logged via the add method.  I addressed the missing condition for `_check_version`, removed the ci_validate.py since it's useful to nobody and edited the title.

I also refactored the validate step into its own orchestration file.  The admin can get the original behavior to always run the checks for Stage 0 and 1.  The initial runs will look as expected

```
deepsea_minions          : valid
master_minion            : valid
salt_version             : valid
```

Subsequent runs will be faster and appear as
```
deepsea_minions          : skipping
master_minion            : skipping
ceph_version             : skipping
salt_version             : skipping
```

tserong:
> The admin can get the original behavior to always run the checks for Stage 0 and 1

By setting stage_validate=default-nobypass? I wonder if maybe it would be best to run all checks always on stage 0, but have them auto-skip on stage 1 for running clusters?

swiftgist:
Okay, I'll go with that.  I had to turn it over a few times. :)

Besides the fresh deployment, Stage 0 should normally only be necessary when adding new minions.  Stage 1 is necessary when adding new minions, additional disks or replacing disks.  My goal with the bypass is to match the original intention of reducing the time of checks.  The checks start to be superfluous at a certain point, but I suspect that's subjective.

Admins will see that Stage 1 skips the check with a working cluster and that may be enough incentive to ask for other checks to be treated similarly.  

I avoided using

```
include:
  - ...validate.default-nobypass
```
since that cannot be overridden trivially.  The one level of indirection does not benefit the admin.

PR #1195, opened 2018-06-25
[RFC] Trigger reactor on inode events

jschmid1:
Description:

In the past we encountered quite some issues with 'missing modules' which could've been solved by running salt \* saltutil.sync_all. To avoid the resulting frustration we can use a beacon to monitor the relevant directories ( /srv/salt/_modules and /srv/salt/_states ) for modifications and creations and trigger the needed `saltutil.sync_all` using salt's reactor system.

-----------------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,9 @@ pyc: setup.py
 	find cli/ -name '*.py' -exec python3 -m py_compile {} \;
 
 copy-files:
+	# salt-minion config files
+	install -d -m 755 $(DESTDIR)/etc/salt/minion.d
+	install -m 644 etc/salt/minion.d/beacons.conf $(DESTDIR)/etc/salt/minion.d/
 	# salt-master config files
 	install -d -m 755 $(DESTDIR)/etc/salt/master.d
 	install -m 644 etc/salt/master.d/modules.conf $(DESTDIR)/etc/salt/master.d/
--- a/deepsea.spec.in
+++ b/deepsea.spec.in
@@ -397,6 +397,7 @@ systemctl try-restart salt-api > /dev/null 2>&1 || :
 %{_mandir}/man5/deepsea*.5.gz
 %{_mandir}/man1/deepsea*.1.gz
 %config(noreplace) %attr(-, salt, salt) /etc/salt/master.d/*.conf
+%config(noreplace) %attr(-, salt, salt) /etc/salt/minion.d/*.conf
 /srv/modules/modules/*.py*
 /srv/modules/runners/*.py*
 /srv/modules/utils/*.py*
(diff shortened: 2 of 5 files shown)


jschmid1:
> Technically, we only need to trigger the reactor under the right circumstances. We could send an event after the salt-master restart in both the postinstall of the rpm and in the "make install". This would be explicit to the admin (i.e. there's a command) and should be easier to debug for other developers. We would also have only a single event per install.

I'd be not against that solution as the prerequisite package for the inode beacon is not present in our repositories. Currently this comes from pip.

jschmid1:
> Hmm, I'm mixed on this one. Having this fire for every touch of /srv/salt could lead to exciting things where a "chown -R salt:salt /srv/salt" by an admin triggers on onslaught of "sync_all" jobs. If Salt experienced failures or side effects, the admin will have fun debugging.

Yeah, I also though about that. There is the `coalesce` option that avoids duplicates, but those wouldn't be duplicates.

jschmid1:
> The beacons would be new for DeepSea. That's one issue in the cost of somebody new having to learn one more Salt component. (I know it's inevitable to use more Salt components, but I do not think this use case is compelling enough for beacons.) The other is that I think they are easy to get wrong. If I knew some other automation is the only thing that writes to /var/run/somedirectory, then a beacon might be the right thing. For /srv/salt (or even changing it to /srv/salt/_modules, /srv/salt/_states), we have admins touching things, package installers changing files and possibly other automation.

I personally don't have any concerns adding new functionality which runs in the background/ is not user facing. The beacon/reactor could help us solve some more problems we have. The first thing that comes to my mind would be the /var/run/cache/salt/master/jobs directory that can be sanitized asynchronously -> faster stage executing times.

jschmid1:
> One of the problems/side effects with the reactor is that Salt cannot detect loops like it can with the compiled Salt states. I will say it is hard to track down the death spiral when a second reactor file is triggering a beacon that kicks off the first reactor file which calls a stage, etc.

I'm not saying it's not dangerous :)

jschmid1:
> Different topic: I really think this should be an issue rather than a PR. The same discussions would happen. The difference is build automation does not need to run against this and notify everyone. Also, every PR that is not ready/will not be merged adds to the list for everyone. I think converting 190+ issues to PRs and having issues only represent bug reports would hinder communication more than help it.

No, I don't think so. There is code attached to it so that people can see how things would be done theoretically and comment on specific lines. The build automation runs exactly one time which is managable.

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
retest this please

PR #1198, opened 2018-06-25
[backport][SES5] Add tuned profiles to Deepsea

mogeb:
Stage 3 now deploys tuned profiles, and activates the profiles ses-mgr,
ses-mon and ses-osd. Precedence goes to ses-osd, then ses-mon and
finally ses-mgr.

Signed-off-by: Mohamad Gebai mgebai@suse.com


swiftgist:
Since we do not have a separate issue for discussions about susebot, I will litter this PR with an off-topic issue.  Apologies to @mogeb.

I would like the susebot successes to look like the above failure.  That is, short and to the point.  I do not need a wall of text telling me how successful the tests were.  Think Unix or Linux.  Success is quiet.

Now with regard to the error, I think "Some failures" can be improved on this one since it's a common one for developers that do not run "make rpm".  Specifically, build failure with the "Installed (but unpackaged) file(s) found" would get more to the point.  

Ideally, this check could be similar to what it was before and live in the branch region and not be a post, but I don't know how difficult Github Apps are to implement.  If it were, then displaying the dozen files missing from the spec file might get things fixed quicker.  When the build succeeds, the check would change back to a success (or the next failure.)  As a post, I'd rather not have the historical issues of passing the tests distracting from the conversations about the PR itself.

srv/salt/ceph/rescind/tuned/default.sls
```diff
@@ -0,0 +1,3 @@
+uninstall tuned:
+    pkg.removed:
```
mogeb: Should we remove tuned?

srv/salt/ceph/smoketests/tuned/mgr/init.sls
```diff
@@ -0,0 +1,6 @@
+Check tuned for ses-mgr roles:
+  salt.state:
+    - tgt: "I@roles:mgr and not I@roles:storage and not I@roles:mon"
+    - tgt_type: compound
+    - sls: ceph.tests.tuned.mgr
```
mogeb: The problem with the smoke tests I have added is that they are targeted. The test `ceph.tests.tuned.mgr` is expected to fail on nodes where there's an OSD (even if there's also a mgr there). The `tgt` here takes care of that, but if one was to run `ceph.tests.tuned.mgr` manually and targets all nodes (or all nodes that have a mgr), he'll see failures and think that there's something wrong. Thoughts on this? Will people run the tests manually without going through `ceph.smoketests.tuned.*`? If they do, maybe adding a message or a comment in the sls file is enough?

mogeb: The alternative is to have the tests look for which roles exist on a node, and determine which profile it should have.

srv/salt/ceph/rescind/tuned/default.sls
swiftgist: I think this is fine since we install it.  Also, if the one admin that is using DeepSea to manage their Ceph cluster also created their own profiles, then the init.sls override is there to disable the uninstall of tuned.

swiftgist:
@susebot Why are there so many runs and successes followed by failures for code that has not changed?  If we need a test PR, then coordinate.

mogeb:
@swiftgist I've updated the PR, you can take it for a spin. To run the tests:

```
salt-run state.orch ceph.smoketests.tuned
```

To change tuned profiles:
```
tuned-adm profile <profile name>
```
Where `<profile name>` can be `ses-osd`, `ses-mon` or `ses-mgr`. To disable tuned:
```
tuned-adm off
```
To enable tuend:
```
tuend-adm profile <profile name> 
```
To view the current profile:
```
tuned-adm active
```

denisok:
@swiftgist github plugin watches for the branch for new commits, and apparently he found some (here few examples):
https://github.com/mogeb/DeepSea/commit/393f2fbec4af587635643b7a4e605dd5b6cf48a8
https://github.com/mogeb/DeepSea/commit/5d3a1ffee23d4f9aff8c871ed7f23daa453c6170

you could see commit hashes in the job description in each job.

But now those commit are gone from the branch tree, probably reset, squashed or whatever I don't know. Lets ask @mogeb - what are those commit? have you committed and pushed them to the branch but then hard reset?

It also might be a bug in the github plugin, but lets check with @mogeb first.

mogeb:
@denisok that is correct. Upon finding minor issues I rewrote history to this branch (rebase + push -f).

swiftgist:
For anybody reading along, please understand that it's perfectly fine to make incremental fixes and rebases as needed.  We will be addressing the susebot reporting soon.

swiftgist:
I had to install libgio-2_0-0.  It seems like tuned is missing a dependency, but adding to ceph/packages/common/default.sls will work too.

Would you put ses-mon, ses-mgr and ses-osd under ../tuned/files?  That would match the Salt convention of configuration files and templates.

The smoketest always works.  :)   I'm wondering though, if we need an "tuned-adm off".   That would give a trivial method for disabling an individual profile.  Maybe a `default-off.sls` for each config.  This would allow a second smoketest as well to toggle each profile to detect if some future somebody breaks the on/off functionality of the states.

The only question that I do not know the correct answer is what to do about the remaining roles.  Should the smoketest be extended to detect that the other roles have "No current active profile" ?

kshtsk:
retest this please

kshtsk:
retest this please

mogeb:
@jschmid1 I cherry-picked the latest changes from #931, the PR is ready for review.

jschmid1:
@susebot run teuthology

jschmid1:
I only saw one infrastructure related issue. We should be good here

kshtsk:
@jschmid1 OVH is not in a good mood today

mogeb:
@susebot  run teuthology

PR #1204, opened 2018-06-27
Drop scsi prefixed devices

### LINKED ISSUE #1203: unreliable scsi symlinks
With #952 and #1061, DeepSea started selecting other symlinks.  However, the udev rules from 58-scsi-sg3_symlink.rules do not create persistent and unique names.  On hardware with these devices, zapping an OSD causes the device name to be given to another device.

"I obviously cannot choose the wine in front of you" - Vizzini, Princess Bride

After examining 58-scsi-sg3_symlink.rules and 60-persistent-storage.rules, I believe there is no reason to use any of the `scsi-` prefixed names.  The most reliable symlink would be the `scsi-3` prefixed devices but these are redundant with `wwn-0x` devices.

Rather than have complicated logic including or excluding particular expressions of devices, removing the `scsi-` prefix entirely will simplify the entire process.  For any hardware that has an `ata-` prefix, that will get chosen.  Otherwise, the `wwn` will be selected.

swiftgist:
The udev rules in 58-scsi-sg3_symlink.rules will produce symlinks
that are not persistent nor unique.  Rely on the remaining devices
such as ata* and wwn*.

Fixes #1203 
bsc#1097914

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -1773,10 +1773,15 @@ class OSDDevices(object):
     def _uuid_device(self, device, pathname="/dev/disk/by-id"):
         """
         Return the uuid device, prefer the most descriptive
+
+        Note for future refactor:  Consider using
+          udevadm info -q symlink -n /dev/sdb
+        for discovering alternate device names.  The cephdisks and
+        proposal module share similar functionality.
         """
         if os.path.exists(device):
             if os.path.exists(pathname):
-                cmd = (r"find -L {} -samefile {} \( -name ata* -o -name scsi* "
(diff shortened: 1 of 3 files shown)


Martin-Weiss:
It seems that we have to deal with the geneal "no 100% solution is available" problem.
On one hand we have to make the device management persistent and on the other hand devices are used that do not allow to do that / use that persistence.

If the OS does not provide anything reliable - the user has to control that and we need to find a way to give the user the control - maybe allow to set labels / aliases etc.?

swiftgist:
I will close this one for now since the current issue may only affect RAID devices.

PR #1206, opened 2018-06-27
ensure proper hardware detection tool in makefile

### LINKED ISSUE #1205: salt minion crash: missing hw detect tool

The salt minion faulted/crashed during DeepSea installed.  Logs reported root cause-
```
Jun 26 20:48:49 masterminion salt-minion[8642]: [WARNING ] The minion function caused an exception
Jun 26 21:00:23 masterminion salt-minion[8642]: [WARNING ] The function "module.run" is using its deprecated 
Jun 26 21:00:28 masterminion salt-minion[8642]: [ERROR   ] Can not find a proper hardware detection tool.
Jun 26 21:00:28 masterminion salt-minion[8642]:             Install lshw or hwinfo in order to retrive hardwa
```

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

Testing on k8s container.

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

Test DeepSea inside Container missing hw probing tools

### Versions Report
(Provided by running:
`salt-run deepsea.version`
`rpm -qi salt-minion`
`rpm -qi salt-master`
)

noelmcloughlin:
Fixes #1205

Description:

Include hardware probing tools for DeepSea inside Makefile. If missing throw exception but allow `make install` to continue.

-----------------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ VERSION ?= $(shell (git describe 2>/dev/null || echo '0.0.0') | sed -e 's/^v//' 
 PYTHON_DEPS=python3-setuptools python3-click python3-tox
 PYTHON=python3
 PIPCMD=""
+HW_INFO=hwinfo
 
 OS=$(shell source /etc/os-release 2>/dev/null ; echo $$ID)
 ifeq ($(OS), opensuse)
@@ -24,6 +25,7 @@ PKG_INSTALL=zypper -n install
 else
 USER=root
 GROUP=root
+HW_INFO=lshw
 ifeq ($(OS), centos)
 PKG_INSTALL=yum install -y
 PYTHON_DEPS=python-setuptools python-click python-tox
@@ -841,7 +843,7 @@ copy-files:
 
 install-deps:
 	# Using '|| true' to suppress failure (packages already installed, etc)
-	$(PKG_INSTALL) $(PYTHON_DEPS) || true
+	$(PKG_INSTALL) $(PYTHON_DEPS) $(HW_INFO) || true
 	$(PIPCMD) >/dev/null 2>&1 || true
 
 install: pyc install-deps copy-files


swiftgist:
This would only install the package on the admin node.  Is /srv/salt/ceph/packages/common/default.sls missing what's needed for your platform?  I do not recall needing the package earlier than .../ceph/stage/0/minion/default.sls.  

The packages/common directory is a candidate for switching to the `os_switch` macro to make platform support a bit more obvious if we do need to make the change there.

noelmcloughlin:
@swiftgist Good point, okay there is no point updating Makefile - my bad.  I can close this PR leave open and update the commit with alternative solution for #1206 based on your comment - wdyt?

I looked at `salt/ceph/macros/os_switch.sls` and am willing to migrate packages/common.  I want to `os_family` grain into that macro first - its best grain for OS/Package filtering imo.

How do you prefer issue & PR to be handled in this situation. Happy to follow your conventions.

swiftgist:
@noelmcloughlin Take a look at #998 for how all that started.  I swear I found examples where "osfamily" has two or more words.  (I will go look in a bit.)  Thought spacing in directory/files would make that difficult.  I believe the current macro will allow for both "CentOS" and "CentOS-7.1" name files that would behave as expected.

Do you have something not covered?

noelmcloughlin:
@swiftgist look for `os_family` grain which should be added to os_switch anyway - maybe to help #998 work. For example I wrote this [osfamilymap.yaml ](https://github.com/saltstack-formulas/mysql-formula/blob/master/mysql/osfamilymap.yaml) two weeks ago - it matches against `os_family` grain - so RedHat matches Fedora/Centos/RHEL and SUSE matches SLES/OpenSuse, etc.  I'm not sure when it was introduced to salt .. in 2017 at least I think.

swiftgist:
@noelmcloughlin I'll take a look.  Now, I have to find what I am remembering as well. :)

swiftgist:
Okay, re-reading my own issue, it's `osfinger` that may give spaces.  That's why I added `osrelease`.    

@noelmcloughlin  Now, back to `osfamily`.  Out of curiosity, which families are you supporting?  The SUSE family according to Salt is "Suse".  I doubt we would use it to represent SLE or openSUSE simply because there's a whole history of  various spellings such as S.u.S.E., SuSE and SUSE.  (And I'd rather avoid some of those quasi religious wars on that one.)

That said, I'm curious whether you are wanting 'RedHat' for Fedora/CentOS/Redhat or 'Debian' for Debian/Ubuntu/others.  The  [osfamilymap.yaml ](https://github.com/saltstack-formulas/mysql-formula/blob/master/mysql/osfamilymap.yaml) has 'SUSE' which is the `os` grain and not the `os_family`.

I have seen some Salt complaints about whether distros should or should not be part of a family and I do wonder about collisions between `os_family` and `os` for some distros like Debian.  I looked through some of the Salt code and the same spelling is used for both.  How would I add a state that is only intended to run on the Debian distribution but not the Debian family?  If I start adding Jinja conditionals inside the state file to distinguish the other distros then I'm back where I started before using os_switch.  

I am not against adding it, but I do not want to cause a headache for a Debian contributor that is not interested in Ubuntu but would be breaking another distro unintentionally by updating a Debian specific state file.  I also think that seeing the os name such as Fedora is a stronger signal that your distro is actually supported rather than the RedHat file that "should" work.

***
The original problem is that the hardware detection tools are not getting installed.  We need to get that addressed.  I'll see how full my plate is this coming week.

noelmcloughlin:
@swiftgist, okay its blurred. Most developers I work with like Ubuntu, Production Engineering guys used Centos, DBA guys have few Oracle Linux, and my desktop is MacOS/Darwin.  I dab with Fedora sometiimes - I first used SuSE around 1994 and liked the lizard - I have an ancient SuSE CD set somewhere ;-)

Regarding DeepSea, I am targeting SuSE and Ubuntu in the first instance.  I have written a [deepsea-formula](https://github.com/saltstack-formulas/deepsea-formula/blob/master/deepsea/osfamilymap.yaml) to automate some READE steps so I can get basic SW installed on Ubuntu/Suse/Redhat. Then I'm testing DeepSea, and finally I'm trying to integrate DeepSea into [openSDS project](https://www.opensds.io/) to offer some choice over ceph-ansible.  

I'm not into religious debate on `os_family` categorization is right/wrong or ansible v.s. salt. My motivation is to build solutions and do hobby work on things I'm interested, and experienced with, therefore my interest in Software Defined Storage from a broader perspective.  DeepSea is mix of Salt and python and I find the concept interesting.  Thats where I'm coming from.

My approach to `os_switch` would be to include `os_family` but not force using it. I want to contribute to your project/community collaboratively and forcing one's opinion is unhelpful, trying to influence opinions and challenge ways of doing things is fine - but one thing at a time.

I have a `pattern` which I follow regarding salt grains. I'll stick common defaults into defaults.yaml, osfamily grains into osfamilymap.yaml to override defaults, and (only if necessary) use the `os` grain inside a `osmap.yaml` and use a `map.jinja` to merge dictionaries logically as per example below. 

It's not the right or wrong way, just a pattern.  I'm open to new patterns, and DeepSea includes technologies (runners, mines) which I have zero experience with....  

```
DEFAULTS.YAML
mysql:
  serverpkg: mysql-server
  clientpkg: mysql-client


OSFAMILYMAP.YAML
Debian:
  serverpkg: mariadb-server
  service: mariadb


OSMAP.YAML
Ubuntu:
  service: mysql


MAP.JINJA
{% import_yaml "mysql/defaults.yaml" as defaults %}
{% import_yaml "mysql/osfamilymap.yaml" as osfamilymap %}
{% import_yaml "mysql/osmap.yaml" as osmap %}

{% set mysql = salt['grains.filter_by'](
    defaults,
    merge=salt['grains.filter_by'](
        osfamilymap,
        grain='os_family',
        merge=salt['grains.filter_by'](
            osmap,
            grain='os',
            merge=salt['pillar.get']('mysql', {}),
            ),
        ),
    base='mysql',
) %}
````

I hope this explanation is helpful.  

This PR can probably be closed because updating the Makefile is wrong solution as you pointed out.  #1205 can track the solution for hardware.   

I can open second issue suggesting os_switch includes `os_family` but if thats not preferred solution, I 'll work with what is there.  The `os` grain is usable, maybe too specific, but one step at a time.

swiftgist:
I have to test https://github.com/SUSE/DeepSea/tree/wip-package-common, but I think this should cover the cases while explicitly listing the hwinfo/lshw packages.

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
retest this please

noelmcloughlin:
Hi @swiftgist.

Just to followup.
1. This PR should be cancelled as only fixes #1205 for master, not minions.
2. Any further discussion should move to #1205.  I'm interested to revisit the common package handling after my summer vacation.

thanks,

PR #1210, opened 2018-07-02
Only restart with a running cluster

### LINKED ISSUE #954: stage 0, restart orchestration
I may have found a catch-22.  Stage 0 calls the restart orchestration whether the previous minion orchestration ran as a fresh install or on an existing cluster.  The issue with a fresh install is that the restart eventually calls wait.out.  This fails on a conf_read_file.

We seem to have restart logic in the stage.0.minion as well as at the top level.  The restart should only be needed on an existing cluster.

swiftgist:

Fixes #954


smithfarm:
backport is https://github.com/SUSE/DeepSea/pull/1380

Issue #1218, opened 2018-07-11
Add new jq dependency to spec file


smithfarm:
https://github.com/SUSE/DeepSea/pull/1117 introduced a new runtime dependency, `jq`, that needs to be installed on the Salt Master to avoid a regression in Stage 0:

```
[ERROR   ] Rendering exception occurred: Jinja error: Unable to
load json from /bin/sh: jq: command not found
Traceback (most recent call last):
   File
"/usr/lib/python2.7/site-packages/salt/utils/templates.py",
line 368, in render_jinja_tmpl
     output = template.render(**decoded_context)
   File
"/usr/lib/python2.7/site-packages/jinja2/environment.py", line
989, in render
     return self.environment.handle_exception(exc_info, True)
   File
"/usr/lib/python2.7/site-packages/jinja2/environment.py", line
754, in handle_exception
     reraise(exc_type, exc_value, tb)
   File "<template>", line 26, in top-level template code
   File "/usr/lib/python2.7/site-packages/salt/utils/jinja.py",
line 403, in load_json
     'Unable to load json from {0}'.format(value))
TemplateRuntimeError: Unable to load json from /bin/sh: jq:
command not found

; line 26

---
[...]
wait until all active mds but one have stopped:
   salt.state:
     - tgt: {{ master }}
     - sls: ceph.wait.mds.1-mds

{% set standbys = salt['saltutil.runner']('cmd.run', cmd='ceph
--format=json fs dump 2>/dev/null | jq -c [.standbys[].name]')
| load_json %}    <======================
{% for standby in standbys %}
shutdown standby daemon {{ standby }}:
   salt.state:
     - tgt: {{ standby }}
     - sls: ceph.mds.shutdown
[...]
---
[CRITICAL] Rendering SLS 'base:ceph.restart.mds.default-shrink'
failed: Jinja error: Unable to load json from /bin/sh: jq:
command not found
Traceback (most recent call last):
   File
"/usr/lib/python2.7/site-packages/salt/utils/templates.py",
line 368, in render_jinja_tmpl
     output = template.render(**decoded_context)
   File
"/usr/lib/python2.7/site-packages/jinja2/environment.py", line
989, in render
     return self.environment.handle_exception(exc_info, True)
   File
"/usr/lib/python2.7/site-packages/jinja2/environment.py", line
754, in handle_exception
     reraise(exc_type, exc_value, tb)
   File "<template>", line 26, in top-level template code
[...]

PR #1219, opened 2018-07-11
build/ops: add jq runtime dependency to spec file

smithfarm:
Fixes: https://github.com/SUSE/DeepSea/issues/1218

### DIFF
--- a/deepsea.spec.in
+++ b/deepsea.spec.in
@@ -36,6 +36,7 @@ Requires:       salt-master
 Requires:       salt-minion
 Requires:       salt-api
 Requires:       iputils
+Requires:       jq
 Requires:       python3-netaddr
 Requires:       python3-rados
 Requires:       python3-configobj


jan--f:
Huh I guess the jq install in packages.common is not enough?

smithfarm:
RPM is a pretty good mechanism for installing dependencies of code (Salt Runners in this case) that runs on the machine where an RPM (deepsea in this case) is installed.

I understand that Stage 0 is supposed to install DeepSea's runtime dependencies on the minions that need them, and that's the only option for installing dependencies on the non-master minions, but for the master minion dependencies RPM has a big advantage in that RPM runs before any DeepSea code gets run.

In other words, given two mechanisms for installing dependencies - one that uses an external tool which runs before the dependent code runs, and another which relies on the dependent code itself to install the dependency, IMO the former will always be preferable.

smithfarm:
backport is https://github.com/SUSE/DeepSea/pull/1225

PR #1220, opened 2018-07-16
Always return an entry from Device Files, if possible

swiftgist:
bsc: 1097532

### DIFF
--- a/srv/salt/_modules/proposal.py
+++ b/srv/salt/_modules/proposal.py
@@ -239,9 +239,7 @@ def _device(drive):
     if 'Device Files' in drive:
         devices = drive['Device Files'].split(', ')
         index = _prefer_underscores(devices)
-        # Prefer 'Device File' over last item
-        if index > -1:
-            return devices[index]
+        return devices[index]
     return drive['Device File']
 
 


swiftgist:
Just a note: `make lint` is succeeding for me locally.

kshtsk:
@swiftgist Try to run 'make lint' in clean environment.

swiftgist:
@kshtsk what error do you get?  I do not see one when I click Details.

kshtsk:
@swiftgist Click at " Console Output"

jschmid1:
Is the description of `_prefer_underscore` actually correct? It states:

`In the worst case, return the last device`. And the function actually does that by returning a `-1` if there is no item with an underscore in the list.
However the invoking method `_device` has:

``` python
if index > -1:
    return devices[index]
return drive['Device File']
```

This logic implies that if there is no drive with an underscore, ( returns -1  and supposedly return the _last_ item ) the conditional does not assert and we drop down to returning the drive['Device File'] which is most of the time the short name. ( or the _first_ item in the drives list )

swiftgist:
This fix was originally to address an issue in the field.  The first device in 'Device Files' is the short name so the last device would often begin with 'scsi-3'.  

If #1222 is accepted, then this PR is not needed.

jschmid1:
Closing in favor of #1222

PR #1221, opened 2018-07-16
Convert unicode to strings due to issues with yaml dumps

jschmid1:

Convert unicode to strings due to issues with yaml dumps 

SUSEINTERNAL: bsc#1100136

-----------------

### DIFF
--- a/srv/modules/runners/proposal.py
+++ b/srv/modules/runners/proposal.py
@@ -12,6 +12,7 @@ import os
 from sys import exit
 import logging
 # pylint: disable=import-error,3rd-party-module-not-gated,redefined-builtin
+from salt.ext import six
 import salt.client
 import yaml
 # pylint: disable=import-error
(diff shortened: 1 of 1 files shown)


jschmid1:
something has changed between #1219 and now in the linter/dependencies. @kshtsk Do you know anything about that?

jan--f:
Don't have a way of checking at the moment, but wouldn't it be enough to add ```allow_unicode=True``` to the ```yaml.dump()``` call?

kshtsk:
@jschmid1 I haven't change anything, I only know that it became an issue for master branch as well.

jan--f:
lets use ```yaml.safe_dump```

jschmid1:
That is not needed for master anyways, Closing and reimplement that atop SES5 with the 'safe_dump' method, which seems to be less verbose. 'allow_unicode' does not do the trick.

PR #1229, opened 2018-07-23
qa: add functional.sh integration test script

smithfarm:
The new "functional.sh" integration test script runs Stages 0-3
followed by the "ceph.smoketests" orchestration.


smithfarm:
The objective is to enable running the functional test orchestrations via teuthology, and I realized teuthology has better ways of reaching this objective.

Issue #1234, opened 2018-07-24
ceph.smoketests fails on SLE15 because it can't find an executable "aa-enforce"


smithfarm:
Environment: SLE15/SES6

How to reproduce: deploy a single-node cluster with MDS and RGW enabled by running Stages 0-4 with default settings. Then do `salt-run state.orch ceph.smoketests`.

The AppArmor smoke tests fail because the orchestration tries to run all the scripts in `/etc/apparmor.d` using an executable called `aa-enforce`, but there is no such executable in the system.

Sample of one of the errors generated:

```
                        ID: aa-enforce /etc/apparmor.d/usr.bin.ceph-mon
                  Function: cmd.run
                    Result: False
                   Comment: Command "aa-enforce /etc/apparmor.d/usr.bin.ceph-mon" run
                   Started: 16:12:23.688331
                  Duration: 13.443 ms
                   Changes:   
                            ----------
                            pid:
                                55667
                            retcode:
                                127
                            stderr:
                                /bin/sh: aa-enforce: command not found
                            stdout:
```

smithfarm:
Either the `ceph.smoketests` orchestration could install the dependency, or it could be added to the spec file as a dependency of the `deepsea-qa` package?

smithfarm:
Feel free to just suggest the "right" fix and I'll be happy to implement it, or at least try to.

jschmid1:
So.. in theory:

```
ceph.smoketests 
  -> ceph.smoketests.apparmor
      -> ceph.apparmor.default-enforce
          -> ceph.apparmor.install
              -> zypper -n in -t pattern apparmor ( where aa-enforce is included )
          -> ceph.apparmor.profiles
```

But I get an error also on my machine. Need to look deeper into that

note: @jan--f is out for ~2weeks

smithfarm:
So, `aa-enforce` is part of the `apparmor-utils` package. I suggest to make this a dependency of `deepsea-qa`. That will solve the problem once and for all.

jschmid1:
> UPDATE: On the other hand, that might mask a real bug, so probably better to avoid that.

That's what I thought.. it should actually be installed by the time the smoketest runs..

smithfarm:
@jschmid1 The DeepSea README says:

```
Note that the Cluster deployment with DeepSea will not work with Firewall/AppArmor enabled.

Check whether the AppArmor service is running and disable it on each cluster node to avoid problems:

# systemctl disable apparmor.service
```

[...]

PR #1235, opened 2018-07-25
apparmor: install apparmor-utils package on SUSE

smithfarm:
On SLE15 at least, installing "patterns-base-apparmor" does not cause
apparmor-utils to be installed, and apparmor/default-enforce.sls fails
due to missing executable "aa-enforce".

Fixes: https://github.com/SUSE/DeepSea/issues/1234

### DIFF
--- a/srv/salt/ceph/apparmor/install/SUSE.sls
+++ b/srv/salt/ceph/apparmor/install/SUSE.sls
@@ -1,6 +1,9 @@
 
-patterns-base-apparmor:
-  pkg.installed
+install apparmor:
+  pkg.installed:
+    - pkgs:
+      - patterns-base-apparmor
+      - apparmor-utils
 
 apparmor:
   service.running:


smithfarm:
I put "needs backport" label because it makes sense to be double-sure that the package is installed, even if the apparmor package in SLE-12-SP3 installs it (it might stop installing it due to some future update).

jschmid1:
Instead of invoking `pkg.install` twice, we may use the *arg passing for pkg.installed like [here](https://github.com/smithfarm/DeepSea/blob/e02d1c648934278b4a817a699d4f35b2cee3e44c/srv/salt/ceph/apparmor/install/default.sls)

adapted version:

```
install apparmor:
  pkg.installed:
    - pkgs:
      - patterns-base-apparmor
      - apparmor-utils
```

that'd probably safe 1 invocation of zypper.

jschmid1:
> I put "needs backport" label because it makes sense to be double-sure that the package is installed, even if the apparmor pattern in SLE-12-SP3 installs it (it might stop installing it due to some future update).

agreed

smithfarm:
@jschmid1 OK, done.

PR #1236, opened 2018-07-25
passed arg was not changed in sls file after changed in function header

jschmid1:

Description:

Stage.5 reports:

> drain osd.6: The following arguments are missing: osd_id

which comes from ` ceph.remove.storage.drain (/srv/salt/ceph/remove/storage/drain)`

~dd688e193~  was changing the func header but we missed to change that also in the sls files

EDIT: it was actually fa78d09df00d3197ecc6d084c186514e82569311

-----------------


jschmid1:
created trackerissue to improve functional tests -> https://github.com/SUSE/DeepSea/issues/1237

PR #1249, opened 2018-07-27
build/ops: streamline qa %files section

smithfarm:
Merely mentioning the top-level directory is sufficient to bring in that
directory and everything under it.

This is a Big Stick, and not to be used lightly, but since (a) the deepsea-qa
package is not intended for wide consumption and (b) the directories in
question contain only QA-related stuff, it seems OK to use here.


smithfarm:
Doing it this way is probably against various packaging guidelines, but owning files in `/srv` is, too...

PR #1248, opened 2018-07-27
[DNM] qa: enable (optionally randomized) testing of OSD configurations

smithfarm:
Apparently this has already found one bug: #1247 

TODO:

- [x] implement `--profile=<name_of_config>`
- [x] implement randomization

Issue #1247, opened 2018-07-27
osd-config "bs_dedicated_db.yaml" breaks Stage 3


smithfarm:
On DeepSea master branch, if I use the OSD configuration from `qa/osd-config/ovh/bs_dedicated_db.yaml` Stage 3 fails like this:

(steps 1-27 succeed)

```
2018-07-27T14:33:23.495 INFO:teuthology.orchestra.run.target217182143060.stdout:[28/29] Executing state ceph.processes.osd...
2018-07-27T14:33:23.495 INFO:teuthology.orchestra.run.target217182143060.stdout:         |_  target217182143060.teuthology: wait for osd processes(cephprocesses.wait) fail
2018-07-27T14:33:23.496 INFO:teuthology.orchestra.run.target217182143060.stdout:         target217182143060.teuthology... fail
2018-07-27T14:33:23.496 INFO:teuthology.orchestra.run.target217182143060.stdout:Finished stage ceph.stage.3: succeeded=27/29 failed=1/29
2018-07-27T14:33:23.496 INFO:teuthology.orchestra.run.target217182143060.stdout:
2018-07-27T14:33:23.496 INFO:teuthology.orchestra.run.target217182143060.stdout:Failures summary:
2018-07-27T14:33:23.496 INFO:teuthology.orchestra.run.target217182143060.stdout:ceph.processes.osd (/srv/salt/ceph/processes/osd):
2018-07-27T14:33:23.496 INFO:teuthology.orchestra.run.target217182143060.stdout:  target217182143060.teuthology:
2018-07-27T14:33:23.497 INFO:teuthology.orchestra.run.target217182143060.stdout:    wait for osd processes: Module function cephprocesses.wait executed
2018-07-27T14:33:23.588 INFO:teuthology.orchestra.run.target217182143060.stdout:deepsea exit status: 1
```

Note: this is a single-node cluster in OVH with root (/dev/vda) + 4 (/dev/vd[bcde]) disk drives.

Relevant excerpt from pillar data:

```
2018-07-27T14:27:08.593 INFO:teuthology.orchestra.run.target217182143060.stdout:    ceph:
2018-07-27T14:27:08.593 INFO:teuthology.orchestra.run.target217182143060.stdout:        ----------
2018-07-27T14:27:08.593 INFO:teuthology.orchestra.run.target217182143060.stdout:        storage:
2018-07-27T14:27:08.593 INFO:teuthology.orchestra.run.target217182143060.stdout:            ----------
[...]

PR #1250, opened 2018-07-28
qa: health-ok.sh: enable testing of OSD configurations

smithfarm:
This PR refactors the health-ok.sh script to allow OSD configuration to be chosen at runtime from among those in `qa/osd-configs/ovh` (this directory is hard-coded for now) -- either explicitly (`--policy=<name>`) or randomly (`--policy=random`).

Continue to support deployment of all filestore and all encrypted OSDs via `--policy=filestore` and `--policy=dmcrypt`.


smithfarm:
Replaces #1248

smithfarm:
@jschmid1 The git history is pretty messed up, since this is a refactor. I propose to squash all the commits into one. I will of course backport the feature to SES5, but it won't be a cherry-pick.

jschmid1:
> @jschmid1 The git history is pretty messed up, since this is a refactor. I propose to squash all the commits into one. I will of course backport the feature to SES5, but it won't be a cherry-pick.

That's fine.

> See it working here: http://137.74.25.20:8081/ubuntu-2018-07-28_09:24:13-suse:smoke:health-ok-wip-qa-osd-config---basic-openstack/
> 
> especially the part between Stages 1 and 2! The failures are due to #1247

I think we have what we need to bet a better understanding of why things failed. +1

The failures mentioned in #1247 need to be analyzed of course, but for now we have the tests in place and can proceed. 

Imo we also need a way to iteratively run _all_ configurations.

smithfarm:
@jschmid1 This PR *enables* all kinds of testing of OSD configurations, including testing of *all* the configs (in parallel, or sequentially - however you like).

In teuthology, we'll test them all in parallel via this commit: https://github.com/SUSE/ceph/commit/fd9f88656263c7ad6c388e91a5d75937e8620732

jschmid1:
> @jschmid1 This PR enables all kinds of testing of OSD configurations, including testing of all the configs (in parallel, or sequentially - however you like).

If possible, always paralllel :) I just thought of the hardware/resource constraints that would prevent us from doing so. If that's not a problem, let's go with parallel.



> In teuthology, we'll test them all in parallel via this commit: SUSE/ceph@fd9f886

Ah, great :+1:

smithfarm:
@jschmid1 Good, because deploying clusters sequentially is not one of teuthology's strong points :-)

Issue #1215, opened 2018-07-06
proposal.generate sorting


swiftgist:
### Description of Issue/Question
The current yaml files generated by Stage 1 are pseudo random.  As a result, regenerating a configuration may result in devices appearing in a different order.  For standalone Filestore and Bluestore configurations, this does not matter.  For split Filestore and Bluestore configurations, the new configuration is valid, but a migration may be necessary.

One suggestion from a couple of discussions is to sort the devices based on the `/dev/disk/by-path` entries.  That would prevent accidental shuffling for disk replacements.

denisok:
@jan--f @jschmid1 what are your thoughts on this?

jschmid1:
> One suggestion from a couple of discussions is to sort the devices based on the /dev/disk/by-path entries. That would prevent accidental shuffling for disk replacements.

This is what the discussions resulted in. The downside of this approach is that a user will have to do initial migrations(probably) after changing to the new sorting scheme.

jan--f:
I believe the profile generation should be deterministic, i.e. all things equal the code should spit out identical profiles.

What is the situation where profiles would be regenerated? Would that be with the same hardware? Are we talking about changing profiles for OSD replacement?

For the latter it might be better to change the existing profiles where necessary. The yaml files can easily be loaded into python objects and then altered and dumped again. In my head that should be straight forward, though I'm not sure regarding the use case.

jschmid1:
```
proposal.populate ( runner )
  -> proposal.generate ( module )
    -> cephdisks.list
      -> hwinfo/lshw
```

That would mean; **If** the profile generation is deterministic we currently rely on the hwinfo/lshw output to be sorted.
By looking at a simple hwinfo output ( `hwinfo --only disk` [example](http://paste.opensuse.org/86078068) ) I assume that the internal sorting is done by looking at the _Storage Controller #_ . 

Looking at a more sophisticated cluster, things are different and hwinfo does not seem to sort by _Storage Controller #_ 

Next step is to have a look at the hwinfo implementation and check how / if it's sorted.

jschmid1:
Things change if attached to a rais bus controller..

jschmid1:
I added a test that tries to figure out how the module behaves with unsorted input.

jschmid1:
If this however not assert as true we have to go the other route, like @jan--f already stated, to read-in the moved 'profile-replaced' file and make sure that the newly generated proposal matches the old layout and wal-db matching.

[...]

PR #1260, opened 2018-08-02
[RFC] Test determinism for proposal.generate

jschmid1:

Addresses #1215 ( Adding verify tests for this )

Description:

I'm not sure if that actually proves it. Please review thoroughly!

```
        Theory:
        The fear is that in a 'replace-disk' operation we _may_ end up with
        different osd - wal/db mappings than before. This would mean that
        a user will have to edit the profiles manually or face a migration.
        We never introduced pre-sorting of disk information that we get from
        hwinfo, which means that we can't guarantee a consistently sorted
        input.

        In order to test this, I try to do the following:

        1) Generate a profile
        2) 'replace' a disk ( replaced.disk() )
        2.1) Generate a 'new' disk ( suffixed with '_GENERATED' )
        2.2) Inject it a at a random index in the output from hwinfo
        2.3) Find the old, replaced disk to compare the wal/db pointer
        3) Compare the new and the old proposal for a match of
           old wal/db pointer and the new wal/db pointer

        To eliminate the factor of 'luck', repeat if 1000 times.

        Caveat:
          * Currently we are only testing if the newly replaced disk
            is matching it's old entry.
            Maybe we should test it for _every_ disk.
EDIT:    ->> I do. <<-
          * The number of repetitions could me reduced by adding more disks
            and therefore an increase in probability of a 'missplacement'
          * We have 3k tests more :/
```

-----------------

### DIFF
--- a/tests/unit/_modules/test_proposal.py
+++ b/tests/unit/_modules/test_proposal.py
@@ -2,10 +2,11 @@
 
 import pytest
 import sys
+import random
 sys.path.insert(0, 'srv/salt/_modules')
 from srv.salt._modules import proposal
 from srv.salt._modules import helper
-from tests.unit.helper.output import OutputHelper
+from tests.unit.helper.output import OutputHelper, GenHwinfo
 
 
 class TestProposal(object):
@@ -13,6 +14,18 @@ class TestProposal(object):
     @pytest.fixture(scope='module')
     def output_helper(self):
         yield OutputHelper()
+
+    @pytest.fixture()
+    def replaced_disk_hdd(self):
+        return self.replace_disk()
+
+    @pytest.fixture()
+    def replaced_disk_ssd(self):
+        return self.replace_disk(type='ssd')
+
+    @pytest.fixture()
+    def replaced_disk_nvme(self):
+        return self.replace_disk(type='nvme')
 
     def test_parse_filter(self, output_helper):
         # test defaults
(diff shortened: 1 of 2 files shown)


jschmid1:
retest this please

tests/unit/_modules/test_proposal.py
```diff
@@ -129,3 +146,142 @@ def test_create(self, output_helper):
...
+        # gen new proposal
+        new_p = proposal.Proposal(hwinfo_out_new)
+        new_prop = new_p.create()
+
+        for prop_name, prop in  old_prop.items():
+            for disk_set in old_prop[prop_name]:
+                # ensure that its a proposal with dedicated wal/db
+                if disk_set.keys()[0] in replaced_hdd['Device Files']:
+                    if prop_name in ['ssd-spinner', 'nvme-ssd-spinner', 'nvme-ssd', 'nvme-spinner']:
+                        wal_db_replaced = disk_set.values()[0]
```
alexandergraul: As we discussed yesterday: this overwrites `wal_db_replaced`, turning it into a dict with `prop_name` as the key should be better.

tests/unit/_modules/test_proposal.py
```diff
@@ -129,3 +146,142 @@ def test_create(self, output_helper):
...
+
+                if not matching_disk_from_old_prop and osd_disk not in replaced_hdd['Device Files']:
+                    if 'GENERATED' in osd_disk:
+                        if wal_db != wal_db_replaced:
+                            return False, matching_disk_from_old_prop, osd_disk, wal_db, wal_db_replaced
+
+                if matching_disk_from_old_prop:
+                    assert osd_disk == matching_disk_from_old_prop[0].keys()[0]
+                    assert wal_db == matching_disk_from_old_prop[0].values()[0]
+                return True, [], osd_disk, wal_db, wal_db_replaced
```
alexandergraul: This return is too early, only one `disk_set in old_prop[prop_name]` gets checked.

tests/unit/_modules/test_proposal.py
```diff
@@ -129,3 +146,142 @@ def test_create(self, output_helper):
...
+        # idx 0-11 -> HDD
+        # idx 12-17 -> SSD
+        # idx 18 -> NVME
+        wal_db_replaced = None
+        hwinfo_out = OutputHelper().cephdisks_output
+        p = proposal.Proposal(OutputHelper().cephdisks_output, ratio=ratio)
+        old_prop = p.create()
+
+        # replace a HDD
+        hwinfo_out_new = hwinfo_out
```
alexandergraul: this is a bit misleading, `hwinfo_out_new` is not a copy `hwinfo_out`, just a second reference to the same list. It does not matter here because `hwinfo_out` is only used for the max index when inserting the new disk at a random spot, but can be confusing.

Since only one `hwinfo_out` is needed, we don't need to create a second reference. If we want to make the code easier to understand by having a second list, we can use list slicing to copy all values: `hwinfo_out_new = hwinfo_out[:]`

jschmid1:
Closing this PR as we moved to another way of replacing disks. Introduced #1302

PR #1261, opened 2018-08-02
Add an option to disable write cache of the rotational disks

### LINKED ISSUE #1238: DeepSea should disable write cache for all disks that do not have a battery backed write cache
When ever we have spinning disks that do not have battery backed write cache - we have to disable the write cache to ensure data-consistency also in case of power outages.
For SSD/NVMe this is most often not required to be turned of because they have capacitors that allow to flush the write cache also in case of a power outage without loosing data.

--> could we get the command "sdparm --clear=WCE --save /dev/$DISK" added to disable WCE on all spinning disks that do not have battery backed write cache?

(check the status with "sdparm /dev/$DISK|grep WCE" --> should be "0")

kbaikov:
Fixes #1238 

When ever we have spinning disks that do not have battery backed write cache - we have to disable the write cache to ensure data-consistency also in case of power outages.

Description:

This sdparm command does not always work with the --save parameter, hense the persistent=False default argument. Apparently the special manufacturer specific utility is needed for disabling WCE.
But with some manufacturers sdparm is enough.

-----------------

### DIFF
--- a/srv/salt/_modules/cephdisks.py
+++ b/srv/salt/_modules/cephdisks.py
@@ -512,29 +512,33 @@ def list_(**kwargs):
     return hwd.assemble_device_list()
 
 
-def disable_write_cache(disk=None, persistent=False, **kwargs):
+def disable_write_cache(disk="All", persistent=False, **kwargs):
     """
     Disable write cache (WCE) on rotational disks
     """
     hwd = HardwareDetections(**kwargs)
     sdparm_path = hwd._which('sdparm')
-    base_disk_path = "/sys/block/{}".format(disk)
-    disk_path = "/dev/{}".format(disk)
 
-    if persistent:
-        cmd = "{} --clear=WCE --save {}".format(sdparm_path, disk_path)
+
+    if disk == "All":
+        hwd_list = hwd.assemble_device_list()
+        disks_to_modify = [d['device'] for d in hwd_list if d['rotational'] == '1']:
     else:
-        cmd = "{} --clear=WCE {}".format(sdparm_path, disk_path)
+        disks_to_modify = [disk]
 
-    if hwd._is_rotational(base_disk_path):
+    for d in disks_to_modify:
+        base_disk_path = "/sys/block/{}".format(d)
+        disk_path = "/dev/{}".format(d)
+
+        if persistent:
+            cmd = "{} --clear=WCE --save {}".format(sdparm_path, disk_path)
+        else:
+            cmd = "{} --clear=WCE {}".format(sdparm_path, disk_path)
+
         proc = Popen(cmd, stdout=PIPE, stderr=PIPE, shell=True)
         stdout, stderr = proc.communicate()
         if stderr:
-            log.warning("Could not change write cache settings for {}".format(disk))
-            return False
-    else:
-        log.warning("Disk {} is not rotational".format(disk))
-        return False
+            log.warning("Could not change write cache settings for {}".format(d))
     return True
 
 def version():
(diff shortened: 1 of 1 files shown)


srv/salt/_modules/cephdisks.py
```diff
@@ -512,6 +512,35 @@ def list_(**kwargs):
...
+        if persistent:
+            cmd = "{} --clear=WCE --save {}".format(sdparm_path, disk_path)
+        else:
+            cmd = "{} --clear=WCE {}".format(sdparm_path, disk_path)
+
+        proc = Popen(cmd, stdout=PIPE, stderr=PIPE, shell=True)
+        stdout, stderr = proc.communicate()
+        if stderr:
+            log.warning("Could not change write cache settings for {}".format(d))
+    return True
```
jschmid1: You return True even if changing the write cache settings did not work.

srv/salt/_modules/cephdisks.py
```diff
@@ -512,6 +512,35 @@ def list_(**kwargs):
...
 
+def disable_write_cache(disk="All", persistent=False, **kwargs):
+    """
+    Disable write cache (WCE) on rotational disks
+    """
+    hwd = HardwareDetections(**kwargs)
+    sdparm_path = hwd._which('sdparm')
+
+
+    if disk == "All":
```
jschmid1: What about inverting the logic. If the user provides no `disk` parameter, it's assumed to auto-select all disks. If there is a disk parameter set the user only wants to change the settings for one.

Also hardcoded it to (capital)'All' which will be hard to guess.

A couple of more considerations:

* If someone disabled it, he maybe also wants to enable it again.
* If we have a disable and a enable function, we may also need a help function

Taking these two facts into account, it might even be wise to move that to it's own module where you can make use of a `xxxxx.help` notation.

srv/salt/_modules/cephdisks.py
jschmid1: On success we may also want a debug log statement.

srv/salt/_modules/cephdisks.py
alexandergraul: > What about inverting the logic. If the user provides no disk parameter, it's assumed to auto-select all disks. If there is a disk parameter set the user only wants to change the settings for one.

Isn't that already the case with the default `disk="All"`?

I do agree with your criticism of "All", something like `if disks.lower() == "all":` would guard against wrongly capitalized input.

> * If someone disabled it, he maybe also wants to enable it again.

Sounds reasonable.

> * If we have a disable and a enable function, we may also need a help function

I also agree with this. As of now, there is only `cephdisks.list`, could we use `cephdisks.help` to cover `list`, `disable_write_cache` and `enable_write_cache` (plus maybe a mention of `version`)? Moving the dis-/enablement of the write cache to another module might not be needed.

kshtsk:
retest this please

kshtsk:
retest this please

srv/salt/_modules/cephdisks.py
```diff
@@ -512,6 +512,35 @@ def list_(**kwargs):
...
+        hwd_list = hwd.assemble_device_list()
+        disks_to_modify = [d['device'] for d in hwd_list if d['rotational'] == '1']:
```
jschmid1: Syntax Error. the ":" after the expression

jschmid1:
any update @kbaikov

kbaikov:
@jschmid1 not yet. Perhaps Monday.

jschmid1:
stale - please reopen once this comes active again.

PR #1268, opened 2018-08-14
Install smartmontools while stage.0

### LINKED ISSUE #1068: deepsea discovery fails in case smartmontools are not installed

When executing the stage.1 it fails due to raid controller in the server but smartmontools not installed and due to that smartctl not available

### Setup
Lenovo servers with hardware raid controller

### Steps to Reproduce Issue
Execute stage.1 on one of these servers

Expectation is that deepsea takes care of installing all required rpms - so could we get smartmontools added?

jschmid1:

Fixes #1068

Description: See issue

-----------------

**Checklist:**
- [ ] Adapted documentation


jschmid1:
[Bug 1104779 - deepsea stage 1 fails in proposal generate in case lshw or hwinfo is not installed but deepsea should install that in stage 0](https://trello.com/c/sgJpWkPw/2-bug-1104779-deepsea-stage-1-fails-in-proposal-generate-in-case-lshw-or-hwinfo-is-not-installed-but-deepsea-should-install-that-i)

jschmid1:
> @jschmid1 does this actually fix the bug you mentioned? hwinfo is not provided by smartmontools and the hwinfo package is also not required by it either. (same for lshw I assume)

hwinfo/lshw is usually part of the base installation of their respective operating systems ( opensuse, ubuntu ) while smartmontools is not.

Technically we should also add hwinfo to 'SUSE' and lshw for 'Ubuntu'

jschmid1:
I think this won't _hurt_ so I'll add it.

tserong:
Do we also need to add hwinfo for centos?

jschmid1:
@tserong for centos it adds a repo for hwinfo, and installs it. See [here](https://github.com/SUSE/DeepSea/blob/28b57ec24aef396f68981bb4abb578e03120facf/srv/salt/ceph/packages/common/default.sls#L72)

tserong:
Ah, silly me, I just read the diff without noticing hwinfo was already there for CentOS :-/

smithfarm:
@susebot run teuthology

jschmid1:
@susebot run teuthology

PR #1275, opened 2018-08-15
qa/deploy: handle possibility that deepsea-cli is not installed

smithfarm:
When installing DeepSea from packages, the "deepsea-cli" package might
not get installed (it is not strictly required), and depending on which tests
are being run, the absence of the CLI might not be a problem.

In any case, there is no reason why failure of "deepsea --version" should
fail the entire test.

Fixes #

Description:

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/qa/common/deploy.sh
+++ b/qa/common/deploy.sh
@@ -122,7 +122,7 @@ function initialization_sequence {
     python --version || true
     python2 --version || true
     python3 --version
-    deepsea --version
+    deepsea --version || true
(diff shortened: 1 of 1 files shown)


smithfarm:
@susebot run teuthology

jschmid1:
The failure here is expected:

```
 No matching sls found for 'ceph.functests.1node.replace' in env 'base'
```

PR #1278, opened 2018-08-16
rescind: prevent empty rendering of ceph.rescind.tuned state file

smithfarm:
(cherry picked from commit 4789754a253e1700406ad19c0b2376e47fffd7ca)

### DIFF
--- a/srv/salt/ceph/rescind/tuned/default.sls
+++ b/srv/salt/ceph/rescind/tuned/default.sls
@@ -1,3 +1,7 @@
+prevent empty rendering:
+  test.nop:
+    - name: skip
+
 {% set roles = salt['pillar.get']('roles') %}
 
 {% if 'storage' not in roles and 'mon' not in roles and


smithfarm:
@jschmid1 It used to be that the salt in SLE-12-SP3 was not exposed to this "empty rendering" phenomenon. But https://bugzilla.suse.com/show_bug.cgi?id=1104921 would seem to indicate otherwise. . .

jschmid1:
@smithfarm good catch, but how could be improve our testsuite to avoid empty rendering in the future alltogether.

We already run stage.5 now in all our tests, right? How could me missed that.

jschmid1:
> No, not yet. The test is written and in master, but it is not yet being run. Also, the test itself has not been backported to SES5. (I was on vacation all last week.)

woops, seems like I was getting ahead of ourselves. So we should be covered once the stage.5 is in place and backported?

smithfarm:
@jschmid1 Yes. The Stage 5 test coverage is there, but not yet in the CI (waiting for M7 dust to settle) and not yet backported (again, waiting for M7 dust to settle).

PR #1246, opened 2018-07-27
cli: filter out test.nop steps and don't report [init] step errors

rjfd:
Fixes: #1243

Description:

This PR introduces the following changes to fix the problems reported in #1243 
* filter out test.nop states from the execution plan
* [init] step errors are not reported in the failure summary

PR #1280, opened 2018-08-16
[backport][ses5] cli: filter out test.nop steps and don't report [init] step errors

smithfarm:
Backport of https://github.com/SUSE/DeepSea/pull/1246

### DIFF
--- a/cli/.coveragerc
+++ b/cli/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+omit = tests/*
--- a/cli/monitors/terminal_outputter.py
+++ b/cli/monitors/terminal_outputter.py
@@ -151,7 +151,7 @@ class SimplePrinter(MonitorListener):
         self.current_step.append({'endl': False})
 
     def step_runner_finished(self, step):
-        if not step.success:
+        if step.order > 0 and not step.success:
             if step.name not in self.errors:
                 self.errors[step.name] = step.end_event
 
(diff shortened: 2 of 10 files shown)


smithfarm:
I guess this is relevant for SES5 after all. . .

smithfarm:
The new tests introduced by this PR are failing with

```
E   SaltClientError: Could not access /var/cache/salt. Please give jenkins read permissions.
```

I'm guessing some other commit needs to be cherry-picked, but which one?

smithfarm:
@jschmid1 It looks like they *are* excluded in master PR test runs, but by what mechanism? In other words, what is missing to exclude them for SES5 PR test runs as well?

jschmid1:
I think it should be that https://github.com/SUSE/DeepSea/blob/master/tox.ini#L17-L17

smithfarm:
@jschmid1 Passing Jenkins now: please have another look?

smithfarm:
@alexandergraul It seems like you're saying we could just drop the py-coverage/codecov commits and fix the unit tests by simply adding `--ignore cli/` to the relevant line in tox.ini. Correct?

If you guys think that's a better way to go, I'll be happy to do it.

alexandergraul:
@smithfarm yes, the only relevant part of e931102 and 1d52c88 is ` --ignore cli/` unless I am missing something.
I think it would be better to just add that without those two commits to keep the intention clear and only change what is actually needed.

PR #1294, opened 2018-08-21
cephinspector: assume filestore OSD if type file not present (bsc#1104773)

tserong:
Engulfing a cluster with OSDs that were created before the Ceph Jewel release will result in empty profile files, because the OSD's 'type' file doesn't exist.

-----------------

**Checklist:**
- [ ] Adapted documentation


tserong:
>facepalm<

Issue #1264, opened 2018-08-14
engulf does not fill profile import and does not use persistent names for scsi devices


Martin-Weiss:
### Description of Issue/Question
After an engulf of an existing SES 4 cluster the profile-import yml files are empty - just one has a single OSD in and that is using non-persistent device names

### Setup
Admin on SES 5 and cluster on SES 4. Patches applied up to end of august 2018

### Steps to Reproduce Issue
engulf an existing ses 4 cluster
Hint: we have dell hardware with RAID 0 setup for all HDDs.

### Versions Report
0.8.5

jschmid1:
![](https://github.trello.services/images/mini-trello-icon.png) [Bug 1104773 - "type" file is missing in /var/lib/ceph/osd/ceph-*/ after upgrading SES step by step to SES 4 causing engulf to fail in creation of OSD proposal yml files](https://trello.com/c/N906DSp7/3-bug-1104773-type-file-is-missing-in-var-lib-ceph-osd-ceph-after-upgrading-ses-step-by-step-to-ses-4-causing-engulf-to-fail-in-cr)

PR #1299, opened 2018-08-23
cephinspector: use cephdisks.device in engulf (bsc#1082994)

tserong:
This ensures the same code is used for determining /dev/disk/by-id paths
during engulf, as is used when invoking stage.1 to generate profiles.

Fixes: https://github.com/SUSE/DeepSea/issues/1264

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/salt/_modules/cephinspector.py
+++ b/srv/salt/_modules/cephinspector.py
@@ -73,18 +73,14 @@ def _get_device_of_partition(partition):
 def _get_disk_id(partition):
     """
     Return the disk id of a partition/device, or the original partition/device if
-    the disk id is not available.
+    the disk id is not available.  This is essentially the same thing as _uuid_device()
+    in srv/salt/_modules/osd.py
     """
-    disk_id_cmd = Popen("find -L /dev/disk/by-id -samefile " + partition +
-                        r" \( -name ata* -o -name nvme* \)", stdout=PIPE,
-                        stderr=PIPE, shell=True)
-    # pylint: disable=unused-variable
-    out, err = disk_id_cmd.communicate()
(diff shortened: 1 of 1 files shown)


tserong:
I agree, the engulf desperately needs automated tests, probably both unit and functional (e.g. I'd love a test that did a regular deploy then engulfed itself and verified equivalent profile and policy files were produced). That said, I've been doing manual testing of the SES5 and master branches, the SES5 branch looks OK, but I've found at least two other bugs in master, so I wonder if we can address other fixes to master and proper automated tests separately to this PR?

jschmid1:
> I agree, the engulf desperately needs automated tests, probably both unit and functional (e.g. I'd love a test that did a regular deploy then engulfed itself and verified equivalent profile and policy files were produced). That said, I've been doing manual testing of the SES5 and master branches, the SES5 branch looks OK, but I've found at least two other bugs in master, so I wonder if we can address other fixes to master and proper automated tests separately to this PR?

Fine by me. But let's keep track of that with a issue/trello-card. Creating one.

jschmid1:
Created #1304

smithfarm:
Ah, right: "Before starting the procedure below, you need to upgrade the Salt master node to SUSE Linux Enterprise Server 12 SP3 and SUSE Enterprise Storage 5 by running zypper migration (or your preferred way of upgrading)."

smithfarm:
I see the need to have engulf in the SES5 branch, but what is the use case for the engulf code in master?

Martin-Weiss:
I would love the engulf to work in SES 5, I.e. too to allow recovery of admin / storage profiles or to migrate non SES clusters in the future, too.

tserong:
What Martin said :-)

PR #1302, opened 2018-08-23
OSD Migration for changed disk slots

### LINKED ISSUE #1215: proposal.generate sorting

alexandergraul:
Fixes #1215

Please have a look already, there are still a few things to do (I've noted them below) but the core business logic should stay more or less the same (unless you find some problems :wink:)

Description:
OSD Migration currently only works if the same slot is used. This works
because the generation of proposals is deterministic, so if all disks
are in the same position we end up with the same proposal.

This commit changes that behaviour: instead of creating a new proposal,
the old one is used and adapted. This happens in the following steps:

1) When `salt-run replace.osd X` is used and a proposal file (e.g.
    "data1.ceph.yml" in /srv/pillar/proposals/...) gets the "-replace"
    suffix, a new attribute is written for the disks that get replaced
2) When `salt-run proposals.populate` is used (e.g. in stage.1), all
    proposal files that have the "-replace" suffix get parsed and the
    disks in the proposal are compared to the currently installed disks
    on the node.
3) Disks in the old proposal that have the attribute inserted in 1) and
    are still present (i.e. replaced in place) plus new disks (disks
    that are not in the old proposal but present) are considered free.
4) The disks in the old proposal get interated over and if they are
    labeled for replacement they are swapped with one of the free disks.
5) This modified proposal gets saved in the place of the original one
    and can be used by later stages

Things that should still be done:
* Have a way to roll the "destroyed" mark in the CRUSH map back if renaming and inserting of "replace: true" in proposal file fails
* Clean up and add more unittests
* Add functional tests 

-----------------

**Checklist:**
- [x] Added unittests ~and or functional tests~
- [ ] Adapted documentation
- [x] Ran integration tests successfully ( trigger with @susebot run teuthology )

### DIFF
--- a/srv/modules/runners/proposal.py
+++ b/srv/modules/runners/proposal.py
@@ -6,19 +6,24 @@ Generates the hardware profiles for a minion
 from __future__ import absolute_import
 from __future__ import print_function
 import pprint
-from os.path import isdir, isfile
+from os.path import isdir, isfile, join
 import os
+import re
+
 # pylint: disable=redefined-builtin
 from sys import exit
 import logging
+from collections import namedtuple
+
 # pylint: disable=import-error,3rd-party-module-not-gated,redefined-builtin
 import salt.client
 import yaml
+
 # pylint: disable=import-error
 
 log = logging.getLogger(__name__)
 
-USAGE = '''
+USAGE = """
 The proposal runner compiles and either shows or writes storage profile
 proposals. It offers two methods: 'peek' and 'populate'.
 The 'peek' method simply collects the proposals from the minions and displays
@@ -86,7 +91,7 @@ List of recognized parameters and their defaults:
                     number with a unit suffix. Unit suffix' as accepted by
                     sgdisk can be used, i.e. kibibytes (K), mebibytes (M),
                     gibibytes (G), tebibytes (T), or pebibytes (P).
-'''
+"""
 
 
 # pylint: disable=too-few-public-methods
@@ -94,44 +99,48 @@ class Target(object):
     """
     Wrapperclass around __utils__ to avoid early access to dunders
     """
+
     @staticmethod
     def show():
         """
         Wraps deepsea_minions.show
         """
-        return __utils__['deepsea_minions.show']()
+        return __utils__["deepsea_minions.show"]()
 
 
 class StdArgs(object):
     """
     Wrapper around STD_ARGS to avoid early expanding of Target().show()
     """
+
     @staticmethod
     def write_out():
         """
         Return STD_ARGS
         """
         return {
-         'leftovers': False,
-         'standalone': False,
-         'nvme-ssd-spinner': False,
-         'nvme-ssd': False,
-         'nvme-spinner': False,
-         'ssd-spinner': False,
-         'ratio': 5,
-         'db-ratio': 5,
-         'target': Target().show(),
-         'data': 0,
-         'journal': 0,
-         'wal': 0,
-         'name': 'default',
-         'format': 'bluestore',
-         'encryption': '',
-         'journal-size': '5g',
-         'db-size': '500m',
-         'wal-size': '500m'}
+            "leftovers": False,
+            "standalone": False,
+            "nvme-ssd-spinner": False,
+            "nvme-ssd": False,
+            "nvme-spinner": False,
+            "ssd-spinner": False,
+            "ratio": 5,
+            "db-ratio": 5,
+            "target": Target().show(),
+            "data": 0,
+            "journal": 0,
+            "wal": 0,
+            "name": "default",
+            "format": "bluestore",
+            "encryption": "",
+            "journal-size": "5g",
+            "db-size": "500m",
+            "wal-size": "500m",
+        }
 
-BASE_DIR = '/srv/pillar/ceph/proposals'
+
+BASE_DIR = "/srv/pillar/ceph/proposals"
 
 
 def _parse_args(kwargs):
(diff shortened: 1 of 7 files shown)


srv/modules/runners/proposal.py
```diff
@@ -308,6 +310,159 @@ def _record_filter(args, base_dir):
...
+
+        Following steps take place:
+        (0.) Use a new disk (by-path) for the proposal if the physical location
+            has changed.
+        1. Remove all "replace: " attributes from the proposal
+        2. Write proposal to the "normal" proposal file
+        3. Remove proposal file with "-replace" in its name
+        """
+
+        # import ipdb; ipdb.set_trace()
```
alexandergraul: I'll remove this, that should not be in here

srv/modules/runners/proposal.py
```diff
@@ -308,6 +310,159 @@ def _record_filter(args, base_dir):
...
 
 
+def _find_minions_to_replace(profile_dir):
+    """
+    Search through a given profile directory for files that end with '-replace'.
+
+    :param profile_dir: Profile directory, e.g. "/srv/pillar/ceph/proposals/profile-default"
+    """
+    ToReplace = namedtuple('ToReplace', ['fullpath', 'filename'])
+    dir = '{}/stack/default/ceph/minions'.format(profile_dir)
```
jan--f: ```dir``` is a python built-in function. maybe use ```dir_``` or ```_dir``` to avoid confusion?

srv/modules/runners/proposal.py
```diff
@@ -308,6 +310,159 @@ def _record_filter(args, base_dir):
...
+    files = [f for f in os.listdir(dir) if isfile(join(dir, f))]
```
jan--f: You could add the ```f.endswith('-replace')``` here and thus only iterate once over all profile files. Somehting like
```files = [f for f in os.listdir(dir) if isfile(join(dir, f)) and f.endswith('-replace')]```

**Comment by alexandergraul** [comment]
[Adapting "replacement procedure"](https://trello.com/c/QwcfG4jd/11-adapting-replacement-procedure)

File: srv/modules/runners/replace.py [hunk 4]
```diff
@@ -147,15 +149,45 @@ def minion_profile(minion):
...
+                    os.rename(filename, "{}-replace".format(filename))
+                    _insert_replace_tag(grains, minion, osds, "{}-replace".format(filename))
+                except:
+                    log.error("Failed to rename minion {} profile".format(minion))
+                    os.rename("{}-replace".format(filename), filename)
+
     return ""
 
 
+def _strip_enddigits(string):
```
jan--f: Doe this do somehting different then ```string.rstrip('0123456789')```? If I'm not missing anything this method is unnecessary.

srv/modules/runners/proposal.py
```diff
@@ -317,28 +472,37 @@ def populate(**kwargs):
...
-
-    return True
+    if isdir(profile_dir):
+        minions_to_replace = _find_minions_to_replace(profile_dir)
+
+    if minions_to_replace:
+        for minion in minions_to_replace:
+            ReplaceDiskOn(minion).replace()
+        return True
+    else:
```
jan--f: Does this mean that if a replacement is ongoing no new proposals are created? Wondering if that'll bite us at some point when someone replaces a disk and adds a new node in one go...

srv/modules/runners/proposal.py
alexandergraul: oh, right! Will change that.

srv/modules/runners/proposal.py
alexandergraul: Good catch, that does not need two iterations over all profiles

srv/modules/runners/replace.py
alexandergraul: No, this is a case of "do before properly looking if there is an built-in that does the same" :laughing: I'll adapt the code.

srv/modules/runners/proposal.py
alexandergraul: That is actually a good point I did not think about. I guess we could remove the minions in minions_to_replace from the targets and run `proposal.generate` on the rest.

srv/modules/runners/replace.py
```diff
@@ -52,23 +53,24 @@ def osd(*args, **kwargs):
 
     local = salt.client.LocalClient()
     host_osds = local.cmd('I@roles:storage', 'osd.list', tgt_type='compound')
+    assert isinstance(host_osds, list)
```
jschmid1: `local.cmd('I@roles:storage', 'osd.list', tgt_type='compound')` returns a dict of host: list_of_osds

srv/salt/_modules/osd.py
```diff
@@ -1793,8 +1793,7 @@ def partitions(self, osd_id):
 
         for device_type in ['journal', 'block', 'block.db', 'block.wal', 'block_dmcrypt']:
             result = self._uuid_device("{}/{}".format(mount_dir, device_type))
-            if result:
-                _partitions[device_type] = result
+            _partitions[device_type] = result
```
jschmid1: the `if result:` is still needed as the `partitions` method also feeds the grains.

jschmid1:
Tried:

1) salt-run replace.osd 1
stage 1,2,3
got the expected result

2) salt-run replace.osd 1 5(5 being on another host)
stage 1,2,3
got the expected result

3) salt-run replace.osd 1 5(5 being on another host)
a node is about to be added ( not deployed, no proposal generated )
stage 1,2,3
got the expected result

jschmid1:
lgtm after those two flapping unittests are fixed.

jschmid1:
Real hardware yields some problems. In-depth explanation following

jschmid1:
@susebot run teuthology

alexandergraul:
retest this please

jschmid1:
8/9 failures are infrastructure related

PR #1289, opened 2018-08-20
[backport][ses5] qa: refactor health-ok.sh, introduce storage profile and Stage 5 tests

smithfarm:
**Checklist:**
~~- [ ] Added unittests and or functional tests~~
~~- [ ] Adapted documentation~~
~~- [ ] Referenced issues or internal bugtracker~~
~~- [ ] cherry-pick #1297 into this PR~~
- [x] Enable susebot to trigger custom suites, e.g.  `@susebot run teuthology for suse/tier2`
- [x] Test suse/tier1/functional
- [x] Test suse/tier2
- [x] port health-igw.sh, health-nfs-ganesha.sh, and health-openattic.sh to health-ok.sh
- [x] suse/tier1/igw and suse/tier1/nfs-ganesha pass
- [x] forward-port health-nfs-ganesha.sh migration to master (as part of https://github.com/SUSE/DeepSea/pull/1305)
- [x] suse/tier1/openattic passes
- [x] open PR with forward-port of "qa: more flexible role assignment in policy.cfg"
- [x] add more CLI test cases
- [x] https://github.com/SUSE/DeepSea/pull/1257 merged
- [x] cherry-pick `--no-update` patch from #1308
- [x] Adapt deepsea-pr.sh to run `suse/tier1` instead of `suse/smoke`
- [x] suse/tier1 running "acceptably"
- [x] Merge SUSE/ceph.git wip-qa-ses5 into ses5
- [x] Ran integration tests successfully ( trigger with @susebot run teuthology )

PR #1305, opened 2018-08-23
stage/prep: reinstate single quotes around deepsea_minions tgt, plus some qa fixes

smithfarm:
Typically, deepsea_minions will be set to `'*'`. Even if it's not set to exactly that, it is very likely to
contain a `*` character. That's why it needs to be enclosed in single-quotes.
Without the single quotes, Stage 0 fails like so:

```
    Data failed to compile:
----------
    Rendering SLS 'base:ceph.stage.0.minion.default' failed: expected alphabetic or numeric character, but found '\n'; line 3

---
repo:
  salt.state:
    - tgt: *    <======================
    - tgt_type: compound
    - sls: ceph.repo

metapackage minions:
  salt.state:
[...]
```

Also add some miscellaneous qa commits. One fixes the following error:

```
suites/basic/health-nfs-ganesha.sh: line 71: break: only meaningful in a 'for',
'while', or 'until' loop
```

Another gets rid of the separate `health-nfs-ganesha.sh` qa script by moving its functionality into the main `health-ok.sh` script.


smithfarm:
Backport included in #1289

smithfarm:
@susebot run teuthology

smithfarm:
@susebot run teuthology suse/tier1/functional

smithfarm:
@susebot run teuthology

smithfarm:
@jschmid1 5 tests failed (out of 14) and the failures were expected given the current state of the CI.

PR #1308, opened 2018-08-24
qa: more flexible role assignment in policy.cfg

smithfarm:
Spread out the {openattic,rgw,igw,ganesha} roles among different cluster nodes,
if possible.

Signed-off-by: Nathan Cutler <ncutler@suse.com>

Conflicts:
	qa/common/deploy.sh

PR #1309, opened 2018-08-24
qa/health-ok: drop --fsal option (no longer needed)

smithfarm:
Forward-port of a commit from #1289


smithfarm:
Ah, nevermind. Will put this in #1308 since it fixes a regression introduced by that PR.

PR #1334, opened 2018-08-29
Improve migration tests

### LINKED ISSUE #1324: Fix and extend tests for OSD Migration
#### Follow up of #1302

There are two flickering tests that should be fixed and the tests can be updated to cover more setups.

alexandergraul:
Fixes #1324

### Description

Refactored tests and test data so new fake minions can be added easily. I migrated the test data from virtual disk paths to disk paths from blueshark.

~~### Do not merge~~

~~I am not sure if the tests work in CI but already open the PR to try it out.~~

Edit: Looks like sorting the proposal is needed because the comparison to the pytest fixture fails otherwise

-----------------

**Checklist:**
- [x] ~~Added~~ Adapted unittests ~~and or functional tests~~
~~- [ ] Adapted documentation~~

### DIFF
--- a/tests/unit/runners/test_proposal.py
+++ b/tests/unit/runners/test_proposal.py
@@ -8,6 +8,8 @@ sys.path.insert(0, "srv/modules/pillar")
 
 
 NUM_MINIONS = 4
+
+
 @pytest.fixture
 def minions():
     minion_tuple = namedtuple("minion_tuple", ["fullpath", "filename"])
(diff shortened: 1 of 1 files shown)


alexandergraul:
tests still fail (sometimes) on first run but not on second :/

alexandergraul:
that is also weird: make-rpm failed but make-test passed. this inconsistency is getting annoying

alexandergraul:
retest this please

alexandergraul:
when using tox to only create the virtualenv and then sourcing it and running the test inside it manually the tests pass on first try (locally on my machine)

Issue #1358, opened 2018-09-06
add unittests to replace.py/remove.py runners


jschmid1:
### Description of Issue/Question

After finding bugs in there, it's time to add unittests.

PR #1359, opened 2018-09-06
remove/replace interference fixes

### LINKED ISSUE #1357: remove.osd and replace.osd interfere [tracker][wip]

`remove.osd` internally calls `replace.osd` which in-turn renames the minion profile.

Previously we relied on having a \<profile>-replace file in the profiles-dir that signaled the next iteration of the proposal.populate runner to re-create a proposal.
With the change in #1302 we moved to a more surgically approach where we _actually replace_ the contents of the profile. (due to the findings in #1215)
Adapting to this method we should remove the matching entry in the profile in a `remove` operation.

wip-branch: wip-remove-profile-surg

jschmid1:
Description:

* pass a \<context> variable in kwargs that indicate if
  we have a replace or remove operation
* delete the key from the profile yaml file when removing a osd
* Retrun a meaninful message when a osd is not in the
  cluster but you try to remove/replace it.
* refactor(extract) _get_profile_keys

Fixes #1357 

-----------------

**Checklist:**
~- [ ] Added unittests and or functional tests~ postponed #1358
- [x] Adapted documentation

### DIFF
--- a/srv/modules/runners/replace.py
+++ b/srv/modules/runners/replace.py
@@ -46,6 +46,7 @@ def osd(*args, **kwargs):
     supported = ["force", "timeout", "delay"]
     passed = ["{}={}".format(k, v) for k, v in kwargs.items() if k in supported]
     log.debug("Converted kwargs: {}".format(passed))
+    context = kwargs.get('remove', 'replace')
 
     # OSDs to remove
     osds = list(str(arg) for arg in args)
@@ -69,7 +70,7 @@ def osd(*args, **kwargs):
                 continue
 
             # Rename minion profile
-            minion_profile(host, osds, grains)
+            minion_profile(host, osds, grains, context)
 
     if "called" in kwargs and kwargs["called"]:
         # Return for remove.osd
@@ -146,10 +147,11 @@ def _find_host(osd_id, host_osds):
     return ""
 
 
-def minion_profile(minion, osds, grains):
+def minion_profile(minion, osds, grains, context):
     """
     Rename a minion profile to indicate that the minion profile needs to be
-    recreated.
+    recreated if the replace context is given
+    Delete the entry in the minion's profile if the remove context is given
 
     Note: Nobody is required to have profile entries in the policy.cfg.  Some
     might be modifying their pillar data directly.  Also, the file will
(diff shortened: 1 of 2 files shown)


srv/modules/runners/replace.py
```diff
@@ -46,6 +46,9 @@ def osd(*args, **kwargs):
     supported = ["force", "timeout", "delay"]
     passed = ["{}={}".format(k, v) for k, v in kwargs.items() if k in supported]
     log.debug("Converted kwargs: {}".format(passed))
+    context = 'replace'
+    if kwargs.get('remove', False):
+        context = 'remove'
```
jan--f: why not `context = kwargs.get('remove', 'replace')`?

jschmid1:
@susebot run teuthology

srv/modules/runners/replace.py
```diff
@@ -164,16 +168,28 @@ def minion_profile(minion, osds, grains):
...
+                    try:
+                        print("Renaming minion {} profile".format(minion))
+                        os.rename(filename, "{}-replace".format(filename))
+                        _insert_replace_flag(
+                            grains, disks, minion, osds, "{}-replace".format(filename)
+                        )
+                    # pylint: disable=bare-except
+                    except:
+                        log.error("Failed to rename minion {} profile".format(minion))
+                        os.rename("{}-replace".format(filename), filename)
```
jan--f: This seem strange. We revert the rename, if the rename has failed? Maybe we do expect multiple exceptions here?

alexandergraul: I  can give some context (as the original author):

The log message is a bit misleading, that is right. Two things can fail: renaming the file or inserting the replace flags. In each case, the file is brought back to the state before it got renamed and a "generic" error message is provided. This can definitely be improved.

The bare except is here because I am not quite sure what exceptions we are looking for (probably IO and dict look up related)

jschmid1: This was initially intended to act as sort of a 'rollback' mechanism for the `_insert_replace_flag` method. I think this can stay in there for now as it should not cause any harm. If any of the above operations fail, we just end up with what we had.

jschmid1:
@susebot run teuthology

srv/modules/runners/replace.py
jan--f: Well I'm wondering...if the `os.rename(src, dest)` fails (OSError if `dest` exists) we then take `dest` and try to rename it to `src`. This will throw another exception since `src` exists. Granted this shouldn't happen™ (or is somewhat unlikely) but it defeats the purpose of the exception handling. Have I missed anything?
If not I'd vote for proper exception handling here. The rollback-rename should only happen for `KeyError` coming from `_insert_replace_flag`. OSError should log a message but probably rethrow the exception (or is there a sensible way to handle it and continue?). The potential IOError (from `open` in `_insert_replace_flag`) should bubble up imho.

jan--f: Why not `except KeyError` then (and maybe YAMLError in case the dump can gor wrong?)? If nothing else it makes the intent and assumptions obvious to the next person.
I will concede this can be a code hygiene fix in another PR.

jschmid1:
```python
try:
   ...
except FirstException:
   handle_first_one()

except SecondException:
   handle_second_one()
```

I wasn't aware that this is even possible :>

I'll give in if we move this to a separate PR.

PR #1380, opened 2018-09-08
[backport][SES5] Only restart with a running cluster and  functests: fix problems that occur when process unexpectedly not running

smithfarm:
Backports:

* #1210 
* #1381 (WARNING: NOT MERGED YET)

---

- [x] Referenced issues or internal bugtracker
- [x] Ran integration tests successfully (trigger with "@susebot run teuthology" in a GitHub comment; see the [wiki](https://github.com/SUSE/DeepSea/wiki/Testing) for more information)

PR #1381, opened 2018-09-08
functests: fix problems that occur when process unexpectedly not running

smithfarm:
The restart tests were assuming that "pgrep $CEPH_DAEMON" would always return a pid. See the commit messages for more rationale.

SES5 backport is included in https://github.com/SUSE/DeepSea/pull/1380

-----------------

**Checklist:**

### DIFF
--- a/srv/salt/ceph/tests/restart/rgw/same_pid.sls
+++ b/srv/salt/ceph/tests/restart/rgw/same_pid.sls
@@ -1,7 +1,7 @@
 
-Check pid:
+Check pid radosgw:
   cmd.run:
-    - name: "[ `pgrep radosgw` ==  `cat /tmp/restart.pid` ]"
+    - name: "test \"$(pgrep radosgw)\" = \"$(cat /tmp/restart.pid)\""
     - failhard: True
 
 /tmp/restart.pid:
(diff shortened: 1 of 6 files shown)


srv/salt/ceph/tests/restart/rgw/same_pid.sls
```diff
@@ -1,7 +1,7 @@
 
 Check pid:
   cmd.run:
-    - name: "[ `pgrep radosgw` ==  `cat /tmp/restart.pid` ]"
+    - name: "test \"$(pgrep radosgw)\" = \"$(cat /tmp/restart.pid)\""
```
smithfarm: Another possibility would be to use the bash built-in `[[ . . . ]]` which doesn't require the variables to be in double-quotes.

Remember: running shell commands via `cmd.run` is fun!

smithfarm:
@susebot run teuthology

smithfarm:
@susebot run teuthology

srv/salt/ceph/tests/restart/save_pid.sls
```diff
@@ -1,6 +1,10 @@
 
-Save pid:
+Save pid {{ service }}:
   cmd.run:
     - name: "pgrep ceph-{{ service }} > /tmp/restart.pid"
```
smithfarm: If `ceph-{{ service }}` is not running, this command will create an empty file `/tmp/restart.pid`

It returns status code 1, but Salt ignores that. . .

smithfarm:
Merging upon agreement in the DS call.

PR #1388, opened 2018-09-12
qa/deploy: fix error-reporting issues

smithfarm:
Issue #1: When the command salt-key -L -l acc | grep -v '^Accepted Keys' fails,
the script was aborting immediately due to "set -e".

Issue #2: When running in an older version of SLES, python3 might not be
installed, and "python3 --version || true" was putting a confusing error
message into the log:

    /usr/lib/deepsea/qa/common/deploy.sh: line 128: python3: command not found

Fixes #

Description:

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/qa/common/deploy.sh
+++ b/qa/common/deploy.sh
@@ -38,7 +38,6 @@ function _set_deepsea_minions {
 }
 
 function _initialize_minion_array {
-    set +x
     local m=
     local i=0
     if type salt-key > /dev/null 2>&1; then
@@ -53,7 +52,6 @@ function _initialize_minion_array {
         exit 1
     fi
     echo "There are $i minions in this Salt cluster"
-    set -x
 }
 
 function _update_salt {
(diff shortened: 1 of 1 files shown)


qa/common/deploy.sh
```diff
@@ -38,7 +38,6 @@ function _set_deepsea_minions {
 }
 
 function _initialize_minion_array {
-    set +x
```
kshtsk: Not sure how this will improve the logging.

```
salt-key -L -l acc | grep -v ...
```

fails, it does not show nothing, and since '-e' is enabled, it will exit from the script.

kshtsk: Oh, it will log which command failed.
Then looks good for me, but initial intention was to hide keys, right?

qa/common/deploy.sh
```diff
@@ -118,17 +116,22 @@ function _zypper_ps {
     salt '*' cmd.run 'zypper ps -s' 2>/dev/null || true
 }
 
+function _python_versions {
+    type python2 > /dev/null 2>&1 && python2 --version || echo "Python 2 not installed"
+    type python3 > /dev/null 2>&1 && python3 --version || echo "Python 3 not installed"
+}
+
 function initialization_sequence {
     set +x
```
smithfarm: @kshtsk `_initialize_minion_array` is called from this function (`initialization_sequence`), and `set +x` is in effect for the entire function. That's why I dropped it from `_initialize_minion_array`.

I suggest we try this and see if the logs are more readable/less confusing.

PR #1389, opened 2018-09-13
build: Remove duplicate functests `install` lines

tserong:
8e2592c2 seems to have resulted in a bunch of duplicate `install` lines
in the Makefile; there's two blocks labelled "# functests/1node", the
second being a superset of the first, so I've gone and removed the first.


jschmid1:
@susebot run teuthology

jschmid1:
@susebot run teuthology

jschmid1:
Doesn't seem the case in SES5. No backport required. :+1:

PR #1403, opened 2018-09-26
proposal: remove target and std_args globals (bsc#1100609)

tserong:
When invoking `salt-run`, all runners are loaded, and thus all globals
in those runners are evaulated.  The proposal runner was invoking
deepsea_minions.DeepseaMinions() to set the target global, which results
in a pillar refresh and a couple of pillar gets across all nodes.  If
any nodes were down, these three operations would each take 15 seconds
to time out, resulting in a delay of up to 45 seconds invoking any
runner if any nodes were down.  It also added a slight (2-3 second)
delay in the normal case where all nodes are up.

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/modules/runners/proposal.py
+++ b/srv/modules/runners/proposal.py
@@ -92,33 +92,31 @@ List of recognized parameters and their defaults:
                     gibibytes (G), tebibytes (T), or pebibytes (P).
 '''
 
-target = deepsea_minions.DeepseaMinions()
-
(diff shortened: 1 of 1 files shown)


kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
@susebot run teuthology

PR #1404, opened 2018-09-27
build: Add %pretrans script to deal with dir->symlink change

tserong:
Commit 8e2592c2 changed the directory /srv/salt/ceph/remove/migrated into
a symlink to /srv/salt/ceph/remove/destroyed.  RPM can't cope with directories
becoming symlinks (see https://bugzilla.redhat.com/show_bug.cgi?id=447156),
so we have to add a %pretrans script to the spec file to delete the old
'migrated' directory if it exists before RPM tries to do anything else.
Without this, the package can't be upgraded (it will fail complaining of
either file conflicts, or cpio rename failures, depending on whether you
do the upgrade with `rpm` or `zypper`).


smithfarm:
backport PR already open: https://github.com/SUSE/DeepSea/pull/1405

let's hope this PR doesn't see any changes before it merges...

PR #1405, opened 2018-09-27
[backport][ses5] build: Add %pretrans script to deal with dir->symlink change

tserong:
backport of https://github.com/SUSE/DeepSea/pull/1404

---

Commit 8e2592c2 changed the directory /srv/salt/ceph/remove/migrated into
a symlink to /srv/salt/ceph/remove/destroyed.  RPM can't cope with directories
becoming symlinks (see https://bugzilla.redhat.com/show_bug.cgi?id=447156),
so we have to add a %pretrans script to the spec file to delete the old
'migrated' directory if it exists before RPM tries to do anything else.
Without this, the package can't be upgraded (it will fail complaining of
either file conflicts, or cpio rename failures, depending on whether you
do the upgrade with `rpm` or `zypper`).

(cherry picked from commit 1dc5c9d7085467fce7f6b256eaf6b4eabc2f702c)


jschmid1:
@susebot run teuthology

PR #1410, opened 2018-10-04
Add cluster shutdown, startup functionality

swiftgist:
Fate: 323666

-----------------

**Checklist:**
- [x] Adapted documentation


srv/salt/ceph/shutdown/default.sls
```diff
@@ -0,0 +1,74 @@
+
+{% if salt['saltutil.runner']('disengage.check', cluster='ceph') == False %}
+safety is engaged:
+  salt.state:
+    - tgt: {{ salt['pillar.get']('master_minion') }}
```
tserong: Should this be `salt['master.minion']()`?

srv/salt/ceph/terminate/storage/default.sls
```diff
@@ -0,0 +1,10 @@
+
+{% for id in salt['osd.list']() %}
+
+removing {{ id }}:
```
tserong: s/removing/stop/

srv/salt/ceph/start/mds/default.sls
```diff
@@ -0,0 +1,6 @@
+
+start mds {{ grains['host'] }}:
+  service.running:
+    - name: ceph-mds@{{ grains['host'] }}
+
```
jschmid1: any particular reason why you don't do the `test -f ` check here?

srv/salt/ceph/start/mgr/default.sls
```diff
@@ -0,0 +1,5 @@
+
+start mgr {{ grains['host'] }}:
+  service.running:
+    - name: ceph-mgr@{{ grains['host'] }}
+
```
jschmid1: any particular reason why you don't do the `test -f ` check here?

srv/salt/ceph/start/mon/default.sls
```diff
@@ -0,0 +1,6 @@
+
+start mon:
+  service.running:
+    - name: ceph-mon@{{ grains['host'] }}
+
```
jschmid1: any particular reason why you don't do the `test -f ` check here?

srv/salt/ceph/start/storage/default.sls
```diff
@@ -0,0 +1,9 @@
+
+{% for id in salt['osd.list']() %}
+
+starting {{ id }}:
+  service.running:
+    - name: ceph-osd@{{ id }}
+
```
jschmid1: any particular reason why you don't do the `test -f ` check here?

jschmid1:
left some comments inline. 

We should also add the functests to the `srv/salt/ceph/functests/1node/init.sls` file.

jschmid1:
[Graceful system shutdown for all components](https://trello.com/c/WAWUnpRS/88-graceful-system-shutdown-for-all-components)

srv/salt/ceph/shutdown/default.sls
swiftgist: probably :)

srv/salt/ceph/terminate/storage/default.sls
swiftgist: easy enough

swiftgist:
For the "test -f", I followed what we did with respect to restarts and such.  Mandatory services such as mon, mgr, etc. are always installed.  If somehow a role is assigned to a mon and yet `ceph-mon` is missing, I don't think we should pass.  That's a problem for a startup or shutdown.

The Stage 4 services are technically optional and an admin could be in the "I've configured Stage 2 and intend to add two more RGW nodes, but I need to do a graceful shutdown right now".

swiftgist:
There is a question about mds.  It's there because of ceph-common.  Now, we consider it a service because CephFS is optional.  Do we add a no-op test for mds?  Is that less confusing?  Or do we just need a comment in that state file explaining this?

swiftgist:
Addressed @tserong changes.

srv/salt/ceph/startup/default.sls
```diff
@@ -0,0 +1,52 @@
...
+
+{% if salt.saltutil.runner('select.minions', cluster='ceph', roles='igw') %}
+Starting iscsi:
+  salt.state:
+    - sls: ceph.start.igw
+    - tgt: "I@roles:igw and I@cluster:ceph"
+    - tgt_type: compound
+{% endif %}
+
+{% if salt.saltutil.runner('select.minions', cluster='ceph', roles='mds') %}
```
tserong: I just realised something ... This (and other 'select.minions' blocks) shouldn't really be necessary, because if there's no mds roles, the tgt line below won't match any hosts. Or is there some other effect I'm forgetting?

tserong:
> There is a question about mds. It's there because of ceph-common. Now, we consider it a service because CephFS is optional. Do we add a no-op test for mds? Is that less confusing? Or do we just need a comment in that state file explaining this?

Possibly I'm being dense, but I don't understand the question :-)

srv/salt/ceph/startup/default.sls
swiftgist: Salt will complain about "target matched no minions" and also count it as a failure.

srv/salt/ceph/start/ganesha/default.sls
```diff
@@ -0,0 +1,7 @@
+
+start nfs-ganesha:
+  service.running:
+    - name: nfs-ganesha
+    - enable: False
```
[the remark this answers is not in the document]
swiftgist: fixed... so much for using the original

swiftgist:
@jschmid1 asked the question about the inconsistency of applying `test -f` to the different components.  My argument was that we only needed the check on the services that are optional.  The mandatory services always have a binary present so no need to check.

That follows Stage 3 vs. 4 conventions as well until you get to mds.  I checked and was wrong about ceph-mds being in ceph-common.  It's has its own package ceph-mds.  However, we still have it installed everywhere.  Do we add a `test -f /usr/bin/ceph-mds`check that we know will always succeed to keep the statement "we only check optional services"?  That seems unnecessary to me.

srv/salt/ceph/startup/default.sls
tserong: Ah, of course. Thanks.

tserong:
Now I understand. Yeah, I'd not bother with the `test -f` for ceph-mds, given the way the packaging works (we install the ceph package, which in turn requires ceph-mgr, ceph-mds, ceph-mon, ceph-osd).

jschmid1:
@susebot run teuthology

smithfarm:
@jschmid1 I updated ceph from 13.2.1 to 13.2.2 and it looks like the f:c:m repo was republished during the run of that job.

smithfarm:
@susebot run teuthology

smithfarm:
Note there is also a connectivity problem with ceph-ci.suse.de which might prevent zypper from installing RPMs. We'll be looking into that today.

smithfarm:
@kshtsk @jschmid1 OK, so I think I know what the problem is here. The DeepSea RPM building process is divided into two containers: "base" and "rpm". The filesystems:ceph:mimic repo is added when the first container ("base") is built.

As it is now, the deepsea-rpm-builder job will continue to fail until the base container is rebuilt. I will open a PR to fix the issue.

smithfarm:
https://github.com/SUSE/sesci/pull/5

jschmid1:
@susebot run teuthology

PR #1174, opened 2018-05-29
Refactor cephprocesses module and add tests for it

jschmid1:
Signed-off-by: Joshua Schmid <jschmid@suse.de>

Targets #1150 
Fixes #1157 

Description:

The cephprocesses module existed to identify 'down' services in order to stop deepsea form causing harm to the cluster in a sequential orchestration run. The reporting just works fine for all services besides the OSDs. If you happen to have 1k+ OSDs and Deepsea tells you 'There is at least one OSD down' you are kind of puzzled. This patch tries to find the osd_id from the psutil object to identify 'down' OSDs better.

The old implementation was also not really unit-testable, so I made some adjustments to make it more readable and testable. Added 40-45~ tests to the module.

~The implementation of the blacklist feature needs some more discussion though..~

make test:
  - successful

make lint:
  - successful

smoketests:
  - successful

Sidenote:

~Added a fix for the smoketests/init.sls file for RPMs~ rebased

-----------------

[Please find more information on how to run integration tests here](https://github.com/SUSE/DeepSea/wiki/Integration-tests)

PR #1423, opened 2018-10-10
fix: only check OSDs after restarting an OSD (bsc#1110759)

jschmid1:

suse_internal: bsc#1110759

For some reason it makes a different to salt in which context the module is executed from:

from the CLI the cephprocesses.check works as expected.

in the context of $changed_file the functions goes crazy and returns:

```
 {'down': ['ceph-radosgw', 'rgw', 'nfs-ganesha'],
 'up': {'rpc.statd': [96081], 
        'ceph-mds': [41269], 
        'rpcbind': [96061],  
        'ganesha.nfsd': [689925],
        'radosgw': [689097],
        'ceph-osd': [715680, 715777, 719338, 719435, 719532]}}
```

By adding the kwargs: {roles: ['storage']} we bypass the failing checks. This can be considered a workaround. We only loose imo unnecessary checks as a restart of a osd should not correlate with any other service. (I guess)

-----------------

**Checklist:**

### DIFF
--- a/srv/salt/ceph/osd/restart/controlled/default.sls
+++ b/srv/salt/ceph/osd/restart/controlled/default.sls
@@ -10,6 +10,8 @@ restart osd {{ id }}:
 wait on processes after processing osd.{{ id }}:
   module.run:
     - name: cephprocesses.wait
+    - kwargs:
+        'roles': ["storage"]
     - fire_event: True
     - failhard: True
 


jschmid1:
The **why** is still an open :question:

tserong:
That means it's checking ganesha and rgw processes, which it should only be doing if those roles are assigned to the node in the pillar data. I don't think there's anything wrong with forcing this to only check storage roles, because in this context we really only care about the OSDs, regardless of if there are any other roles assigned. Still, I'd be checking WTF is going on with pillar data here, somehow...

jschmid1:
> That means it's checking ganesha and rgw processes, which it should only be doing if those roles are assigned to the node in the pillar data.

Right, and those roles _are_ indeed assigned to that node. The odd thing is that they are both 'up' and 'down' at the same time (compare the output in the pr body).

> I don't think there's anything wrong with forcing this to only check storage roles, because in this context we really only care about the OSDs

ack



> regardless of if there are any other roles assigned. Still, I'd be checking WTF is going on with pillar data here, somehow...

I'd be really curious to find the root cause of this ..

tserong:
> The odd thing is that they are both 'up' and 'down' at the same time (compare the output in the pr body).

I must be going blind. That is _bizarre_.

tserong:
Some discrepancy between actual process names, and the global processes array? Or the logic to determine what's down is busted (note that `zypper_ps()` injects extra process names into that array, some of which match the names of the down processes).

jschmid1:
> Some discrepancy between actual process names, and the global processes array? Or the logic to determine what's down is busted. Note that `zypper_ps()` injects extra process names into that array, some of which match the names of the allegedly "down" processes.

something in that area probably. I'd still like to get #1174  in. The maintainability/readability of  the current code makes it very hard to debug

jschmid1:
for now, I think this should workaround should suffice

tserong:
> for now, I think this should workaround should suffice

Yeah, but maybe open a separate issue to rejig the process stuff, just so it's tracked.

jschmid1:
Oh snap.. no tests were run.

jschmid1:
@susebot run teuthology

jschmid1:
[[Bug 1110106] New: cephprocesses.wait incorrectly assumes that just because a role is assigned, a process should be running](https://trello.com/c/Qn8l0mzR/70-bug-1110106-new-cephprocesseswait-incorrectly-assumes-that-just-because-a-role-is-assigned-a-process-should-be-running)

jschmid1:
[Bug 1110759 - Update to ses 5.5 using stage.0 fails in "wait on processes after processing osd.9"](https://trello.com/c/2dM46wik/75-bug-1110759-update-to-ses-55-using-stage0-fails-in-wait-on-processes-after-processing-osd9)

PR #1424, opened 2018-10-10
Remove hardcoded nvme devices, support generic convention

swiftgist:
Bcache and NVMe devices follow the same device naming convention.
Remove the hardcoded strings and check whether the device ends with a
digit to add the 'p' partition delimiter.  Likewise, truncate the 'p'
after separating the partitions when necessary.

bnc#1107925

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/tests/unit/_modules/test_osd.py
+++ b/tests/unit/_modules/test_osd.py
@@ -1194,6 +1194,16 @@ class TestOSDPartitions():
         assert type(ret) is int
 
     @mock.patch('srv.salt._modules.osd.glob')
+    def test__last_partition_digits(self, glob_mock):
+        glob_mock.glob.return_value = ['/dev/sdx11']
+        osd_config = OSDConfig()
+        test_module = helper_specs(module=DEFAULT_MODULE)()
+        obj = test_module.OSDPartitions(osd_config)
+        ret = obj._last_partition(osd_config.device)
+        glob_mock.glob.assert_called_with('/dev/sdx[0-9]*')
+        assert ret == 11
+
+    @mock.patch('srv.salt._modules.osd.glob')
     def test__last_partition_no_pathnames(self, glob_mock):
         glob_mock.glob.return_value = []
         osd_config = OSDConfig()


jschmid1:
looks good, but we might want to add a generic test case for that regex.

tserong:
The referenced bug mentions an error occurring when there's more than 10 disks, and https://bugzilla.suse.com/show_bug.cgi?id=1107925#c3 includes this backtrace:

````
suse01:~ # salt 'suse01*' osd.deploy
suse01.suse.com:
    The minion function caused an exception: Traceback (most recent call last):
      File "/usr/lib/python2.7/site-packages/salt/minion.py", line 1455, in _thread_return
        return_data = executor.execute()
      File "/usr/lib/python2.7/site-packages/salt/executors/direct_call.py", line 28, in execute
        return self.func(*self.args, **self.kwargs)
      File "/var/cache/salt/minion/extmods/modules/osd.py", line 1861, in deploy
        if not is_prepared(device):
      File "/var/cache/salt/minion/extmods/modules/osd.py", line 1926, in is_prepared
        if osdc.highest_partition(readlink(device), 'lockbox') != 0:
      File "/var/cache/salt/minion/extmods/modules/osd.py", line 1101, in highest_partition
        partitions = sorted([ re.sub(r"{}p?".format(device), '', p) for p in pathnames ], key=int, reverse=True)
    ValueError: invalid literal for int() with base 10: '1p2'
````

`r"{}p?".format(device)` looks like it only handles a single digit after the `p`, which I assume breaks if there's more than ten disks. Does this need to be addressed?

swiftgist:
@jschmid1 I looked back through the unit tests for _find_paths and I think they are covered.  Examples include device names like 'nvme0n1p100'.  

@tserong I didn't change that section, but the expression still works since it's not anchored.  I added a unit test for a two digit device.

jschmid1:
> @jschmid1 I looked back through the unit tests for _find_paths and I think they are covered. Examples include device names like 'nvme0n1p100'.

That's great. lgtm from my side then.

PR #1429, opened 2018-10-16
Refactor cephprocesses module and add tests for it V2

### LINKED ISSUE #1157: rework cephprocesses

The module cephprocesses.py became quite ugly to maintain and is not covered with tests. This is an issue to track ideas in order to improve the overall quality of this module.

Idea for this came from the need for a blacklist #1150 and the long-time FIXME in the code which would enable us to track the OSD_IDs alongside it's PIDs.

Assigning to self.

jschmid1:
* Add file based blacklisting

* disabled && non-running services will be considered as
  'not down' hence excluded from the checks.

Adds 60~ unittests

Fixes #1157 

TODO:

- [ ] Thinking about a way to utilize functests to test the 'disabled && not running' statement.
- [x] Adapt docs (especially the blacklist & is-disabled feature)

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/salt/_modules/cephprocesses.py
+++ b/srv/salt/_modules/cephprocesses.py
@@ -33,7 +33,6 @@ processes = {'mon': ['ceph-mon'],
              'rgw': ['radosgw'],
              'ganesha': ['ganesha.nfsd', 'rpcbind', 'rpc.statd'],
              'admin': [],
-             'openattic': ['httpd-prefork'],
              'client-cephfs': [],
              'client-iscsi': [],
              'client-nfs': [],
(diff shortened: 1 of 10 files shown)


jschmid1:
@susebot run teuthology

srv/salt/_modules/cephprocesses.py
```diff
@@ -49,101 +49,348 @@
...
+
+    def __repr__(self):
+        """
+        custom __repr__ to identify the class a bit easier while debugging
+        """
+        return "Process <{}>".format(self.name)
+
+    def _map_osd_proc_to_osd_id(self):
+        """
+        Looking in the list of open_files you can _often_ see
```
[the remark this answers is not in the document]
jschmid1: That seems nicer.. I wasn't aware of the `.cmdline` method. I'll look into it and adapt to it.

srv/salt/_modules/cephprocesses.py
```diff
@@ -49,101 +49,348 @@
...
+        In our usecase we get the data from the salt-master
+        as a central place to annotate blacklisted OSDs
+        The current implementation only allows OSDs.
+        Other functionality such as MON/MDS etc needs to be
+        discussed and implemented. Explanation in comments
+        below.
+        """
+        # Imho the __blacklist struct is not ideal..
+        # the problem we are facing here is that you can't say;
+        # blacklist mon-1 (by dns-name) and then map this to a process
```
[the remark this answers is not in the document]
jschmid1: I recently thought about this as to if we still need this anymore after implementing the 'is-disabled' checks:

The reason I implemented the ceph-osd blacklist feature was to a reasonable request from users who said they are sometimes in a situation where a faulty disk causes weird behavior on a single OSD (or multiple when a wal/db disk is affected). They didn't want to dispose or disable them yet but rather keep them for investigation.

I can't really think of a use-case for a mon/mds/etc.. Those services don't exist in such a multitude like OSDs do and hence should be replaced when a quorum/failover is not present anymore. Blacklisting is also quite dangerous if not used with caution. So, all-in-all I'm not 100% convinced to implement that for other services besides OSDs anymore.

What do you think?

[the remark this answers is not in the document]
tserong: Good point. Seems reasonable to me.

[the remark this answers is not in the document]
jschmid1: Ok, I'll change the identification method to use your approach anyways

jschmid1:
ready for merge @tserong ?

srv/salt/_modules/cephprocesses.py
```diff
@@ -49,101 +48,326 @@
...
+    @property
+    def is_disabled(self):
+        """
+        Reach out to systemctl and call is-enabled.
+        Property for this state
+        """
+        if not self.service_names:
+            return False
+
+        for fsn in self.service_names:
```
jan--f: I'd be really happy about a comment that this loop is here for nfs-ganesha, since it depends(?) on rpcbind. would it maybe be better to check for rpcbind if service_name == nfs-ganesha?

srv/salt/_modules/cephprocesses.py
```diff
@@ -49,101 +48,326 @@
...
+
+    def _service_names(self):
+        """
+        The construction of systemd unit names vary
+        from service to service
+        """
+        service_names = []
+        if self.osd_id and self.proc_name == 'ceph-osd':
+            service_names = ["{}@{}".format(self.proc_name, self.osd_id)]
+        if self.proc_name == 'ceph-mon':
```
jan--f: Maybe merge these if branches into one?

```
if self.service_name in ['ceph-mon', 'ceph-mgr', ...]
```

Martin-Weiss:
> I'm wondering about the blacklist functionality...do we expect users to have down OSDs, that they can't disable via systemd?

How do you disable a single OSD via systemd? 
(AFAIK in the past we had to create a file somewhere that prevents the auto-start during bootup).
Also there are noup noin settings for OSDs that might keep them down?
And especially in "not healthy" cluster situations deepsea needs to work properly to help fixing / building other OSDs etc..

jschmid1:
> I'm wondering about the blacklist functionality...do we expect users to have down OSDs, that they can't disable via systemd?

Yes, apparently there are people that want to keep OSDs in and up although they are failing periodically. 

The other 'pro' that would come to my mind is manageability. Instead of ssh/salting to all nodes and systemd disable ceph-osd@id you can use a single central file..

If you however feel that this is redundant, we can also drop it.

srv/salt/_modules/cephprocesses.py
jschmid1: Well, true. This check loop-check is mainly there for nfs-ganesha, but maybe in the future we'll have services that are dependent on more than just one service aswell..

> would it maybe be better to check for rpcbind if service_name == nfs-ganesha?

nfs-ganesha _and_ rpcbind have a systemd unit respectively. Imo we need to check both.

jan--f: > Well, true. This check loop-check is mainly there for nfs-ganesha, but maybe in the future we'll have services that are dependent on more than just one service aswell..

Like what exactly?

> > would it maybe be better to check for rpcbind if service_name == nfs-ganesha?
> 
> nfs-ganesha _and_ rpcbind have a systemd unit respectively. Imo we need to check both.

Thats what I mean. Drop the loop and add the condition to additionally check rpcbind if service_name == nfs-ganesha

jan--f:
> How do you disable a single OSD via systemd?
> (AFAIK in the past we had to create a file somewhere that prevents the auto-start during bootup).

Via the ceph-osd@<osd_id> unit? The udev rules should only automount the osd partition iirc. But I'm happy to double check this.

> Also there are noup noin settings for OSDs that might keep them down?

noup and noin are ceph settings. That doesn't necessarily affect the osd processes running status.

> And especially in "not healthy" cluster situations deepsea needs to work properly to help fixing / building other OSDs etc..

That is exactly what @jschmid1 is working on but we need to make sure this happens in an orderly fashion as to not make things worse.

Martin-Weiss:
> 
> 
> > How do you disable a single OSD via systemd?
> > (AFAIK in the past we had to create a file somewhere that prevents the auto-start during bootup).
> 
> Via the ceph-osd@<osd_id> unit? The udev rules should only automount the osd partition iirc. But I'm happy to double check this.
> 

There was some problem we had in the past in case of corrupt filesystem scenarios on the ODSs that caused a server hang when booting up. And the only way to block that was to disable the udev rules completely or to create a special file somewhere in /var/lib/ceph that was blocking the specific OSD from being started. Unfortunately I can not find the details on that at the moment.

> > Also there are noup noin settings for OSDs that might keep them down?
> 
> noup and noin are ceph settings. That doesn't necessarily affect the osd processes running status.
> 

True - just in case deepsea needs to deal with these OSDs that are marked "noup or noin".

> > And especially in "not healthy" cluster situations deepsea needs to work properly to help fixing / building other OSDs etc..
> 
> That is exactly what @jschmid1 is working on but we need to make sure this happens in an orderly fashion as to not make things worse.

100% agreed :-)

jschmid1:
> My question would be why operators wouldn't disable such OSDs? A flapping OSD only causes unnecessary cluster rebalancing.

Are you @Martin-Weiss, saying that this is not always possible and blacklisting is something useful?



> My worry is that having two different ways of marking OSDs as intentionally down will serve to people shoot themselves in the foot. For adding OSDs that might be fine, but the check runs in other situations as well (stage 5 comes to mind). We might loose our (already less then perfect) safe guards. Not so much worried about redundancy but rather about operational complexity.

I'm also not 100% convinced that having two different methods for avoiding checks is the right way to go.  
When Martin says, that in the field there is absolutely such a case where systemd units can't be disabled or shouldn't be disabled, we have our answer. If not, let's remove it.

Martin-Weiss:
> 
> 
> > My question would be why operators wouldn't disable such OSDs? A flapping OSD only causes unnecessary cluster rebalancing.
> 
> Are you @Martin-Weiss, saying that this is not always possible and blacklisting is something useful?
> 

Depending on the scenario I believe blacklisting can be a useful function - yes. But maybe I did not fully understand the discussion about blacklisting, here.

I can imagine that a OSD is down and an administrator wants it to be down and stay down even on reboots.

I can also imagine that an OSD is down without the admin even knowing about it - as if the OSD goes down and ceph re-balances the cluster is fine once enough replicas are there again and some admins will just realize this when digging deeper (depends on their monitoring and reaction processes).

> > My worry is that having two different ways of marking OSDs as intentionally down will serve to people shoot themselves in the foot. For adding OSDs that might be fine, but the check runs in other situations as well (stage 5 comes to mind). We might loose our (already less then perfect) safe guards. Not so much worried about redundancy but rather about operational complexity.
> 
> I'm also not 100% convinced that having two different methods for avoiding checks is the right way to go.
> When Martin says, that in the field there is absolutely such a case where systemd units can't be disabled or shouldn't be disabled, we have our answer. If not, let's remove it.

I see usecases where some systemd services get deactivated by admins manually - i.e. disable an iSCSI gateway due to network maintenance or broken iSCSI NIC. For OSDs I am not sure but I believe disabling the ceph-osd@<nr> does not survice a reboot and the OSD will be up and running again after the reboot..

I am also not sure if I am the only one we should base such a decision on ;-).

jschmid1:
Let's do a vote here again:

Based on what I've seen so far I still consider this feature to be helpful. If we properly communicate it, I think we are fine.

+1 from my side for keeping blacklist in.

Martin-Weiss:
+1 also from my side for keeping blacklist in.

(with proper documentation for usability ;-))

srv/salt/_modules/cephprocesses.py
jschmid1: > Like what exactly?

What will the monitoring stack look like once it's a separate role? Should we check grafana & prometheus ? That's probably the only example I can think of right now.





> Thats what I mean. Drop the loop and add the condition to additionally check rpcbind if service_name == nfs-ganesha

The only 'con' to that suggestion would just be style. With the current design it's easy to add an arbitrary number of dependencies to a service whilst we have to catch (maybe eventually occuring) exceptions to the is_disabled property.

If we can't find more examples than nfs-ganesha I'm happy to change the code towards your suggestion

jschmid1:
@jan--f @tserong ?

tserong:
I *hope* that needing the blacklist is a rare case, but when it _is_ needed, it'll be easier to specify a list of OSD IDs in that file, than in will be do individually figure out which nodes the busted OSDs are on, then mask and disable the individual systemd units on those nodes, so I support keeping the blacklist.

jan--f:
I think what's rubbing me wrong about the blacklist feature is that cephprocesses checks two sources of truth...blacklist and disabled service. How about we add a little function in cephprocesses that reads the blacklist and calls stop and disable on them before checking for stopped and disabled.

This would add the convenience of the blacklist but we'd still only have the "stopped && disbaled -> intentionally down" criteria.

jschmid1:
> How about we add a little function in cephprocesses that reads the blacklist and calls stop and disable on them before checking for stopped and disabled.

I personally don't think that we should change the state of the systemd units ourselves. This might cause confusion and is also probably not what people expect to happen.

tserong:
> I personally don't think that we should change the state of the systemd units ourselves. This might cause confusion and is also probably not what people expect to happen.

To add to that, OSDs are a bit weird; the udev magic triggers ceph-disk, which in turn invokes `systemctl enable` for the relevant units, so even if we disabled them, after a reboot, they'd come right back up again. Unless we `systemctl mask`ed them. So because OSDs are already "weird" I'm OK with having the blacklist function for them here.

jan--f:
> ... So because OSDs are already "weird" I'm OK with having the blacklist function for them here.

Sigh...fair enough.

PR #1431, opened 2018-10-17
Add validation for pending updates

### LINKED ISSUE #1126: Salt updates
DeepSea cannot gracefully upgrade Salt.  This is simply a function of how orchestrations work.  When the Salt master restarts, the orchestration is terminated.  Shoehorning this into the reactor would likely cause more issues than it solves.

Salt updates are not terribly frequent and DeepSea is completely dependent on a working Salt cluster.  I believe Stage 0 needs to fail with instructions on installing and upgrading Salt.  Some users may be using Salt for other purposes and bringing these changes to their attention should prevent any confusion.  Also, updating and not restarting Salt daemons will likely end in heartache and confusion when a minion is upgraded prior to the master restarting.

jschmid1:
* add to setup(stage.0) validation
* notifies if there is a pending update
  with 'ceph' in its name.
* stops stage.0 when there is salt-master or salt-minion
  package in the queue and gives advise on how to proceed
* adds test

Fixes #1126

The checks add 3~seconds to the execution time respectively.

open questions:

1) Should the 'ceph-package' check also be included in stage.3 (adds ~3seconds)

2) I was _trying_ to use `mine.get` in the `validate.py` calls to reduce the invocation time
    for those checks. I had to find out that the mines lie quite often.. This might be something we want
    to look into a bit closer as we also rely on mines in the `replace.py` (where we do a complete `mine.flush` with a successive `mine.update` .. that's not ideal at best, imo.

3) How to handle the `salt has updates` for the CI? Do we force to update the base image? Should be build in a switch do disable this? (CI_MODE, ( we had this discussion in the past about another feature))

-----------------

**Checklist:**
~- [ ] Adapted documentation~

### DIFF
--- a/srv/modules/runners/populate.py
+++ b/srv/modules/runners/populate.py
@@ -73,24 +73,6 @@ def _cmp(x, y):
 
     return (x > y) - (x < y)
 
-class Settings(object):
-    """
-    Common settings
-    """
-
-    def __init__(self):
-        """
-        Assign root_dir, salt __opts__ and stack configuration.  (Stack
-        configuration is not used currently.)
-        """
-        __opts__ = salt.config.client_config('/etc/salt/master')
-        self.__opts__ = __opts__
-
-        for ext in __opts__['ext_pillar']:
-            if 'stack' in ext:
-                self.stack = ext['stack']
-        self.root_dir = "/srv/pillar/ceph/proposals"
-
 
 class SaltWriter(object):
     """
@@ -847,7 +829,7 @@ def show(**kwargs):
     print ("DEPRECATION WARNING: the generation of storage profiles is now"
            " handled by the proposal runner. This function will go away in the"
            " future.")
-    settings = Settings()
+    settings = __utils__['settings.self_']()
 
     salt_writer = SaltWriter(**kwargs)
 
@@ -894,7 +876,7 @@ def proposals(**kwargs):
     Collect the hardware profiles, all possible role assignments and common
     configuration under /srv/pillar/ceph/proposals
     """
-    settings = Settings()
+    settings = __utils__['settings.self_']()
 
     salt_writer = SaltWriter(**kwargs)
 
@@ -1077,9 +1059,9 @@ def engulf_existing_cluster(**kwargs):
 
     search = __utils__['deepsea_minions.show']()
     local = salt.client.LocalClient()
-    settings = Settings()
-    salt_writer = SaltWriter(overwrite=True)
     exception = kwargs.get('exception', False)
+    settings = __utils__['settings.self_']()
+    salt_writer = SaltWriter(overwrite=True)
 
     # Make sure deepsea_minions contains valid minions before proceeding with engulf.
     from . import validate
--- a/srv/salt/_modules/packagemanager.py
+++ b/srv/salt/_modules/packagemanager.py
@@ -177,6 +177,7 @@ class Zypper(PackageManager):
     def _refresh(self):
         """
         Refresh Zypper before updating
+        Raise RefreshFailed when returncode != 1
         """
         log.info("Refreshing Repositories..")
 
@@ -189,10 +190,10 @@ class Zypper(PackageManager):
         proc = Popen(cmd, stdout=PIPE, stderr=PIPE)
         proc.wait()
         if proc.returncode != 0:
-            log.error('Refreshing failed. Check the repos')
-            log.debug('Executing {}'.format(cmd))
+            log.error('Refreshing failed. There might be an issue resolving the repos')
+            log.debug('Executed {}'.format(cmd))
             return False
-        return None
+        return True
 
     # pylint: disable=no-self-use
     def _upgrades_needed(self):
@@ -233,16 +234,21 @@ class Zypper(PackageManager):
         """
         List all pending updates (transformed from XML)
         """
-        self._refresh()
+        ret = {'status': True, 'packages': []}
+        ret_refresh = self._refresh()
+        if ret_refresh is False:
+            log.error("Error while refreshing the repos.")
+            ret['status'] = ret_refresh
         cmd = "zypper -x lu"
         log.debug('Executing {}'.format(cmd))
         proc = Popen(cmd, stdout=PIPE, stderr=PIPE, shell=True)
         stdout, _ = proc.communicate()
         stdout = __salt__['helper.convert_out'](stdout)
-        update_list = self._parse_xml(stdout)
+        ret['packages'] = self._parse_xml(stdout)
         if _filter:
-            return [x for x in update_list if any(w in x['name'] for w in _filter)]
-        return update_list
+            ret['packages'] = [x for x in ret['packages'] if any(w in x['name'] for w in _filter)]
+            return ret
+        return ret
 
     def _patches_needed(self):
         """
--- a/srv/salt/_modules/packagemanager.py
+++ b/srv/salt/_modules/packagemanager.py
@@ -177,7 +177,6 @@ class Zypper(PackageManager):
     def _refresh(self):
         """
         Refresh Zypper before updating
-        Raise RefreshFailed when returncode != 1
         """
         log.info("Refreshing Repositories..")
 
--- a/srv/salt/_modules/packagemanager.py
+++ b/srv/salt/_modules/packagemanager.py
@@ -5,6 +5,7 @@ Intervention into special conditions during package management
 """
 from __future__ import absolute_import
 from subprocess import Popen, PIPE
+import xml.etree.ElementTree as ET
 import logging
 import os
 # pylint: disable=import-error,3rd-party-module-not-gated,redefined-builtin
@@ -217,6 +218,32 @@ class Zypper(PackageManager):
         log.info('No Update Needed')
         return False
 
+    def _parse_xml(self, xml, find_str='.//update'):
+        """
+        utils method to parse xml from a str
+        """
+        root = ET.fromstring(xml)
+        ret = list()
+        for child in root.findall(find_str):
+            ret.append(child.attrib)
+        return ret
+
+    # pylint: disable=dangerous-default-value
+    def list_updates(self, _filter=[]):
+        """
+        List all pending updates (transformed from XML)
+        """
+        self._refresh()
+        cmd = "zypper -x lu"
+        log.debug('Executing {}'.format(cmd))
+        proc = Popen(cmd, stdout=PIPE, stderr=PIPE, shell=True)
+        stdout, _ = proc.communicate()
+        stdout = __salt__['helper.convert_out'](stdout)
+        update_list = self._parse_xml(stdout)
+        if _filter:
+            return [x for x in update_list if any(w in x['name'] for w in _filter)]
+        return update_list
+
     def _patches_needed(self):
         """
         Updates that are sourced from an official Update
@@ -361,3 +388,38 @@ def migrate(**kwargs):
     pm = Zypper(**kwargs)
     # pylint: disable=protected-access
     pm._migrate()
+
+
+def updates_needed(**kwargs):
+    """
+    Are updates needed?
+    """
+    obj = PackageManager(**kwargs)
+    # pylint: disable=protected-access
+    return obj.pm._updates_needed()
+
+
+def list_all_updates(**kwargs):
+    """
+    List updates
+    """
+    obj = PackageManager(**kwargs)
+    return obj.pm.list_updates()
+
+
+def list_ceph_updates(**kwargs):
+    """
+    Convenience public method
+    List updates (only ceph)
+    """
+    obj = PackageManager(**kwargs)
+    return obj.pm.list_updates(_filter=['ceph'])
+
+
+def list_salt_updates(**kwargs):
+    """
+    Convenience public method
+    List updates (only salt)
+    """
+    obj = PackageManager(**kwargs)
+    return obj.pm.list_updates(_filter=['salt-minion', 'salt-master'])
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -987,7 +987,13 @@ class Validate(Preparation):
         updates = self.local.cmd(self.matches,
                                  'packagemanager.list_salt_updates',
                                  tgt_type='list')
-        updates = [item for sublist in list(updates.values()) for item in sublist]
+        updates_list = list(updates.values())
+        status = [x['status'] for x in updates_list]
+        packages = [x['packages'] for x in updates_list]
+        if False in status:
+            self.warnings['refresh_repos'] = ["Experienced trouble refreshing repositories."]
+        # flatten the packages list to avoid iterating over all minion['packages']
+        updates = [item for sublist in packages for item in sublist]
         if not updates:
             self.passed['salt_updates'] = "valid"
         else:
@@ -1009,7 +1015,14 @@ class Validate(Preparation):
         updates = self.local.cmd(self.matches,
                                  'packagemanager.list_ceph_updates',
                                  tgt_type='list')
-        updates = [item for sublist in list(updates.values()) for item in sublist]
+
+        updates_list = list(updates.values())
+        status = [x['status'] for x in updates_list]
+        packages = [x['packages'] for x in updates_list]
+        if False in status:
+            self.warnings['refresh_repos'] = ["Experienced trouble refreshing the repositories."]
+        # flatten the packages list to avoid iterating over all minion['packages']
+        updates = [item for sublist in packages for item in sublist]
         if not updates:
             self.passed['ceph_updates'] = "valid"
         else:
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -1020,14 +1020,15 @@ class Validate(Preparation):
         status = [x['status'] for x in updates_list]
         packages = [x['packages'] for x in updates_list]
         if False in status:
-            self.warnings['refresh_repos'] = ["Experienced trouble refreshing the repositories."]
+            # pylint: disable=line-too-long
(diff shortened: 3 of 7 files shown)


jschmid1:
@susebot run teuthology

jschmid1:
[Smoothen stage.0 when a salt update is pending](https://trello.com/c/B7th2z6z/94-smoothen-stage0-when-a-salt-update-is-pending)

kshtsk:
@jschmid1 @smithfarm I suppose this failure introduced after recent sle-15.0 image updates.

kshtsk:
@smithfarm Don't know, may be the source SLE repos are bad?

smithfarm:
> How to handle the salt has updates for the CI? Do we force to update the base image?

There are no salt RPMs installed in the base image anymore, and teuthology itself does not use salt anymore. For SES6 I have a PR pending - https://github.com/SUSE/ceph/pull/209 - which introduces a new "salt" task which installs the salt RPMs and deploys the salt cluster.

> Should be build in a switch do disable this? (CI_MODE, ( we had this discussion in the past about another feature))

Not needed for the CI. The CI will always have some set of repos, and will always have installed the latest salt available in *those* repos.

smithfarm:
@susebot run teuthology

smithfarm:
> Don't know, may be the source SLE repos are bad?

The originals? I doubt it. I'm using SLE-15-Module-Basesystem-POOL-x86_64-Build668.1 on ECP and it works fine there.

kshtsk:
```  basesystem:
    src: /mnt/dist/ibs/SUSE:/SLE-15:/GA:/TEST/images/repo/SLE-15-Module-Basesystem-POOL-x86_64-Media1
    url: http://158.69.90.90/artifacts/ci/e9bf06179ab8222426cd8b22e7e5d63f40d2b12f67d6a47a4898983f5735ccdc/SLE-15-Module-Basesystem-POOL-x86_64-Build668.1
```
It is possible IBS changed the source directory once it has been copying.
```
du -sb /mnt/dist/ibs/SUSE:/SLE-15:/GA:/TEST/images/repo/SLE-15-Module-Basesystem-POOL-x86_64-Media1
2186551068	/mnt/dist/ibs/SUSE:/SLE-15:/GA:/TEST/images/repo/SLE-15-Module-Basesystem-POOL-x86_64-Media1
```
```
du -sb ci/e9bf06179ab8222426cd8b22e7e5d63f40d2b12f67d6a47a4898983f5735ccdc/SLE-15-Module-Basesystem-POOL-x86_64-Build668.1
2186374905	ci/e9bf06179ab8222426cd8b22e7e5d63f40d2b12f67d6a47a4898983f5735ccdc/SLE-15-Module-Basesystem-POOL-x86_64-Build668.1
```

kshtsk:
@susebot run teuthology

smithfarm:
@kshtsk Which teuthology branch/SHA1 is Jenkins using? The install task is not working properly.

kshtsk:
```(virtualenv) ubuntu@teuth-ci:~/teuthology$ git log --oneline | head -10
5d6bcb9 Merge pull request #150 from SUSE/wip-test-repo-priority
e3d6e70 Merge pull request #154 from SUSE/wip-tweak-install-task
25f01dc Add optional priority for --test-repo
8899c90 task/install: rename install_packages param to install_ceph_packages
7f5aec1 Merge pull request #149 from SUSE/wip-fix-job-owner
3f852a8 Merge pull request #148 from SUSE/wip-flat2gen
30de8bb Fix worker can't figure out owner of a job
d0dcc4a task/install: allow ceph-test to be installed with vendor change
ed393ad task/install: allow specification of multiple zypper options
6bb0cff Merge pull request #140 from SUSE/wip-extra-packages-backport
```

smithfarm:
@kshtsk Yes, my fault. Fixing.

smithfarm:
@susebot run teuthology

kshtsk:
@smithfarm Ok, I'll rollback the teuth-ci to the sha1 before your change.

kshtsk:
```2018-10-17T21:37:04.347 INFO:teuthology.orchestra.run.target158069065225:Running: 'sudo rpm -q ceph-test'
2018-10-17T21:37:04.453 INFO:teuthology.orchestra.run.target158069065225.stdout:package ceph-test is not installed
2018-10-17T21:37:04.457 ERROR:teuthology.run_tasks:Saw exception from tasks.
  File "/home/ubuntu/teuthology/teuthology/run_tasks.py", line 95, in run_tasks
    run_one_task(taskname, stack, timer, ctx=ctx, config=config)
  File "/home/ubuntu/teuthology/teuthology/run_tasks.py", line 74, in run_one_task
    manager.__enter__()
  File "/home/ubuntu/teuthology/teuthology/task/__init__.py", line 121, in __enter__
    self.begin()
  File "/home/ubuntu/src/github.com_SUSE_ceph_ses6/qa/tasks/deepsea.py", line 187, in begin
    'ceph-test'
  File "/home/ubuntu/teuthology/teuthology/orchestra/remote.py", line 193, in run
    r = self._runner(client=self.ssh, name=self.shortname, **kwargs)
  File "/home/ubuntu/teuthology/teuthology/orchestra/run.py", line 403, in run
    r.wait()
  File "/home/ubuntu/teuthology/teuthology/orchestra/run.py", line 166, in wait
    label=self.label)
CommandFailedError: Command failed on target158069065225 with status 1: 'sudo rpm -q ceph-test'
```
Didn't help

smithfarm:
> Ok, I'll rollback the teuth-ci to the sha1 before your change.

@kshtsk Not necessary. I already fixed the yaml in both ses5 and ses6 branches. The latest test run got to Stage 0, where it failed - the failure appears to be related to this PR.

smithfarm:
@jschmid1 Stage 0 fails - log excerpt in https://github.com/SUSE/DeepSea/pull/1431#issuecomment-430797334

srv/modules/runners/populate.py
```diff
@@ -1077,9 +1059,9 @@ def engulf_existing_cluster(**kwargs):
 
     search = __utils__['deepsea_minions.show']()
     local = salt.client.LocalClient()
-    settings = Settings()
-    salt_writer = SaltWriter(overwrite=True)
     exception = kwargs.get('exception', False)
+    settings = __utils__['settings.self_']()
+    salt_writer = SaltWriter(**kwargs)
```
tserong: Should be `salt_writer = SaltWriter(overwrite=True)` (see 51a06fa9)

srv/salt/_modules/packagemanager.py
```diff
@@ -217,6 +218,32 @@ def _updates_needed(self):
...
+        for child in root.findall(find_str):
+            ret.append(child.attrib)
+        return ret
+
+    # pylint: disable=dangerous-default-value
+    def list_updates(self, _filter=[]):
+        """
+        List all pending updates (transformed from XML)
+        """
+        self._refresh()
```
tserong: `zypper ref` here on my vagrant test setup (with all local repos for SLE and SES Pool and Updates) takes 1.171 seconds. On my desktop system though, with several remote repos (opensuse.org, etc.), `zypper ref` takes 16.4 seconds, which would make this rather more unpleasant. I'm not sure there's really any way around this though (the repos need to be refreshed to know if there's actually updates; if we skipped the explicit refresh, it would only work properly if the repos were configured to auto-refresh, and they might not be).

srv/modules/runners/populate.py
jschmid1: ack

srv/salt/_modules/packagemanager.py
jschmid1: I tested this on my vagrant setup with 'local' (in-house but non-host-local) repos and the whole process including the refresh took 3 seconds.

I notices that zypper returns something like:

`Repository 'repo_name' is out-of-date. You can run 'zypper refresh' as root to update it.`

That's something we could look for, which would in turn require to run zypper lu twice..
I think for people running their DCs with a SMT+local mirror it's gonna be fine, but people that have mirrors which are a bit further away it will be a considerable time waste..

I do think that it's worse to miss and update at the hence miss the warning and then actually apply the update in a later stage. `_updates_needed` also invokes `_refresh` which means that when we leave out the `_refresh` in `list_updates`, we actually have a change to miss to emit a warning

jschmid1:
I also stepped back from registering the function as a mine since I don't use the mine anymore. See comments

jschmid1:
@susebot run teuthology

jschmid1:
unrelated make-lint issue: (sometimes occurs, sometimes not..)

```
10:45:54 ************* Module srv.salt._modules.rgw
10:45:54 E: 22, 0: No name 's3' in module 'boto' (no-name-in-module)
10:45:54 E: 24, 0: No name 'exception' in module 'boto' (no-name-in-module)
```

srv/salt/_modules/packagemanager.py
jan--f: I was wondering the same. Though I'm not a fan of preventing a stage for this...seems confusing. Ideally we would print the validation messages again after everything run to make it visible, but that seems like a lot of work.

jschmid1: Mhh, I don't know if we have many other options.. Either we hazard the consequences of having a disconnect between what we show in the validation and actually do in the stage or spend more time on (unnecessary)refreshing of repos.. Tough call..

jschmid1:
well.. we need to make a decision:

I vote for keeping the `_refresh` call.

Reason: Rather wait a couple of seconds more than having a discrepancy between validations and actions.

tserong:
Let's keep the `_refresh` call, it's the safest thing to do, and usually (with a sane repo setup, reasonably local mirrors, etc.) won't be expensive. I think we can look at optimizing this as a separate issue if necessary.

Martin-Weiss:
> 
> 
> Let's keep the `_refresh` call, it's the safest thing to do, and usually (with a sane repo setup, reasonably local mirrors, etc.) won't be expensive. I think we can look at optimizing this as a separate issue if necessary.

While personally I also prefer to do a manual "zypper ref" always before doing any upgrade or additional rpm installation - we really should recommend to configure the zypper repos with "auto refresh" (at least for the update repositories - pool / dvd repos normally do not change and might not require a refresh).
Maybe our _refresh call could check this setting on configured repos and skip the refresh in case they are set for auto refresh?

jschmid1:
> if the zypper refresh fails? I'm wondering if we might inhibit people from doing things if, say their SCC is down or some such thing. Imho it would be good to fail gracefully here...no zypper connection? -> warn on it but continue.

Oh, I've missed this. Well we can probably can check for the returncode and judge based on this. Let me add that.

tserong:
> Maybe our _refresh call could check this setting on configured repos and skip the refresh in case they are set for auto refresh?

Good idea.

jschmid1:
```
admin:/deepsea # salt-run validate.setup
deepsea_minions          : valid
master_minion            : valid
salt_version             : valid
salt_updates             : valid
refreh_repos             : ['Experienced trouble refreshing the repositories.']
ceph_updates             : ['On or more of your minions have updates pending that might cause ceph-daemons to restart. This might extend the duration of this Stage depending on your cluster size.']                                                                                                                                                                       
True
admin:/deepsea # zypper lr
Repository priorities are without effect. All enabled repositories share the same priority.

# | Alias             | Name              | Enabled | GPG Check | Refresh
--+-------------------+-------------------+---------+-----------+--------
1 | Storage           | Storage           | Yes     | (r ) Yes  | Yes
2 | basesystyem       | basesystyem       | Yes     | (r ) Yes  | Yes
3 | devtools          | devtools          | Yes     | (r ) Yes  | Yes
4 | sles-applications | sles-applications | Yes     | (r ) Yes  | Yes
5 | test_bad          | test_bad          | Yes     | ( p) Yes  | No
admin:/deepsea # zypper mr -d 5
Repository 'test_bad' has been successfully disabled.
admin:/deepsea # salt-run validate.setup
deepsea_minions          : valid
master_minion            : valid
salt_version             : valid
salt_updates             : valid
ceph_updates             : ['On or more of your minions have updates pending that might cause ceph-daemons to restart. This might extend the duration of this Stage depending on your cluster size.']                                                                                                                                                                       
True
admin:/deepsea # 
```

Looks like this now.

smithfarm:
@jschmid1 Looks awesome. Just one question:

```
refreh_repos             : ['Experienced trouble refreshing the repositories.']
```

Does it really say "refreh_repos"?

Martin-Weiss:
I really like what I see here! ;-)
Just some minor comments:

> refreh_repos             : ['Experienced trouble refreshing the repositories.']

Could we give the admin some more details about "which repo on which server had a problem or where can we find more details"? ((i,e. check /var/log.... for more details)

> ceph_updates             : ['On or more of your minions have updates pending that might cause ceph-daemons to restart. This might extend the duration of this Stage depending on your cluster size.']                                                                                                                                                                       

--> is it possible to present more details on this? The first question I would have as admin is "which minions?"

jschmid1:
> Does it really say "refreh_repos"?

Yeah, it _did_ :)

jschmid1:
> Could we give the admin some more details about "which repo on which server had a problem or where can we find more details"? ((i,e. check /var/log.... for more details)

The issue with this is lucidy.. When a user has let's say 100 nodes and 50 of them have issues, you'll be flooded with a wall of text.

We definitely can implement this, but then we have to adapt the PrettyPrinter to handle such cases better.

> is it possible to present more details on this? The first question I would have as admin is "which minions?"

We have this information, but I feel we have the same issue as mentioned above. It's hard to estimate how big the wall of text will be with large scale clusters. 

The current message just gives you a heads-up that it might take longer, which is a good _start_ imo

jschmid1:
> While personally I also prefer to do a manual "zypper ref" always before doing any upgrade or additional rpm installation - we really should recommend to configure the zypper repos with "auto refresh" (at least for the update repositories - pool / dvd repos normally do not change and might not require a refresh).
> Maybe our _refresh call could check this setting on configured repos and skip the refresh in case they are set for auto refresh?

I fear that auto_refresh does not get triggered by calling `zypper lu`. I specifically tested this case on my VMs and got different results after refreshing manually.

Martin-Weiss:
> 
> 
> > Could we give the admin some more details about "which repo on which server had a problem or where can we find more details"? ((i,e. check /var/log.... for more details)
> 
> The issue with this is lucidy.. When a user has let's say 100 nodes and 50 of them have issues, you'll be flooded with a wall of text.
> 
> We definitely can implement this, but then we have to adapt the PrettyPrinter to handle such cases better.
> 

A minimum for each error message should be a hint where to find more details on it.
I that is printed on the screen or going to a log is something I do not care in the end as long as the admin can find it easily based on the hint given by the command executed ;-).
So yes - if there are 50 nodes having a problem I would also be fine with a long list of nodes with the problem or the info in which log I can find that.

> > is it possible to present more details on this? The first question I would have as admin is "which minions?"
> 
> We have this information, but I feel we have the same issue as mentioned above. It's hard to estimate how big the wall of text will be with large scale clusters.
> 
> The current message just gives you a heads-up that it might take longer, which is a good _start_ imo

Sure - and I like it - just had the assumption that we write warnings and errors already to some log where and admin can look up more details and the info about the "where to look next" is what would be great on such errors or warnings ;-).

Martin-Weiss:
> 
> 
> > While personally I also prefer to do a manual "zypper ref" always before doing any upgrade or additional rpm installation - we really should recommend to configure the zypper repos with "auto refresh" (at least for the update repositories - pool / dvd repos normally do not change and might not require a refresh).
> > Maybe our _refresh call could check this setting on configured repos and skip the refresh in case they are set for auto refresh?
> 
> I fear that auto_refresh does not get triggered by calling `zypper lu`. I specifically tested this case on my VMs and got different results after refreshing manually.

I do not know about the logic behind the autorefresh - but I found this "If autorefresh is enabled for a repository, you don't need to worry about refreshing, it happens automatically when needed." (https://en.opensuse.org/SDB:Zypper_usage). So zypper must have some "magic" included  ;-).
Might be that we have a bug or the way how you test with getting different results needs to be analyzed deeper? (Guessing: I would expect that there is some sort of timeout of autorefresh before it gets triggered again?)

jschmid1:
> Sure - and I like it - just had the assumption that we write warnings and errors already to some log where and admin can look up more details and the info about the "where to look next" is what would be great on such errors or warnings ;-).

we can make that happen :+1:

jschmid1:
> I do not know about the logic behind the autorefresh - but I found this "If autorefresh is enabled for a repository, you don't need to worry about refreshing, it happens automatically when needed." (https://en.opensuse.org/SDB:Zypper_usage). So zypper must have some "magic" included ;-).
> Might be that we have a bug or the way how you test with getting different results needs to be analyzed deeper? (Guessing: I would expect that there is some sort of timeout of autorefresh before it gets triggered again?)

maybe this only happens on a 'update/dist-up'.. we should eventually look into the zypper codebase or consult someone with more insight to the zypper internals

jschmid1:
Added repo checks with https://github.com/SUSE/DeepSea/pull/1431/commits/9524cf31a34cf11d7ec426246169ce2462c9e344

jschmid1:
is everyone happy with the last two commits?

srv/salt/_modules/packagemanager.py
```diff
@@ -177,6 +177,7 @@ def __init__(self, **kwargs):
     def _refresh(self):
         """
         Refresh Zypper before updating
+        Raise RefreshFailed when returncode != 1
```
smithfarm: Did you mean when returncode != 0 ?

jschmid1: I don't even'raise' anymore; deleting the comment

jschmid1:
dump

tserong:
LGTM

PR #1441, opened 2018-10-31
Makefile: Check for correct ID in /etc/os-release

marcosps:
Without this patch, is not possible to install DeepSea using openSuse
Leap 15.0, since the /etc/os-release ID field is "opensuse-leap" [1],
so the Makefile was checking for "opensuse", "opensuse-tumbleweed" and
"sles". Now zypper is able to install sal-api and other necessary
packages since Leap is detected and the PKG_INSTALL variable is
populated.

[1]: https://en.opensuse.org/SDB:Find_openSUSE_version

Signed-off-by: Marcos Paulo de Souza <marcos.souza.org@gmail.com>

Fixes #

Description:

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ PYTHON_DEPS=python3-setuptools python3-click python3-tox
 PYTHON=python3
 
 OS=$(shell source /etc/os-release 2>/dev/null ; echo $$ID)
-ifeq ($(OS), opensuse)
+ifeq ($(OS), opensuse-leap)
 USER=salt
 GROUP=salt
 PKG_INSTALL=zypper -n install


marcosps:
Let me know if you guys prefer a new entry for "opensuse-leap" instead of changing the "opensuse" one, as I don't know if this is done in order to keep compatibility to older openSuse versions.

smithfarm:
Compatibility with pre-Leap opensuse versions is not needed. Earlier (non-updated) Leap 42.x had "opensuse" instead of "opensuse-leap", but these would not be used with master.

It would be good to somehow get rid of the code duplication (actually the same code is inside _three_ consecutive conditionals, so it's code _triplication_), but that can be left for a subsequent PR.

PR #1450, opened 2018-11-06
Makefile: clean up suse conditional blocks

smithfarm:
Follow-up to #1441

---

- [x] tested on openSUSE 15.0
- [x] tested on SLE-15

PR #1454, opened 2018-11-07
Makefile: Fix two missing $DESTDIRs

tserong:
Found while verifying https://github.com/SUSE/DeepSea/pull/1450.
That these $(DESTDIR) pieces were missing wouldn't actually cause
a problem for normal RPM builds (the spec file doesn't invoke
`make install` anyway, so won't hit these lines); it's only a
problem for weirdos like me ;-)


smithfarm:
I'll test this today.

smithfarm:
@tserong It's not just RPM builds that we need to consider. There are also folks, like me, who frequently install DS from source.

smithfarm:
@susebot run teuthology

tserong:
@smithfarm, oh, absolutely. I install from source regularly too (I just don't usually specify DESTDIR, hence not noticing these particular omissions earlier)

PR #1455, opened 2018-11-07
Add mgr_orch runner to support ceph-mgr deepsea orchestrator module

tserong:
This runner is intended to provide a 1:1 mapping between the ceph-mgr
orchestrator API, and DeepSea internals.  This enables us to keep the
ceph-mgr orhcestrator module quite simple, and do any necssary data
aggregation here inside DeepSea itself.  This commit introduces only
one method so far -- inventory -- for returning a list of nodes and
the devices on those nodes.

I've also added a "blank" field in cephdisks.py to indicate whether
a disk is empty or not (the orchestrator API wants to know this).

**Checklist:**
- [ ] Adapted documentation


tserong:
Marked DNM, because I want to at least add something in support of the orchestartor's describe_service method (see http://docs.ceph.com/docs/master/mgr/orchestrator_modules/). We'll ultimately need to provide functions to handle most (but possibly not all) of the functions in the orchestrator API (https://github.com/ceph/ceph/blob/master/src/pybind/mgr/orchestrator.py#L149-L279)

jschmid1:
let me know when it's ready for a review

tserong:
This won't work for LVM devices, because the `cephdisks` module doesn't understand LVM deployments yet -- see https://github.com/SUSE/DeepSea/blob/master/srv/salt/_modules/cephdisks.py#L347-L376

tserong:
@jan--f, do you know offhand what we'll need to do to the `HardwareDetections._osd()` function to make it detect LVM OSDs?

tserong:
Removed "DNM", as this now provides the very basic read-only/inventory functions needed by the orchestartor. Feel free to review :-)

srv/modules/runners/mgr_orch.py
```diff
@@ -0,0 +1,172 @@
...
+log = logging.getLogger(__name__)
+
+
+def _sanitize_devices(devices):
+    """
+    Takes a list of devices and returns only the information necessary for
+    the ceph-mgr orchestrator's InventoryDevice class.
+    """
+
+    # Shamelessly stolen from populate.py, to try to prefer /dev/disk/by-id
```
jschmid1: If this is a 1-1 copy from another module we should think about making it public and importable

srv/modules/runners/mgr_orch.py
```diff
@@ -0,0 +1,172 @@
...
+        device = drive['Device File']
+        if 'Device Files' in drive:
+            for path in drive['Device Files'].split(', '):
+                if 'by-id' in path:
+                    device = path
+                    break
+        return device
+
+    # Attempt to figure out if the drive is hdd, ssd or nvme
+    # TODO: this looks right, but is completely untested for ssd and nvme!
```
jschmid1: Tests would help there :)

srv/modules/runners/mgr_orch.py
```diff
@@ -0,0 +1,172 @@
...
+    def _type(drive):
+        type = "hdd"
```
jschmid1: `type` is a reserved word in python

srv/modules/runners/mgr_orch.py
```diff
@@ -0,0 +1,172 @@
...
+
+    sanitized = []
+    for drive in devices:
+        sanitized.append({
+            "blank": drive["blank"],
+            "type": _type(drive),
+            "id": _device(drive),
+            "size": int(drive["Bytes"]),
+            "extended": None,
+            "metadata_space_free": None     # TODO: calculate this
```
jschmid1: a pending TODO, the linter will probably nag here.

srv/modules/runners/mgr_orch.py
```diff
@@ -0,0 +1,172 @@
...
+        criteria.append("I@roles:{}".format(role))
+    if criteria:
+        search = "{} and ( {} )".format(search, " or ".join(criteria))
+
+    # When search matches no minions, salt prints to stdout.  Suppress stdout.
+    _stdout = sys.stdout
+    sys.stdout = open(os.devnull, 'w')
+
+    local = salt.client.LocalClient()
+    _minions = local.cmd(search, 'cephdisks.list', [], tgt_type="compound")
```
jschmid1: This can be replaced with the ceph-volume inventory function soon

srv/modules/runners/mgr_orch.py
```diff
@@ -0,0 +1,172 @@
...
+    # This is very simple so far; DeepSea always uses short hostnames for
+    # daemon IDs, so the mapping is trivial.
+    daemon_ids = local.cmd(search, 'grains.item', ['host'], tgt_type="compound")
+
+    minion_roles = local.cmd(search, 'pillar.get', ['roles'], tgt_type="compound")
+
+    for minion, roles in minion_roles.items():
+        minion_result = {}
+        for minion_role in roles:
+            if minion_role not in ['mon', 'mgr', 'mds', 'rgw']:
```
jschmid1: let's pull that into a `supported_roles` variable

srv/modules/runners/mgr_orch.py
```diff
@@ -0,0 +1,172 @@
...
+
+    local = salt.client.LocalClient()
+    _minions = local.cmd(search, 'cephdisks.list', [], tgt_type="compound")
+
+    sys.stdout = _stdout
+
+    return {minion: _sanitize_devices(devices) for minion, devices in _minions.items()}
+
+
+def describe_service(role=None, id=None, node=None):
```
jschmid1: id is a reserved word in python

srv/modules/runners/mgr_orch.py
```diff
@@ -0,0 +1,172 @@
...
+    for minion, roles in minion_roles.items():
+        minion_result = {}
+        for minion_role in roles:
+            if minion_role not in ['mon', 'mgr', 'mds', 'rgw']:
+                # skips unsupported roles
+                continue
+            if role and minion_role != role:
+                # skips roles we don't care about if a role was passed in
+                continue
+            if id and daemon_ids[minion]['host'] != id:
```
jschmid1: `id` is reserved

srv/modules/runners/mgr_orch.py
tserong: I wondered about that - is there going to be a new runner that wraps ceph-volume, or are we just going to change cephdisks.list to use ceph-volume internally, given cephdisks.list is already used in various other places?

srv/modules/runners/mgr_orch.py
tserong: I agree in principle, but rather than making this function, and the one from srv/modules/runners/populate.py, a public importable thing, I'd rather figure out how to take the one from srv/salt/_modules/cephdisks.py:device_()  and use that in all _three_ places. Except this might all change with ceph-volume, so I'd kinda like to leave it for later :-)

tserong:
Rebased on master, and now with unit tests with 100% coverage :-)

@jschmid1, if you're happy with this I'll squash it all back to one commit.

jschmid1:
Nevertheless, let's wait for our Jenkins to come back up before merging this

tserong:
Thanks, squashed to one commit.

srv/modules/runners/mgr_orch.py
jan--f: > are we just going to change cephdisks.list to use ceph-volume internally, given cephdisks.list is already used in various other places?

This probably makes the most sense.

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
@susebot run teuthology

tserong:
Looks like Jenkins is back :-)

Issue #1008, opened 2018-03-14
More validations pre upgrade


jschmid1:
### Description of Issue/Question

More validations pre upgrade

1) Are all config files where we expect them to be
2) Is the pillar populated with the information we need. (available-roles etc.)
3) Same for grains
4) Make aware of changes that take place in the new release. (conf in mons)
5) Backup conf files.

jschmid1:
https://github.com/SUSE/DeepSea/tree/wip-upgrade-checks2

jschmid1:
https://github.com/SUSE/DeepSea/tree/wip-upgrade-checks-custom-sls

jschmid1:
https://github.com/SUSE/DeepSea/tree/wip-notify-restarts

jschmid1:
ref #1272

PR #1460, opened 2018-11-09
Check configs for deprecated k:v pairs and report

jschmid1:
This is part of the pre-upgrade validations.
 
Looks like this when executed.

```shell
admin:/deepsea # salt-run validate.config_check cluster=ceph
/srv/salt/ceph/configuration/files/ceph.conf.d/mon.conf::mock tick internal: Key 'mock tick internal' is deprecated since bobtail. Please remove it from your config
False
```

It's not yet 'activated'  in the sense of being part of one of the existing validations (pillar, setup etc..)

Addresses #1008 

The [file](https://github.com/SUSE/DeepSea/pull/1460/files#diff-109c03c6ddf40b08362f7fca3a1651b1) that maintains the map of deprecated k-v is taken from the official ceph documentation. If anyone has a more accurate source, please let me know

-----------------

**Checklist:**
- [x] Adapted documentation

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -32,6 +32,7 @@ import salt.client
 import salt.utils
 import salt.utils.minions
 import salt.utils.error
+from configobj import ConfigObj
 # pylint: disable=relative-import
 
 
@@ -1043,8 +1044,36 @@ class Validate(Preparation):
             self.passed['ceph_updates'] = "valid"
         else:
             # pylint: disable=line-too-long
-            msg = ("On or more of your minions have updates pending that might cause ceph-daemons to restart. This might extend the duration of this Stage depending on your cluster size. If you want to find out which packages will be updated, you can get the full list with salt -I 'cluster:ceph' packagemanager.list_ceph_updates")
+            msg = ("On or more of your minions have updates pending "
+                   "that might cause ceph-daemons to restart. "
+                   "This might extend the duration of this "
+                   "Stage depending on your cluster size. "
+                   "If you want to find out which packages will be "
+                   "updated, you can get the full list with "
+                   "salt -I 'cluster:ceph' packagemanager.list_ceph_updates")
             self.warnings['ceph_updates'] = [msg]
+
+    def config_check(self):
+        """
+        Verify if config does not contain any deprecated config k:v pairs
+        """
+        issue_map = ConfigCheck().run()
+        for conf_obj in issue_map:
+            key = conf_obj.key
+            values = conf_obj.values
+            filename = conf_obj.filename
+            release = conf_obj.release
+            if not values:
+                msg = ("Key '{}' is deprecated since {release}. "
+                       "Please remove it from "
+                       "your config".format(key, release=release))
+            else:
+                values = '/'.join(values)
+                msg = ("Key '{}' with value(s) {} was "
+                       "found (deprecated since {}). "
+                       "Please adapt your config".format(key, values, release))
+
+            self.errors["{}::{}".format(filename, key)] = msg
 
     def report(self):
         """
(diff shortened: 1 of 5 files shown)


smithfarm:
@kshtsk The PR CI (artifacts) is still on Mimic. In other words, when I peek into the test result referenced in the previous comment, I see ceph version 13.2.2 is installed. The CI tests that run on ECP are already on Nautilus and I *thought* the PR CI was on Nautilus before.

srv/modules/runners/validate.py
```diff
@@ -1046,6 +1046,26 @@ def ceph_updates(self):
...
+            values = conf_obj.values
+            filename = conf_obj.filename
+            release = conf_obj.release
+            if not values:
+                # pylint: disable=line-too-long
+                msg = "Key '{}' is deprecated since {release}. Please remove it from your config".format(key, release=release)
+            else:
+                values = '/'.join(values)
+                # pylint: disable=line-too-long
+                msg = "Key '{}' with value(s) {} was found (deprecated since {}). Please adapt your config".format(key, values, release)
```
smithfarm: I can't get this line to display fully in my browser window . . . too long.

srv/modules/runners/validate.py
```diff
@@ -1046,6 +1046,26 @@ def ceph_updates(self):
             msg = ("On or more of your minions have updates pending that might cause ceph-daemons to restart. This might extend the duration of this Stage depending on your cluster size. If you want to find out which packages will be updated, you can get the full list with salt -I 'cluster:ceph' packagemanager.list_ceph_updates")
```
smithfarm: This line is extremely long.

srv/modules/runners/validate.py
```diff
@@ -1054,6 +1074,150 @@ def report(self):
...
+        issues (list): List of found incidents
+    """
+    def __init__(self):
+
+        self.base_path = '/srv/salt/ceph/configuration/files'
+        self.map_file = '{}/deprecated_map.yml'.format(self.base_path)
+        self.conf_path = '{}/ceph.conf.d'.format(self.base_path)
+        self.suffix = '.conf'
+        self.files = glob.glob("{path}/*{suffix}".format(path=self.conf_path,
+                                                         suffix=self.suffix))
```
tserong: Can we add something like the following (untested), to pick up the imported config file produced by engulf?

```
        if os.path.isfile("/srv/salt/ceph/configuration/files/ceph.conf.import"):
            files.append("/srv/salt/ceph/configuration/files/ceph.conf.import")
```

srv/modules/runners/validate.py
```diff
@@ -1054,6 +1074,150 @@ def report(self):
...
+        return obj
+
+    def run(self):
+        """
+        Returns:
+            list: contains objects of DeprecatedConf
+        """
+        for filename in self.files:
+            for line in self.read_lines(filename):
+                conf_object = self.check_line(line)
```
tserong: Python includes a configparser class that knows how to read .ini style config files, which we've used in a couple of places already, which would save adding functions to extract key-value data. Something like this:

````
conf = configparser.RawConfigParser()
with open(filename) as lines:
    conf.read_string('\n'.join(line.strip() for line in lines))
for section in conf.sections():
    if conf.has_option(section, "key to check goes here"):
        # found deprecated option, warn about it
````

jschmid1:
retest this please

srv/modules/runners/validate.py
tserong: > The ConfigParser needs sections in the .conf(.ini) files, which we never agreed to use in any of (/srv/salt/ceph/configuration/files/ceph.conf.d/*.conf).

Ah. That difference didn't occur to me :-/

> Alternatively we could you python-configobj which seems to be able to load a .ini file without sections.

Looks like we've already got "Requires: python3-configobj" in the spec file, so, sure, go for it :-)

jschmid1: Pushed a commit to address your suggestions.

* underscores in keys will be normalized
* configobj is being used
* adapted tests and added tests for normalization method

PR #1467, opened 2018-11-14
Consider iptables not installed as passing

swiftgist:
The passing message will be "not installed" only if all minions do not
have iptables installed.  Otherwise, the original behavior takes
precedent.

bsc#1115990

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/modules/utils/ready.py
+++ b/srv/modules/utils/ready.py
@@ -53,20 +53,27 @@ class Checks(object):
         Scan all minions for the default firewall settings.  Set warnings
         for any differences.
         """
-        contents = self.local.cmd(self.search, 'cmd.shell',
+        contents = self.local.cmd(self.search, 'cmd.run_all',
                                   ['/usr/sbin/iptables -S'],
                                   tgt_type="compound")
+
         for minion in contents:
+            if contents[minion]['retcode'] == 127:
+                self.passed['firewall'] = "not installed"
+                continue
+            if 'firewall' in self.passed:
+                del self.passed['firewall']
+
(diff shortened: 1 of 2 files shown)


srv/modules/utils/ready.py
```diff
@@ -53,20 +53,27 @@ def firewall(self):
...
+        contents = self.local.cmd(self.search, 'cmd.run_all',
                                   ['/usr/sbin/iptables -S'],
                                   tgt_type="compound")
+
         for minion in contents:
+            if contents[minion]['retcode'] == 127:
+                self.passed['firewall'] = "not installed"
+                continue
+            if 'firewall' in self.passed:
+                del self.passed['firewall']
```
tserong: I'm confused -- doesn't this delete the value that was set in the first iteration of this loop?

swiftgist: Yeah, I was trying to do the minimal impact to the current behavior.  In other words, the only way passed is set to "not installed" is if all minions hit the continue.  Then the condition to set passed to "disabled" is not triggered since it's already set.

If any minion has iptables return some other code other than 127, then all bets are off.  Delete the key and see if a warning is triggered.  If not, then the official passed is "disabled".  Another way to think about this is that "not installed" is a bad response if at least one minion does have iptables installed.

tserong: Ah! Now I understand, thank you :-) Is it worth adding a comment to that effect?

kshtsk:
retest this please

PR #1468, opened 2018-11-14
[SES5] Add rebuild orchestration, obliterate state

swiftgist:
Rather than migrate storage nodes incrementally, the rebuild
orchestration removes all OSDs and then recreates the OSDs.  The
filequeue and lock allow for graceful restarts after resolving
failures.

bsc#1105082

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/salt/ceph/rebuild/default.sls
+++ b/srv/salt/ceph/rebuild/default.sls
@@ -1,0 +1,93 @@
+
+{% set master = salt['pillar.get']('master_minion') %}
+
+{% if salt['saltutil.runner']('disengage.check', cluster='ceph') == False %}
+safety is engaged:
+  salt.state:
+    - tgt: {{ master }}
+    - name: "Run 'salt-run disengage.safety' to disable"
+    - failhard: True
+
+{% endif %}
+
+wait on healthy cluster:
+  salt.state:
+    - tgt: {{ master }}
+    - tgt_type: compound
+    - sls: ceph.wait.until.OK
+    - failhard: True
+
+{% if salt['saltutil.runner']('filequeue.empty', queue='rebuild') %}
+{% for host in salt.saltutil.runner('select.minions', cluster='ceph', roles='storage') %}
+queue {{ host }}:
+  salt.runner:
+    - name: filequeue.add
+    - queue: 'rebuild'
+    - item: {{ host }}
+
+{% endfor %}
+
+advise rebuild:
+  salt.runner:
+    - name: advise.rebuild
+
+{% endif %}
+
+{% for host in salt['saltutil.runner']('filequeue.array', queue='rebuild') %}
+lock obliterate {{ host }}:
+  salt.state:
+    - tgt: {{ host }}
+    - tgt_type: compound
+    - sls: ceph.obliterate.lock
+
+remove osds {{ host }}:
+  salt.state:
+    - tgt: {{ host }}
+    - tgt_type: compound
+    - sls: ceph.obliterate
+    - failhard: True
+    - require: 
+      - salt: lock obliterate {{ host }}
+
+create osds {{ host }}:
+  salt.state:
+    - tgt: {{ host }}
+    - tgt_type: compound
+    - sls: ceph.osd
+    - failhard: True
+
+update grains {{ host }}:
+  salt.state:
+    - tgt: {{ host }}
+    - tgt_type: compound
+    - sls: ceph.osd.grains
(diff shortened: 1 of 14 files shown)


srv/salt/ceph/rebuild/default.sls
```diff
@@ -0,0 +1,93 @@
...
+    - tgt: {{ host }}
+    - tgt_type: compound
+    - sls: ceph.osd.grains
+    - failhard: True
+
+wait on healthy cluster {{ host }}:
+  salt.state:
+    - tgt: {{ master }}
+    - tgt_type: compound
+    - sls: ceph.wait.1hour.until.OK
```
tserong: I wonder, if this wait expires, then the orchestration is re-run, that means it'll destroy the OSDs again. Should we perhaps move this check to after `filequeue.remove` and `ceph.obliterate.unlock`?

jschmid1:
We need functional (and/or at least unit) tests for this to be merged.

jschmid1:
@susebot run teuthology

srv/salt/ceph/rebuild/default.sls
```diff
@@ -0,0 +1,93 @@
...
+    - name: advise.rebuild
+
+{% endif %}
+
+{% for host in salt['saltutil.runner']('filequeue.array', queue='rebuild') %}
+lock obliterate {{ host }}:
+  salt.state:
+    - tgt: {{ host }}
+    - tgt_type: compound
+    - sls: ceph.obliterate.lock
```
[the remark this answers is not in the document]
swiftgist: Yeah, it would be redundant.  I just verified with a small example.  I think showing the "remove osds" will not execute in red because the lock failed helps get the point across.

srv/salt/ceph/rebuild/default.sls
swiftgist: If you flip the check, you could inadvertently skip a minion that is having issues.  Imagine a misconfiguration of the pillar.  The obliterate removes all the OSDs, the create does nothing because there is no config, we remove the minion from the queue because we're "done", but then we find out that the Ceph cluster is unhealthy since it's missing some number of OSDs.  If the admin restarts now, we will skip this minion and then proceed to obliterating the next one putting their cluster in a worse state.  For some subset of stubborn admins, I could see this command being rerun multiple times without investigation and obliterating several minions.

With the current ordering, the admin would be stuck on resolving the issue with this minion.  What you mention is true though.  Rerunning the orchestration on a obliterated/partially rebuilt minion would essentially start over.  This topic also came up with Martin in the saga of the related topic.

The issue is the rebuild orchestration could fail in a couple of places.  If the OSD removal fails, it's fine to investigate, address the issue and then rerun the rebuild.  If the creation of the first or second OSD fails, addressing the issue and rerunning the rebuild may still be the easiest/fastest for the admin.  The time lost on creating an OSD or two is less than having to lookup running additional Salt commands manually.  However, if twenty of twenty-four OSDs completed, the admin would not want to start over.  The admin would run the `ceph.osd`, `ceph.osd.grains`, remove the minion from the rebuild queue and `ceph.obliterate.unlock`.  Essentially, the final steps of the rebuild orchestration.

We really do not have any data on how often a failure is more likely to occur when just starting vs. nearly complete.  If the nearly complete scenario becomes a common problem, I think we could extend this then.  Otherwise, one orchestration and a filequeue is already too many moving parts for some. :)

tserong: Makes sense :-)

swiftgist:
I have added some unit tests for filequeue.  I do not feel that they add anything useful since these are essentially wrappers.  The remove/multiple.sls covers the obliterate state.  The rebuild orchestration would have to be exercised outside of Salt.

jschmid1:
@swiftgist ^^

swiftgist:
@jan--f what functional test is not covered?  The individual states are the same as the migration.

jschmid1:
and yet we don't have a `functests.1node.rebuild` equivalent that can be exectued.

swiftgist:
And `/srv/salt/ceph/rebuild` is still an orchestration.

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
retest this please

PR #1470, opened 2018-11-16
monitoring: fix pool usage gauge

jan--f:
Fixes: bsc#1116114


jan--f:
not easily, we'd have to test that panel's show what they claim they show. That would be hard enough, doing this automatically in a web frontend adds some complexity.

jschmid1:
fair enough.

PR #1473, opened 2018-11-19
tests: don't call helper_specs fixture directly

tserong:
Prior to this change, we get an error 'RemovedInPytest4Warning: Fixture
"helper_specs" called directly. Fixtures are not meant to be called
directly, are created automatically when test functions request them as
parameters. See https://docs.pytest.org/en/latest/fixture.html for more
information.'

-----------------

**Checklist:**
- [ ] Adapted documentation


tserong:
`make test` runs through clean for me now with this fix.

PR #1474, opened 2018-11-20
Add runner to enable/disable storage enclosure LED

votdev:
Add runner to enable/disable storage enclosure LED.

The interface is based on
- https://docs.google.com/document/d/1kGoGq5qRWBnuyH7ZTCGfb16uD7sd9hddIY6mqtFy8q8/edit#
- https://pad.ceph.com/p/blinky-lights

Examples:
```
$ salt-run disk_led.device data2.ceph vdf fault off
$ salt-run disk_led.device data2.ceph /dev/vdf fault on
$ salt-run disk_led.device data2.ceph /dev/disk/by-id/scsi-SATA_ST3200XXXX2AS_5XWXXXR6 ident off
$ salt-run disk_led.device data1.ceph SanDisk_X400_M.2_1260_512GB_558924904728 ident on
```

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/modules/runners/disk_led.py
+++ b/srv/modules/runners/disk_led.py
@@ -1,0 +1,115 @@
+# -*- coding: utf-8 -*-
+"""
+This runner is used to turn on/off the ident and fault LED of a disk.
+"""
+from __future__ import absolute_import
+
+import logging
+import re
+
+import salt.client
+
+log = logging.getLogger(__name__)
+
+
+def _process(hostname, device_name, led, status):
+    """
+    Turn on/off the specified LED of the disk on the given host.
+    """
+    assert not device_name.startswith('/dev/')
+    assert led in ['ident', 'fault']
+
+    local_client = salt.client.LocalClient()
+    device_file = '/dev/{}'.format(device_name)
+    status = 'on' if status in [True, 'on'] else 'off'
+
+    # Get the configuration of the command to turn on/off the ident
+    # or fault LED on the target host.
+    result = local_client.cmd(
+        hostname, 'pillar.get', ['disk_led:cmd'], tgt_type='compound')
+    if not result or hostname not in result.keys():
+        raise RuntimeError(
+            'Failed to get "disk_led:cmd" configuration from pillar')
+    cmd_cfg = result[hostname]
+    if not isinstance(cmd_cfg, dict):
+        raise RuntimeError('Invalid "disk_led:cmd" configuration')
+
+    print('Turning {} the {} LED of disk "{}" on host "{}" ...'.format(
+        status, led, device_file, hostname))
+    _cmd_run(hostname, cmd_cfg[led][status].format(device_file=device_file))
+
+
+def _cmd_run(hostname, cmd):
+    """
+    Run the specified command line.
+    .. note:: This function has been introduced for easier unit testing.
+    """
+    local_client = salt.client.LocalClient()
+    result = local_client.cmd(
+        hostname, 'cmd.shell', [cmd], tgt_type='compound', full_return=True)
+    if result and hostname in result.keys():
+        if result[hostname]['retcode'] != 0:
+            raise RuntimeError('Failed to execute "{}" on "{}": {}'.format(
+                cmd, hostname, result[hostname]['ret']))
+
+
+def device(hostname, identifier, led, status):
+    """
+    Manage the storage enclosure LED by device name and host.
+    :param hostname: The host where the storage device is located.
+    :type hostname: str
+    :param identifier: The storage device identifier, e.g.
+        - sdg
+        - SanDisk_X400_M.2_1260_512GB_558924904728
+        - <VENDOR>_<MODEL>_<SERIAL>
+    :type identifier: str
+    :param led: The LED to be processed. This can be 'ident' or 'fault'.
+    :type led: str
+    :param status: Set to 'on' to switch on the light, otherwise 'off'.
+    :type status: str
+    """
+    device_name = None
+    local_client = salt.client.LocalClient()
+
+    grains = local_client.cmd(
+        hostname, 'grains.get', ['disks'], tgt_type='compound')
+    if hostname in grains:
+        # Process list of device names.
+        # {
+        #   ...
+        #   'data1.ceph': ['dm-1', 'vdf', 'vdd', 'sdg', ...],
+        #   ...
+        # }
+        for disk in grains[hostname]:
+            if disk == identifier:
+                device_name = disk
+                break
+
+            # Get all udev-created device symlinks.
+            # {
+            #   'data1.ceph': [
+            #     'disk/by-id/wwn-0x4003c435a4d43cd8',
+            #     'disk/by-path/pci-0000:00:15.0-ata-3',
+            #     'disk/by-id/ata-SanDisk_X400_M.2_2280_512GB_26453645624767'
+            #   ]
+            # }
+            device_links = local_client.cmd(
+                hostname,
+                'udev.links', ['/dev/{}'.format(disk)],
+                tgt_type='compound')
+
+            for device_link in device_links[hostname]:
+                # pylint: disable=line-too-long
+                # Search for device names like:
+                # - Crucial_CT1024M550SSD1_14160C164100 (disk/by-id/(ata|scsi|.+)-Crucial_CT1024M550SSD1_14160C164100)
+                # - wwn-0x4021b384a4d42ca5 (disk/by-id/wwn-0x4021b384a4d42ca5)
+                if re.match(r'^.+\/(ata|scsi|wwn)-{}$'.format(identifier),
+                            device_link):
+                    device_name = disk
+                    break
+
+    if device_name is not None:
+        _process(hostname, device_name, led, status)
+        return ''
+    return 'Could not find device "{}" on host "{}"'.format(
+        identifier, hostname)
--- a/srv/pillar/ceph/disk_led.sls
+++ b/srv/pillar/ceph/disk_led.sls
@@ -1,0 +1,15 @@
+# This is the default configuration for the storage enclosure LED blinking.
+# The placeholder {device_file} will be replaced with the device file of
+# the disk when the command is executed.
+#
+# Have a look into the /srv/pillar/ceph/README file to find out how to
+# customize this configuration per minion/host.
+
+disk_led:
+  cmd:
+    ident:
+      'on': lsmcli local-disk-ident-led-on --path '{device_file}'
+      'off': lsmcli local-disk-ident-led-off --path '{device_file}'
+    fault:
+      'on': lsmcli local-disk-fault-led-on --path '{device_file}'
+      'off': lsmcli local-disk-fault-led-off --path '{device_file}'
--- a/Makefile
+++ b/Makefile
@@ -196,6 +196,7 @@ copy-files:
 	install -m 644 srv/pillar/ceph/init.sls $(DESTDIR)/srv/pillar/ceph/
 	install -m 644 srv/pillar/ceph/deepsea_minions.sls $(DESTDIR)/srv/pillar/ceph/
 	install -m 644 srv/pillar/ceph/blacklist.sls $(DESTDIR)/srv/pillar/ceph/
+	install -m 644 srv/pillar/ceph/disk_led.sls $(DESTDIR)/srv/pillar/ceph/
 	install -d -m 755 $(DESTDIR)/srv/pillar/ceph/stack
 	install -m 644 srv/pillar/ceph/stack/stack.cfg $(DESTDIR)/srv/pillar/ceph/stack/stack.cfg
 	install -m 644 srv/pillar/top.sls $(DESTDIR)/srv/pillar/
--- a/deepsea.spec.in
+++ b/deepsea.spec.in
@@ -450,6 +450,7 @@ systemctl try-restart salt-api > /dev/null 2>&1 || :
 %config %attr(-, salt, salt) /srv/pillar/ceph/benchmarks/templates/*.j2
 %config(noreplace) %attr(-, salt, salt) /srv/pillar/ceph/deepsea_minions.sls
 %config(noreplace) %attr(-, salt, salt) /srv/pillar/ceph/blacklist.sls
+%config(noreplace) %attr(-, salt, salt) /srv/pillar/ceph/disk_led.sls
 %config %attr(-, salt, salt) /srv/pillar/ceph/stack/stack.cfg
 /srv/salt/_modules/*.py*
 /srv/salt/_states/*.py*
--- a/srv/pillar/ceph/init.sls
+++ b/srv/pillar/ceph/init.sls
@@ -2,3 +2,4 @@
 
 {% include 'ceph/deepsea_minions.sls' ignore missing %}
 {% include 'ceph/blacklist.sls' ignore missing %}
+{% include 'ceph/disk_led.sls' ignore missing %}
--- a/srv/salt/ceph/packages/common/default.sls
+++ b/srv/salt/ceph/packages/common/default.sls
@@ -27,6 +27,7 @@ stage prep dependencies suse:
       - tuned
       - libgio-2_0-0
       - polkit
+      - libstoragemgmt
     - fire_event: True
     - refresh: True
 
@@ -72,6 +73,7 @@ stage prep dependencies CentOS:
       - hwinfo
       - python-netaddr
       - jq
+      - libstoragemgmt
     - fire_event: True
     - refresh: True
 
--- a/tests/unit/runners/test_disk_led.py
+++ b/tests/unit/runners/test_disk_led.py
@@ -1,0 +1,102 @@
+import pytest
+import mock
+from srv.modules.runners import disk_led
+
+grains_get_result = {"data1.ceph": ["vdf"]}
+udev_links_result = {
+    "data1.ceph": [
+        "disk/by-id/wwn-0x4003c435a4d43cd8",
+        "disk/by-path/pci-0000:00:15.0-ata-3",
+        "disk/by-id/ata-SanDisk_X400_M.2_2280_512GB_26453645624767"
+    ]
+}
+pillar_get_result = {
+    "data1.ceph": {
+        "ident": {
+            "on": "xxx '{device_file}'",
+            "off": "yyy '{device_file}'"
+        },
+        "fault": {
+            "on": "aaa '{device_file}'",
+            "off": "bbb '{device_file}'"
+        }
+    }
+}
+invalid_pillar_get_result = {"data1.ceph": ""}
+
+
+class TestDiskLed:
+    @mock.patch('srv.modules.runners.disk_led._process', autospec=True)
+    @mock.patch('salt.client.LocalClient', autospec=True)
+    def test_device_by_device_name(self, mock_localclient, mock_process):
+        mock_localclient.return_value.cmd.side_effect = [
+            grains_get_result, udev_links_result
+        ]
+
+        disk_led.device('data1.ceph', 'vdf', 'ident', 'on')
+        mock_process.assert_called_once_with('data1.ceph', 'vdf', 'ident',
+                                             'on')
+
+    @mock.patch('srv.modules.runners.disk_led._process', autospec=True)
+    @mock.patch('salt.client.LocalClient', autospec=True)
+    def test_device_by_device_id(self, mock_localclient, mock_process):
+        mock_localclient.return_value.cmd.side_effect = [
+            grains_get_result, udev_links_result
+        ]
+
+        disk_led.device('data1.ceph',
+                        'SanDisk_X400_M.2_2280_512GB_26453645624767', 'fault',
+                        'off')
+        mock_process.assert_called_once_with('data1.ceph', 'vdf', 'fault',
+                                             'off')
+
+    @mock.patch('srv.modules.runners.disk_led._process', autospec=True)
+    @mock.patch('salt.client.LocalClient', autospec=True)
+    def test_device_failure(self, mock_localclient, mock_process):
+        mock_localclient.return_value.cmd.return_value = {}
+
+        result = disk_led.device('data2.ceph', 'sdg', 'ident', 'off')
+        assert not mock_process.called
+        assert result == 'Could not find device "sdg" on host "data2.ceph"'
+
+    @mock.patch('salt.client.LocalClient', autospec=True)
+    def test_process_get_pillar_failed(self, mock_localclient):
+        mock_localclient.return_value.cmd.return_value = []
+
+        with pytest.raises(RuntimeError):
+            disk_led._process('data1.ceph', 'vdf', 'ident', 'on')
+
+    @mock.patch('salt.client.LocalClient', autospec=True)
+    def test_process_invalid_config(self, mock_localclient):
+        mock_localclient.return_value.cmd.return_value = invalid_pillar_get_result
+
+        with pytest.raises(RuntimeError):
+            disk_led._process('data1.ceph', 'vdf', 'ident', 'on')
+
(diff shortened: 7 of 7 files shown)


jschmid1:
Let us know when it's ready to review

votdev:
> Let us know when it's ready to review

It's ready for review.

srv/pillar/ceph/disk_led.sls
```diff
@@ -0,0 +1,15 @@
+# This is the default configuration for the storage enclosure LED blinking.
+# The placeholder {device_file} will be replaced with the device file of
+# the disk when the command is executed.
+#
+# Have a look into the README file to find out how to customize this
```
tserong: Maybe say `/srv/pillar/ceph/README` rather than 'the README file'?

votdev: Done

votdev:
> The Orchestrator CLI spec mentions specifying primary,journal,db or all when turning OSD LEDs on/off, so you could e.g.: turn the LED on for an OSD and its journal device, or just the OSD, etc. Can we add this?

You're right, this is missing. Any idea how to get the required information from Ceph?

> Given the default tool for fiddling with the LEDs is lsmcli, we should ensure libstoragemgmt is installed by default (add it to srv/salt/ceph/packages/common/default.sls)

Fixed for SUSE and CentOS. Ubuntu does not provide libstoragemgmt packages anymore.

> The commit message seems to be missing the example (it just has the "Examples:" heading)

Fixed

jschmid1:
This is not yet ready to review (judging by the labels), right @votdev

srv/salt/_modules/osd.py
```diff
@@ -232,6 +233,271 @@ def tree_from_any():
     return _tree(**kwargs)
 
 
+def _get_parent_device_name(device_name):
+    """
```
votdev: This could be done nicely with pyudev
```
context = pyudev.Context()
device = pyudev.Devices.from_device_file(context, '/dev/{}'.format(device_name))
return basename(device.parent.device_node)
```

srv/modules/runners/disk_led.py
```diff
@@ -0,0 +1,181 @@
...
+    #     }
+    #   }]
+    # }
+    for hostname in osd_infos:
+        filtered_osd_infos = list([x for x in osd_infos[hostname]
+                                   if x['id'] == osd_id])
+        try:
+            osd_info = filtered_osd_infos[0]
+        except IndexError:
+            continue
```
jan--f: Can we please not abuse Exceptions for checking 
``` python
    if not filtered_osd_infos:
        continue
    osd_info = filtered_osd_infos[0]
```

srv/modules/runners/disk_led.py
```diff
@@ -0,0 +1,181 @@
...
+    Manage the storage enclosure LED by OSD identifier.
+    :param osd_id: The OSD id, e.g.
+        - 4
+        - osd.1
+    :type osd_id: str
+
+    :param device_type: The device to use. This can be:
+        - primary
+        - journal
+        - db
```
jan--f: Bluestore can also be deployed on 3 devices: primary, db and wal

srv/modules/runners/disk_led.py
```diff
@@ -0,0 +1,181 @@
...
+    #   }]
+    # }
+    for hostname in osd_infos:
+        filtered_osd_infos = list([x for x in osd_infos[hostname]
+                                   if x['id'] == osd_id])
+        try:
+            osd_info = filtered_osd_infos[0]
+        except IndexError:
+            continue
+        device_types = ['primary', 'journal', 'db'] if device_type == 'all' \
```
jan--f: Imho if a oneline `if` doesn't fit on one line, use the traditional syntax? This could also simpy merge into the processing below:
``` python
if device_type == 'all':
    for t in ['primary', 'journal', 'db']:
        get_device_name_and_process(t)
else:
    get_device_name_and_process(device_type)

File: srv/modules/runners/disk_led.py [hunk 6]
```diff
@@ -0,0 +1,181 @@
...
+        - /dev/disk/by-id/scsi-SATA_ST3200XXXX2AS_5XWXXXR6
+        - SanDisk_X400_M.2_1260_512GB_558924904728
+    :type identifier: str
+    :param led: The LED to be processed. This can be 'ident' or 'fault'.
+    :type led: str
+    :param status: Set to 'on' to switch on the light, otherwise 'off'.
+    :type status: str
+    """
+    device_name = None
+    identifier = re.sub(r'^\/dev\/', '', identifier)
```
jan--f: re is pretty heavy handed for removing a substring. Maybe 
``` python
if identifier.startswith('/dev/'):
    identifier = identifier[5:]
```

srv/salt/_modules/osd.py
jan--f: If this works, I'd say leave it as is. This info will be provided by ceph-volume in the very near future.

srv/modules/runners/disk_led.py
votdev: Don't understand the intention of your comment. The parameters of this function are derived from the orchestrator interface:
```
ceph orchestrator osd {ident,fault}-{on,off} {primary,journal,db,all} <osd-id>
```

jan--f: Then that interface is incomplete. Bluestore can potentially use 3 devices. Maybe it was left out intentionally as those deployments are indeed rare? @sebastian-philipp any thoughts?

srv/salt/_modules/osd.py
votdev: Had to revert all pyudev changes because the library behaves different within the Salt environment and running the code manually in a Python CLI on the Ceph node.

Maybe it is somehow related to:
https://github.com/pyudev/pyudev/issues/225
https://github.com/openattic/openattic-docker/issues/9

srv/modules/runners/disk_led.py
votdev: @liewegas Any comments on that?

sebastian-philipp: So, the question is: `journal` the same as `wal` or is `journal` the same as `db`?

sebastian-philipp: from #ceph-devel
 [18:46:57] <sageyphone> The original idea was journal was the filestore journal.
[18:47:13] <sageyphone> No equivalence between it and db or wal

votdev: Issue is addressed now

sebastian-philipp:
Relates to https://github.com/ceph/ceph/pull/25634

votdev:
@jschmid1 IMHO this PR is ready. Do you agree?

jschmid1:
There are two 'requested changes' from me and @jan--f 

> THis is on a good way. I'm not sure the disk_led.device interface necessarily needs to be so wide (accepting basically any string and then matching against device names). Imho expecting a device file there is reasonable.

and 

> There is one big problem with keyring permissions for storage nodes in get_metadata which needs to be addressed

The latter is an actual problem which causes the runner to fail for reasons described in https://github.com/SUSE/DeepSea/pull/1474#discussion_r242901148 and https://github.com/SUSE/DeepSea/pull/1474#discussion_r242906687

This need to be reworked, probably by gathering the necessary information (get_metadata()) from the runner (you have proper permissions there) and pass it down to the module. Let me know if you need more explanation on how that works exactly.

jschmid1:
@votdev It seems that all requested changes were marked 'resolved' although they kept untouched.

votdev:
> THis is on a good way. I'm not sure the `disk_led.device` interface necessarily needs to be so wide (accepting basically any string and then matching against device names). Imho expecting a device file there is reasonable.

According to the orchestrator spec this should be a device name, e.g. ``sda`` or ``Crucial_CT1024M550SSD1_14160C164100`` and no device file.

``ceph orchestrator device {ident,fault}-{on,off} <host> <devname>``
``$ ceph device ident-on Crucial_CT1024M550SSD1_14160C164100``

See also:
https://github.com/ceph/ceph/blob/master/src/pybind/mgr/orchestrator_cli/module.py#L109
https://github.com/ceph/ceph/blob/master/src/pybind/mgr/orchestrator_cli/module.py#L118

votdev:
> This need to be reworked, probably by gathering the necessary information (get_metadata()) from the runner (you have proper permissions there) and pass it down to the module.

This shouldn't be a problem.

jschmid1:
> This shouldn't be a problem.

I think it is, as you are calling a mon command from a osd node. A standalone osd node does not have the proper permission to execute such a command.

votdev:
> > This shouldn't be a problem.
> 
> I think it is, as you are calling a mon command from a osd node. A standalone osd node does not have the proper permission to execute such a command.

I meant no problem to implement :-)

jschmid1:
> I meant no problem to implement :-)

aaah :) sorry for the confusion

votdev:
retest this please

jschmid1:
rebase please

votdev:
```
11:14:25 tests/unit/_modules/test_osd.py::TestOSDWeight::test_save_defaults FAILED [ 22%]
11:14:26 tests/unit/_modules/test_osd.py::TestOSDWeight::test_save_custom_values FAILED [ 23%]
11:14:26 tests/unit/_modules/test_osd.py::TestOSDWeight::test_restore_no_files FAILED [ 23%]
11:14:27 tests/unit/_modules/test_osd.py::TestOSDWeight::test_restore FAILED      [ 23%]
```
The make test fails because of tests that are not related to this PR. How shall we proceed?

jschmid1:
> The make test fails because of tests that are not related to this PR. How shall we proceed?

I'll have a look, but let's ignore them for this PR

srv/salt/_modules/osd.py
```diff
@@ -232,6 +233,214 @@ def tree_from_any():
...
+    result = []
+
+    if not osd_metadata:
+        return result
+
+    # Get a list of OSD ID's on this host.
+    osd_ids = list_()
+    if not osd_ids:
+        return result
```
jschmid1: A logging message would be nice here:

"No OSDs found on host {self}. Aborting"

srv/modules/runners/disk_led.py
```diff
@@ -0,0 +1,208 @@
...
+    :type device_type: str
+    :param led: The LED to be processed. This can be 'ident' or 'fault'.
+    :type led: str
+    :param status: Set to 'on' to switch on the light, otherwise 'off'.
+    :type status: str
+    """
+    osd_id = int(re.sub(r'^osd\.', '', str(osd_id)))
+    local_client = salt.client.LocalClient()
+
+    osd_metadata = _get_metadata()
```
jschmid1: That returns in the form of `{'host': [{content}]}`.

You are passing the data straight down to `osd.details` where you do a `for osd_metadata_v in osd_metadata:`.
The passed metadata is not a `list` though.

What you are looking for is the `values` of the metadata.

``` python
list(osd_metadata.values())
```
I'd also suggest to validate all output you get from minions. There is always a chance you get bogus returns due to several reasons (modules not synced, timeouts etc). All consecutive calls would fall with the first error which makes it also very hard to debug

jschmid1:
Sorry that it took me so long to come back to this. I'm currently trying this PR again.

srv/pillar/ceph/disk_led.sls
```diff
@@ -0,0 +1,15 @@
+# This is the default configuration for the storage enclosure LED blinking.
```
jschmid1: This file needs to be installed and registered (deepsea.spec.in and Makefile)

srv/modules/runners/disk_led.py
```diff
@@ -0,0 +1,208 @@
...
+
+    # Get OSD information from all storage nodes.
+    osd_details = local_client.cmd(
+        'I@roles:storage', 'osd.details', [osd_metadata], tgt_type='compound')
```
jschmid1: the argument to `osd.details` is then already in the form of a `list`. You can either drop the square brackets here or flatten the list above.

srv/modules/runners/disk_led.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    Turn on/off the specified LED of the disk on the given host.
+    """
+    local_client = salt.client.LocalClient()
+    device_file = '/dev/{}'.format(device_name)
+    status = 'on' if status in [True, 'on'] else 'off'
+
+    # Get the configuration of the command to turn on/off the ident
+    # or fault LED on the target host.
+    cmd_cfg = local_client.cmd(
```
jschmid1: This will fail due to the missing Makefile and deepsea.spec.in entries.

As already mentioned above, you are working with a distributed system where returns are not always guaranteed. It may be wise to validate output from external calls before passing them along.

srv/modules/runners/disk_led.py
```diff
@@ -0,0 +1,208 @@
...
+        hostname, 'pillar.get', ['disk_led:cmd'],
+        tgt_type='compound')[hostname]
+
+    print("Turning {} the {} LED of disk '{}' on host '{}' ...".format(
+        status, led, device_file, hostname))
+    _cmd_run(hostname, cmd_cfg[led][status].format(device_file=device_file))
```
jschmid1: This **will** return something.

On my machines.

```
data1:~ # lsmcli local-disk-ident-led-on --path /dev/vdb
NO_SUPPORT(153): SCSI VPD page 0x83 is not supported
data1:~ # echo $?
4
```
Which is quite valuable information.

Please consider to validate and report them back to the user.

srv/modules/runners/disk_led.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def _cmd_run(hostname, cmd):
+    """
+    Run the specified command line.
+    .. note:: This function has been introduced for easier unit testing.
+    """
+    local_client = salt.client.LocalClient()
+    local_client.cmd(hostname, 'cmd.run', [cmd], tgt_type='compound')
```
jschmid1: cmd.shell is preferred over cmd.run

srv/modules/runners/disk_led.py
```diff
@@ -0,0 +1,208 @@
...
+        log.debug("hostname: {} osds: {}".format(
+            hostname, pprint.pformat(filtered_osd_details)))
+        if not filtered_osd_details:
+            continue
+        osd_info = filtered_osd_details[0]
+        device_types = [device_type]
+        if device_type == 'all':
+            device_types = ['primary', 'wal', 'db', 'journal']
+        for dev_type in device_types:
+            device_name = osd_info[dev_type]['device_name']
```
jschmid1: device_name is in the format of '/dev/dm-0'. (at least on my cluster, deployed with vagrant-ceph)

in the `_process(..)` function you are prepending the device_name with '/dev/' which results in:

`'/dev//dev/dm-0'`

srv/modules/runners/disk_led.py
```diff
@@ -0,0 +1,208 @@
...
+
+log = logging.getLogger(__name__)
+
+
+def _process(hostname, device_name, led, status):
+    """
+    Turn on/off the specified LED of the disk on the given host.
+    """
+    local_client = salt.client.LocalClient()
+    device_file = '/dev/{}'.format(device_name)
```
jschmid1: As mentioned in https://github.com/SUSE/DeepSea/pull/1474/files#r260701496 , add format checks here.

jschmid1:
As the tests are passing I assume you used wrong mock-data in your tests. 

I'd suggest you to use an _actual_ cluster for testing this patch.

srv/modules/runners/disk_led.py
votdev: Can you please post the output of ``ceph osd metadata``?

jschmid1:
```
snip..
    {
        "id": 35,
        "arch": "x86_64",
        "back_addr": "[v2:172.16.2.21:6816/38848,v1:172.16.2.21:6817/38848]",
        "back_iface": "eth3",
        "bluefs": "1",
        "bluefs_single_shared_device": "1",
        "bluestore_bdev_access_mode": "blk",
        "bluestore_bdev_block_size": "4096",
        "bluestore_bdev_dev_node": "/dev/dm-0",
        "bluestore_bdev_driver": "KernelDevice",
        "bluestore_bdev_partition_path": "/dev/dm-0",
        "bluestore_bdev_rotational": "1",
        "bluestore_bdev_size": "20401094656",
        "bluestore_bdev_support_discard": "0",
        "bluestore_bdev_type": "hdd",
        "ceph_release": "nautilus",
        "ceph_version": "ceph version 14.0.1-3346-g0364ec8d2f (0364ec8d2f8783d21a7153abd3bf67361230622e) nautilus (dev)",
        "ceph_version_short": "14.0.1-3346-g0364ec8d2f",
        "cpu": "Intel Core Processor (Haswell, no TSX, IBRS)",
        "default_device_class": "hdd",
        "device_ids": "",
        "devices": "vdb",
        "distro": "sles",
        "distro_description": "SUSE Linux Enterprise Server 15 SP1",
        "distro_version": "15.1",
        "front_addr": "[v2:172.16.1.21:6816/38848,v1:172.16.1.21:6817/38848]",
        "front_iface": "eth2",
        "hb_back_addr": "[v2:172.16.2.21:6818/38848,v1:172.16.2.21:6819/38848]",
        "hb_front_addr": "[v2:172.16.1.21:6818/38848,v1:172.16.1.21:6819/38848]",
        "hostname": "data1",
        "journal_rotational": "1",
        "kernel_description": "#1 SMP Thu Feb 14 12:19:57 UTC 2019 (e1536c0)",
        "kernel_version": "4.12.14-32-default",
        "mem_swap_kb": "0",
        "mem_total_kb": "1516264",
        "network_numa_unknown_ifaces": "eth2,eth3",
        "objectstore_numa_unknown_devices": "vdb",
        "os": "Linux",
        "osd_data": "/var/lib/ceph/osd/ceph-35",
        "osd_objectstore": "bluestore",
        "rotational": "1"
    }
..snip
```

srv/modules/runners/disk_led.py
votdev: Done

srv/modules/runners/disk_led.py
votdev: Done

srv/modules/runners/disk_led.py
votdev: Done

srv/modules/runners/disk_led.py
votdev: When ``osd_metadata`` is a list of dicts and i want to pass this list to the function, then ``[osd_metadata]`` is still correct, right?

srv/pillar/ceph/disk_led.sls
votdev: Done

srv/modules/runners/disk_led.py
votdev: Fixed

srv/salt/_modules/osd.py
votdev: Done

srv/modules/runners/disk_led.py
jschmid1: `return result[master_minion]`

should do the trick

srv/pillar/ceph/disk_led.sls
```diff
@@ -0,0 +1,15 @@
...
+# The placeholder {device_file} will be replaced with the device file of
+# the disk when the command is executed.
+#
+# Have a look into the /srv/pillar/ceph/README file to find out how to
+# customize this configuration per minion/host.
+
+disk_led:
+  cmd:
+    ident:
+      'on': lsmcli local-disk-ident-led-on '{device_file}'
```
tserong: This (and the three other lsmcli invocations) all need `--path` added, e.g.: `lsmcli local-disk-ident-led-on --path '{device_file}'`

srv/modules/runners/disk_led.py
tserong: Same problem here:

```
# salt-run disk_led.osd 1 all ident on
Turning on the ident LED of disk "/dev//dev/dm-0" on host "data4.ceph" ...
Exception occurred in runner disk_led.osd: Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/salt/client/mixins.py", line 387, in _low
    data['return'] = self.functions[fun](*args, **kwargs)
  File "/srv/modules/runners/disk_led.py", line 158, in osd
    _process(hostname, device_name, led, status)
  File "/srv/modules/runners/disk_led.py", line 37, in _process
    _cmd_run(hostname, cmd_cfg[led][status].format(device_file=device_file))
  File "/srv/modules/runners/disk_led.py", line 51, in _cmd_run
    hostname, result[hostname]['ret']))
RuntimeError: Failed to execute command on "data4.ceph": NOT_FOUND_DISK(209): Disk /dev//dev/dm-0 not found
```

Looks like `blue_*_dev_node` is already the full device path.

tserong:
I'm probably going to annoy everyone now, but here goes...

It occurs to me that we may only need the `disk_led.device` commands, not the `disk_led.osd` commands.

We're going to have `ceph device {fault,ident}-{on,off}` and `ceph osd {fault,ident}-{on,off}` implemented in ceph-mgr, which in turn call out to deepsea/rook/ansible/whatever. So we definitely need DeepSea to be able to activate LEDs on specific devices, but for the OSD commands, the OSD-to-device mapping can be implemented directly inside ceph-mgr (it already knows about all the OSD metadata, right?) then it just call out to DeepSea to activate/deactivate the relevant devices.

srv/modules/runners/disk_led.py
votdev: Oh great, they changed the output format. OK, need to adapt this.

jschmid1:
> the OSD-to-device mapping can be implemented directly inside ceph-mgr

that's the spirit :>

srv/pillar/ceph/disk_led.sls
votdev: Done

srv/modules/runners/disk_led.py
```diff
@@ -0,0 +1,222 @@
...
+
+    local_client = salt.client.LocalClient()
+    result = local_client.cmd(
+        master_minion, 'osd.metadata', tgt_type='compound')
+    if not result or master_minion not in result.keys():
+        raise RuntimeError('Failed to get metadata')
+    return result[master_minion]
+
+
+def osd(osd_id, device_type, led, status):
```
liewegas: I was planning on doing all of the osd -> device mapping in ceph-mgr.  I think all we need at this layer is to turn {ident,fault} led {on,off} for a given device on a given host.

votdev:
OSD processing has been cut out.

srv/modules/runners/disk_led.py
votdev: Done

votdev:
Any news about when this PR will be merged?

jschmid1:
> Any news about when this PR will be merged?

Does it work on hardware?

tserong:
Approved because the code LGTM, but like @jschmid1 said, we still need confirmation that this behaves as expected on actual hardware.

Also, please squash the commits, and update the commit message to remove the OSD commands.

jschmid1:
any update on this @votdev ?

votdev:
@jschmid1 I contacted mlatimer and he tried to get lsmcli to work but had some troubles with the tool. I don't know how i should test this PR without real hardware and without sitting in front of it to check if the LED's are on or off. The PR includes unit tests that ensure that the configured command is executed, so we can be assume that the CLI command is executed in a production environment. So the only thing we need to clarify is that the pre-configured lsmcli command line is correct.

liewegas:
I tested it on my actual hardware and it worked, but that was from a source tree and some random incantations to set up the python environment to point to the code in the source tree.

My plan is to dist-upgrade the boxes soon so that the OS-isntalled lsmcli 
is there...

PR #1481, opened 2018-11-28
Remove populate.engulf_existing_cluster

tserong:
The engulf function is broken for LVM deployments, meaning it can't
engulf Nautilus clusters.  Rather than continue to maintain the engulf
function, and enhance it to support both ceph-disk and ceph-deploy,
I'd rather just remove it.  Existing pre-Nautlius non-DeepSea deployments
can still be engulfed with DeepSea 0.8.x.

-----------------

**Checklist:**
- [ ] Adapted documentation


tserong:
> What about engulfing existing deepsea clusters after loosing the admin node?

Engulf can help with this, but it can't ever be as good a solution as backups of the admin node. Ignoring for a moment the lack of LVM support (i.e. it won't currently work against Nautilus), engulf only ever knew how to detect things that ceph-deploy could deploy (mon, mgr, mds, osd, rgw). It won't recover a lost igw, ganesha or openattic, for example.  It's also incapable of constructing the neatly split ceph.conf.d snippets that DeepSea creates.

It would be technically possible to add some of the above missing functionality (detection of additional services, LVM support), but this would need to be developed, maintained, and tested on an ongoing basis, and even then there are still some things engulf will never be able to do reliably (ceph.conf.d magic, probably some key caps stuff), so there would always be a degree of manual screwing around when attempting to use this function for disaster recovery.

jschmid1:
@susebot run teuthology

Martin-Weiss:
> 
> 
> > What about engulfing existing deepsea clusters after loosing the admin node?
> 
> Engulf can help with this, but it can't ever be as good a solution as backups of the admin node. Ignoring for a moment the lack of LVM support (i.e. it won't currently work against Nautilus), engulf only ever knew how to detect things that ceph-deploy could deploy (mon, mgr, mds, osd, rgw). It won't recover a lost igw, ganesha or openattic, for example. It's also incapable of constructing the neatly split ceph.conf.d snippets that DeepSea creates.
> 
> It would be technically possible to add some of the above missing functionality (detection of additional services, LVM support), but this would need to be developed, maintained, and tested on an ongoing basis, and even then there are still some things engulf will never be able to do reliably (ceph.conf.d magic, probably some key caps stuff), so there would always be a degree of manual screwing around when attempting to use this function for disaster recovery.

We have an other product that is also using salt-orchestration similar to deepsea and we have exactly the problem that rolling back / recovering the status of the orchestration does not work and due to that we can not fulfill customers disaster recovery requirements.

As long as all states work against the cluster even if we re-install the admin server from scratch and we do not need an engulf for that - dropping it might be fine. But I am not sure if this requirement is fulfilled...

PR #1486, opened 2018-12-11
Assign drives based on filter rules (DriveGroups) SES-150

jschmid1:

This is a first draft of the code that will transform a drive_group proposal to an actual list of devices.
Please read the wiki pages for [DriveGroups](https://github.com/SUSE/DeepSea/wiki/Drive-Groups) to get an idea on how specifications will look like.

~If you want to give this patch a try, there is a `drive_group.sls` in `/srv/pillar/ceph/` which can be adapted to your needs.~

I'll write a proper howto in the wikis soon.

~In this implementation the drive_group will reside in the pillar again..~
it will not. It will pull it from a file an stuff it down to the minions.

Expected to fail:

~* linter (there is a FIXME in it, which shouldn't be removed, yet - works besides this)~
done
~The runner call is `disks.test` which is just a temp name.~
done
~I have not added functionality to convert the output to ceph-volume understandable commands yet. (wip)~
done

The naming question also remains. A couple of suggestions:
salt-run
* drivegroups.generate
* drivegroups.match
* drivegroups.allocate
* disks.filter ..

I went with dg.py and disks.py for now.. Not very convinced with my naming scheme though..

~Unittests are not added yet, as the implementation will probably change a bit. I'll add them once we are all happy with this implementation.~

Added

----------

It may be noted that I used the optional static typing system introduced with python3 to harden the code. It runs well on any python3 environment. However this should not be used for any code that may get backported.. This is not one of those candidates I think.

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,4 +125,5 @@ TAGS
 .mypy_cache/
 
 # projectile files
+.#*
 .projectile
--- a/srv/salt/_modules/dg.py
+++ b/srv/salt/_modules/dg.py
@@ -1,0 +1,822 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+""" This module will match disks based on applied filter rules
+
+Internally this will be called 'DriveGroups'
+"""
+
+from __future__ import absolute_import
+import json
+import re
+import logging
+from typing import Set, Tuple
+
+log = logging.getLogger(__name__)
+
+USAGE = """
+
+The (D)rive (G)roup module exists to filter for devices on the node
+based on the drive group specification received from the master.
+
+It mainly exposes three functions (They are not ment to be called manually)
+
+All function expect valid drive group specs as arguments.
+
+list_drives:
+
+This returns a dict() of matching drives.
+
+c_v_commands:
+
+Constructs valid ceph-volume calls based on the drive group specs and returns them
+
+deploy:
+
+A simple function that calls c_v_commands and executes it on the minion.
+
+
+"""
+
+
+class FilterNotSupported(Exception):
+    """ A critical error when the user specified filter is unsupported
+    """
+    pass
+
+
+class UnitNotSupported(Exception):
+    """ A critical error which encouters when a unit is parsed which
+    isn't supported.
+    """
+    pass
+
+
+class Filter(object):
+    """ Filter class to assign properties to bare filters.
+
+    This is a utility class that tries to simplify working
+    with information comming from a textfile/salt (drive_group.yaml)/(pillar)
+
+    """
+
+    def __init__(self, **kwargs):
+        self.name: str = str(kwargs.get('name', None))
+        self.matcher = kwargs.get('matcher', None)
+        self.value: str = str(kwargs.get('value', None))
+        self._assign_matchers()
+        log.debug("Initializing filter for {} with value {}".format(
+            self.name, self.value))
+
+    @property
+    def is_matchable(self) -> bool:
+        """ A property to indicate if a Filter has a matcher
+
+        Some filter i.e. 'limit' or 'osd_per_device' are valid filter
+        attributes but cannot be applied to a disk set. In this case
+        we return 'None'
+        :return: If a matcher is present True/Flase
+        :rtype: bool
+        """
+        return self.matcher is not None
+
+    def _assign_matchers(self) -> None:
+        """ Assign a matcher based on filter_name
+
+        This method assigns an individual Matcher based
+        on `self.name` and returns it.
+        """
+        if self.name == "size":
+            self.matcher = SizeMatcher(self.name, self.value)
+        elif self.name == "model":
+            self.matcher = SubstringMatcher(self.name, self.value)
+        elif self.name == "vendor":
+            self.matcher = SubstringMatcher(self.name, self.value)
+        elif self.name == "rotational":
+            self.matcher = EqualityMatcher(self.name, self.value)
+        elif self.name == "all":
+            self.matcher = AllMatcher(self.name, self.value)
+        else:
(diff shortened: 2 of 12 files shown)


jschmid1:
> Let's say I just want to use all the disks as OSDs, regardless of model and size. Can I express that, or do I always need at least one of model/size/rotational?

In the current implementation this is not possible, but maybe we should maintain the possibility to 'default' like we do with the proposal runner.

jschmid1:
> Also... I know we want this to be declarative, but I have a nagging feeling somehow we'll end up needing to support explicitly specifying devices by path for someone, somewhere (plus, the drive groups google doc talks about specifying device paths), so... Do we need to add this?

mhh, I can't think of a scenario where we would _have_ to do this(this doesn't mean that the concern is invalid).. However we may add a 'overwrite'(?) section where one can specify disk identifiers/paths.

tserong:
> In the current implementation this is not possible, but maybe we should maintain the possibility to 'default' like we do with the proposal runner.

Yeah, I think we should. It makes it easier to use for the case where you just want one OSD per disk.

tserong:
> mhh, I can't think of a scenario where we would _have_ to do this(this doesn't mean that the concern is invalid).. However we may add a 'overwrite'(?) section where one can specify disk identifiers/paths.

Maybe let's worry about it if someone comes up with an actual use case (rather than my "nagging feeling"). @jan--f and I had a chat about this the other day, and we couldn't think of a scenario where it would be strictly necessary to have.

jan--f:
> Let's say I just want to use all the disks as OSDs, regardless of model and size. Can I express that, or do I always need at least one of model/size/rotational?

I think its worth adding some syntax for this and keep this use case explicit. This really only matter for the `data_devices` (I believe). Maybe simply like this: `data_devices: all`?

> Also... I know we want this to be declarative, but I have a nagging feeling somehow we'll end up needing to support explicitly specifying devices by path for someone, somewhere (plus, the drive groups google doc talks about specifying device paths), so... Do we need to add this?

As you mentioned, probably not. However it should be fairly straight forward to add a `path` key to the `device_spec` and do a glob matching or some such.

>  It may be noted that I used the optional static typing system introduced with python3 to harden the code. It runs well on any python3 environment. However this should not be used for any code that may get backported.. This is not one of those candidates I think.

Nice! I wouldn't mind seeing more type hints.

jan--f:
> I think its worth adding some syntax for this and keep this use case explicit. This really only matter for the data_devices (I believe). Maybe simply like this: data_devices: all?

Fwiw I updated [DriveGroups](https://github.com/SUSE/DeepSea/wiki/Drive-Groups) accordingly.

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
retest this please

kshtsk:
retest this please

jschmid1:
for the record:

There was an attempt to use an external_pillar in order to avoid the master<->minion roundtrips.

https://gist.github.com/jschmid1/10f4087eba911de708c63b4f7405d963

That however doesn't work due to deadlocks/deathspirals..

jschmid1:
@susebot run teuthology

jschmid1:
@susebot run teuthology

jschmid1:
Yikes, 

```
 File "/usr/lib/python3.6/site-packages/salt/utils/args.py", line 231, in _getargspec
    raise ValueError('Function has keyword-only arguments or annotations'
ValueError: Function has keyword-only arguments or annotations, use getfullargspec() API which can support them

```

jschmid1:
@susebot run teuthology

jschmid1:
>  File "/usr/lib/python3.6/site-packages/salt/utils/args.py", line 231, in _getargspec
>     raise ValueError('Function has keyword-only arguments or annotations'
> ValueError: Function has keyword-only arguments or annotations, use getfullargspec() API which can support them

Caused by assigning types to public salt methods.. 

```
       File "/usr/lib/python3.6/site-packages/salt/utils/args.py", line 271, in get_function_argspec                                                                                                                
          aspec = _getargspec(func)  # pylint: disable=redefined-variable-type                                                                                                                                       
        File "/usr/lib/python3.6/site-packages/salt/utils/args.py", line 231, in _getargspec                                                                                                                         
          raise ValueError('Function has keyword-only arguments or annotations'               
```

Should be fixed upstream. Some examples on how to do that https://github.com/pyinvoke/invoke/issues/357

jschmid1:
yay ^^^^ :ice_cream:

srv/salt/_modules/dg.py
```diff
@@ -0,0 +1,822 @@
...
+                if not _filter.matcher.compare(self._reduce_inventory(disk)):
+                    continue
+
+                if not self._has_mandatory_idents(disk):
+                    continue
+
+                if self._limit_reached(len(devices), disk.get('path')):
+                    continue
+
+                devices.add(disk.get("path"))
```
jan--f: When a device is added it should be removed from the list of available devices.

.gitignore
```diff
@@ -125,4 +125,5 @@ TAGS
 .mypy_cache/
 
 # projectile files
+.#*
```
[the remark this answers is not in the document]
jschmid1: eh, yes. That are emacs lock files.

srv/salt/_modules/dg.py
jschmid1: True. At least when talking about data_devices. This shouldn't apply for wal/db_devices.

I'll add that to my list and and push it with the next set of patches.

PR #1487, opened 2018-12-11
ceph-volume dummy: translate by-id device paths to short device paths

jan--f:
On real hardware we use by-id paths. Resolve those paths to short device names.


jschmid1:
Will this also match multipath devices? I have a vague memory of a request to also support those in the future.. @jan--f

jan--f:
@jschmid1 tbh no idea. But as this dummy implementation will vanish, does it make a difference?

PR #1488, opened 2018-12-12
cephfs: create pools with appropriate number of PGs

jan--f:
In nautilus the PG count does not include replication anymore. Adjust PG
counts for cephfs pools.


jschmid1:
Could you add the the changelog(or docs) as a reference? @jan--f

PR #1489, opened 2018-12-12
monitoring/dashboard: always refresh templates on time picker change.

jan--f:
Fixes: bcn#1113824

We had a wrong time picker in the last release (set to a fixed date in the past). Some template vars were set to refresh only on dashboard reload. This change refreshes template variables on time-picker change.


jschmid1:
Is this also something we need in master? @jan--f

jan--f:
@jschmid1 nope...the future of the dashboards on master is undecided. Currently it looks like we'll use the upstream dashboards from an RPM.

PR #1492, opened 2018-12-13
Move pytest to base_python 3.6 and adapt tests accordingly

jschmid1:
I think it's time to move our testing to python3 as we solely support python3 with the master-derived releases.

Changes due to:
  * [StandardError was removed - Use Exception](https://docs.python.org/3/whatsnew/3.0.html#changes-to-exceptions)
  * DeprecationWarning: invalid escape sequence \\*
  * [\_\_builtin\_\_ renamed to builtins](https://docs.python.org/3/whatsnew/3.0.html#library-changes)
  * [cmp was dropped](https://docs.python.org/3.0/whatsnew/3.0.html#ordering-comparisons)

Remark 1)

We still have to use a 'pyfakefs version < 3.3' as we rely on it's 'glob'. It's quite some work to move it to the new glob/fakefs api. Maybe after we reworked osd.py

Remark 2)

I skipped one test [test_prepare_device_file](https://github.com/SUSE/DeepSea/compare/pytest-py3?expand=1#diff-dc55f3e34c9ae82715049de455577825R350) which keeps failing due to a *supposedly* missing \_\_lt_\_\(\_\_gt\_\_) dunder method on the MagicMock side.

Backstory:

the test calls `RD.prepare_device_files` which sorts the output before returning. The `sorted` method, which is used for that leverages python3s magic \_\_lt\_\_ (gt, eq, etc) methods for internal comparisons between the objects. If these dunder methods are not defined python raises a 

`'TypeError: '<' not supported between instances of 'Object1' and 'Object2'`

The objects that are usually being compared bring those dunder methods. The test fails as the objects get replaced with the patched `MagicMock` objects. The problem is that these objects actually bring the requested methods:

``` shell
(Pdb) type(new_disk_device_files[0])
<class 'mock.mock.MagicMock'>

(Pdb) dir(new_disk_device_files[0])
['__gt__', '__lt__', 'assert_any_call', 'assert_called', 'assert_called_once', 'assert_called_once_with', 'assert_called_with', 'assert_has_calls', 'assert_not_called', 'attach_mock', 'call_args', 'call_args_list', 'call_count', 'called', 'configure_mock', 'method_calls', 'mock_add_spec', 'mock_calls', 'reset_mock', 'return_value', 'side_effect']
```
I'm not sure on how to solve that yet, so I'll keep that up for discussion. Since this test is a core part of the replacement unit tests, I'd like to get it back in.

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -960,7 +960,7 @@ tarball:
 	rm -r $(TEMPDIR)
 
 test: setup.py
-	tox -e py27
+	tox -e py36
 
 lint: setup.py
 	tox -e lint
--- a/srv/modules/runners/push.py
+++ b/srv/modules/runners/push.py
@@ -390,7 +390,7 @@ def _parse(line):
     Return globbed files constrained by optional slices or regexes.
     """
     if " " in line:
-        parts = re.split('\s+', line)
+        parts = re.split(r'\s+', line)
         files = sorted(glob.glob(parts[0]))
         for optional in parts[1:]:
             filter_type, value = optional.split('=')
--- a/tests/unit/_modules/test_cephdisks.py
+++ b/tests/unit/_modules/test_cephdisks.py
@@ -233,7 +233,7 @@ class TestHardwareDetections():
         """ Assume ret_bus_id returns True"""
         wm.return_value = '/valid/path'
         ret_bus_id.return_value = True
-        po.side_effect = StandardError
+        po.side_effect = Exception
         hwd.HardwareDetections()._query_disktype('sdn', {'controller_name': 'megaraid'}, 'base')
         assert ir.called is True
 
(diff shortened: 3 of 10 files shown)


jschmid1:
The underlying testing server apparently doesn't have python3.6 installed 

> 15:39:58 ERROR: InterpreterNotFound: python3.6

Could you fix that @kshtsk ?

kshtsk:
> The underlying testing server apparently doesn't have python3.6 installed
> 
> > 15:39:58 ERROR: InterpreterNotFound: python3.6
> 
> Could you fix that @kshtsk ?

@jschmid1 the compute node is leap 42.3 it has only python 3.4.6 available.

kshtsk:
retest this please

jschmid1:
@kshtsk fixed the it :raised_hands: 

Once we get that in, we have to rebase the open, master targeting PRs to make the tests pass again.

jschmid1:
does anyone have an idea regarding 'remark 2)' from the description

Issue #1495, opened 2018-12-18
Openstack integrate with prefix output has !!python-unicode in the output


mital-suse:
When running openstack integrate command with prefix, output has '!!python/unicode' 

salt-run --out=yaml openstack.integrate prefix=test
[WARNING ] Although 'dmidecode' was found in path, the current user cannot execute it. Grains output might not be accurate.
[WARNING ] Although 'dmidecode' was found in path, the current user cannot execute it. Grains output might not be accurate.
ceph_conf:
  cluster_network: 192.168.24.0/24
  fsid: 5d466551-ee84-388f-860e-a6b6cc080fd9
  mon_host: 192.168.24.200
  mon_initial_members: ses-cloud-qa1
  public_network: 192.168.24.0/24
cinder:
  key: AQAJOBhcAAAAABAAo5js5dLzr8L5kYtGsME8Rw==
  rbd_store_pool: !!python/unicode test-cloud-volumes
  rbd_store_user: !!python/unicode test-cinder
cinder-backup:
  key: AQAJOBhcAAAAABAAPSb0EiYm5FsX3FIBtdjCXA==
  rbd_store_pool: !!python/unicode test-cloud-backups
  rbd_store_user: !!python/unicode test-cinder-backup
glance:
  key: AQAJOBhcAAAAABAA+XBHE/4AQAgZhREE5an6pA==
  rbd_store_pool: !!python/unicode test-cloud-images
  rbd_store_user: !!python/unicode test-glance
nova:
  rbd_store_pool: !!python/unicode test-cloud-vms
radosgw_urls:
- http://ses-cloud-qa1:8080/swift/v1

`salt-run deepsea.version`
`rpm -qi salt-minion`
`rpm -qi salt-master`
)

mital-suse:
salt-run deepsea.version
[WARNING ] Although 'dmidecode' was found in path, the current user cannot execute it. Grains output might not be accurate.
0.8.6

rpm -qi salt-minion
Name        : salt-minion
Version     : 2016.11.4
Release     : 48.7.1
Architecture: x86_64
Install Date: Fri 19 Oct 2018 07:45:19 PM CEST
Group       : System/Management
Size        : 38128
License     : Apache-2.0
Signature   : RSA/SHA256, Wed 05 Sep 2018 01:17:04 PM CEST, Key ID 70af9e8139db7c82
Source RPM  : salt-2016.11.4-48.7.1.src.rpm
[...]

PR #1497, opened 2018-12-18
Fix possible unicode output error

hemna:
This patch adds a str() call around kwargs['prefix'] access to
eliminate the unwanted addition of "!!python/unicode" as part of the
prefix output.

Fixes #
https://github.com/SUSE/DeepSea/issues/1495

Description:

-----------------

**Checklist:**
- [ ] Adapted documentation


jschmid1:
@hemna I think Tim has a point here, would you mind trying the suggested fix?

Issue #1508, opened 2019-01-21
salt > 2018.x fails on ceph.updates.restart with 'Minion did not return. [No response]'


jschmid1:
The idea of 'ceph.updates.restart' _state_ is to dispatch a reboot command on a node to wait for it to come back in the orchestration. This has always worked since the returncode was either never sent or just not handled correctly. 

In newer version of salt this is not longer true. 

I tried:

- `(sleep 5 && reboot)&` -> waits for returncode and passes 'no response' up to the orchestration.

- `system.reboot` -> waits for returncode and passes 'no response' up to the orchestration.

- `reboot; exit 0`  -> waits for returncode and passes 'no response' up to the orchestration.

- `reboot || true`  -> waits for returncode and passes 'no response' up to the orchestration.

There are arguments to the cmd.run function that are supposed to suppress/ignore the returncode.

I tried:

- `check_cmd: /bin/true` (https://docs.saltstack.com/en/latest/ref/states/requisites.html#check-cmd)

- `ignore_retcode` (undocumented)

- `ignore_timeout` (undocumented) 

Apparently other people also had this issue and there seems to be a [rather ugly](https://github.com/saltstack/salt/issues/24297#issuecomment-274739089) solution.

-> `/usr/bin/nohup /bin/bash -c 'set -x && sleep 5 && systemctl reboot' >> /var/log/salt/minion 2>&1 &`

This is what we'll be using for now until something less hacky shows up I guess :/

PR #1509, opened 2019-01-21
Various fixes for the upgrade procedure

jschmid1:
* Do not execute a salt update in a orchestration
* Implement the ceph-disk->ceph-volume take-over
* Replace command we schedule reboots with

Due to changed behavior on the salt side we have to change the way we issue
reboots. See https://github.com/SUSE/DeepSea/issues/1508 for an explanation and
more information.

This will be the first branch to test the automated upgrade with.

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/salt/ceph/osd/takeover/default.sls
+++ b/srv/salt/ceph/osd/takeover/default.sls
@@ -1,0 +1,14 @@
+{% if 'storage' in salt['pillar.get']('roles') %}
+
+ invoking osd.takeover:
+  module.run:
+    - name: osd.takeover
+    - fire_event: True
+    - failhard: True
+
+ {% else %}
+
+ no osd detected:
+  test.nop
+
+ {% endif %}
--- a/srv/salt/ceph/updates/restart/default.sls
+++ b/srv/salt/ceph/updates/restart/default.sls
@@ -1,4 +1,3 @@
-
 {% set kernel = grains['kernelrelease'] | replace('-default', '')  %}
 {% set installed = salt['kernel.installed_kernel_version']() %}
 
@@ -9,11 +8,11 @@ warning:
     - installed: {{ installed }}
     - unless: "echo {{ installed }} | grep -q {{ kernel }}"
 
-reboot:
+rebootj:
   cmd.run:
-    - name: "shutdown -r now"
+    # This one is nasty! Please see (https://github.com/SUSE/DeepSea/issues/1508) for an explanation
+    - name: "/usr/bin/nohup /bin/bash -c 'set -x && systemctl reboot' >> /var/log/salt/minion 2>&1 &"
     - shell: /bin/bash
     - unless: "echo {{ installed }} | grep -q {{ kernel }}"
-    - failhard: True
+    - failhard: False
     - fire_event: True
-
--- a/.gitignore
+++ b/.gitignore
@@ -111,3 +111,18 @@ tags
 
 # backup
 *.bak
+
+# tag files
+GPATH
+GRTAGS
+GTAGS
+TAGS
+
+# org files
+*.org
+
+# mypy cache dirs
+.mypy_cache/
+
+# projectile files
+.projectile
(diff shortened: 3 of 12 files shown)


srv/salt/ceph/osd/takeover/default.sls
```diff
@@ -0,0 +1,16 @@
+{% if 'storage' in salt['pillar.get']('roles') %}
+
+{% for id in salt['osd.list']() %}
+taking over osd {{ id }}:
+  cmd.run:
+    - name: "ceph-volume simple scan /var/lib/ceph/osd/ceph-{{ id }}"
```
jschmid1: It appears that it's not enough to run this command..

c-v gives you a command to 'activate' after a successful 'simple scan'.

```
x@x: ceph-volume simple scan /dev/vdb1

Running command: /usr/sbin/cryptsetup status /dev/vdb1
--> OSD 1 got scanned and metadata persisted to file: /etc/ceph/osd/1-ccc15e6f-37bd-48dc-af22-70ba053bc6c3.json
--> To take over management of this scanned OSD, and disable ceph-disk and udev, run:
-->     ceph-volume simple activate 1 ccc15e6f-37bd-48dc-af22-70ba053bc6c3
```

Since this is currently running in a sls file we don't have a chance to pick this line up or compute the fsid..

We may have to switch to a python module for this.

jschmid1: A single command for this would be useful. Do we have enough time though?

jan--f: you can scan all OSDs and then run `ceph-volume simple activate --all`

jan--f: did you happen to check if the scan is idempotent? I seem to remember that it is, but best double check if we might need a conditional.

jschmid1: > did you happen to check if the scan is idempotent? I seem to remember that it is, but best double check if we might need a conditional.

it's not. but adding a `--force` works as the content doesn't change

jschmid1: > you can scan all OSDs and then run ceph-volume simple activate --all

neat! I must have overlooked that..

jan--f: > it's not. but adding a `--force` works as the content doesn't change

hmm or we check for the created json file with a `unless` prerequisite...

jschmid1: The thing is that we don't know the second param (osd_fsid) to construct the json file.

> A previously scanned OSD would exist like::

    /etc/ceph/osd/{ID}-{FSID}.json

if we could extract it easily, we also don't have to do the `activate --all`

the scan/activate wasn't designed with automation in mind as it seems

jschmid1: Oh, and at least in my version (

> admin:~ # ceph-volume simple activate --all
usage: ceph-volume simple activate [-h] [--file FILE] [--no-systemd]
                                   [ID] [FSID]
ceph-volume simple activate: error: unrecognized arguments: --all

srv/salt/ceph/updates/restart/default.sls
```diff
@@ -11,9 +10,9 @@ warning:
 
 reboot:
   cmd.run:
-    - name: "shutdown -r now"
+    # This one is nasty! Please see (https://github.com/SUSE/DeepSea/issues/1508) for an explanation
+    - name: "/usr/bin/nohup /bin/bash -c 'set -x && sleep 5 && systemctl reboot' >> /var/log/salt/minion 2>&1 &"
```
tserong: Well, that's horrible, but OK, I guess I have to respect the hack.

jschmid1:
package was successfully built but: `ssh: connect to host 10.86.0.120 port 22: Connection timed out
` blocked the success-status

PR #1524, opened 2019-02-05
Makefile: pass --long option to git describe

smithfarm:
Before this commit, we were generating the version number using
plain "git describe", which (according to the man page)

    "finds the most recent tag that is reachable from a commit. If the tag
    points to the commit, then only the tag is shown."

This had the undesirable effect of sometimes producing version numbers
that look like this:

    0.8.9+2.g14f2f94e

and other times like this:

    0.8.9

Depending on whether HEAD pointed to a tag or not. By passing the --long
option, we assure that the long format is used always.

Fixes #

Description:

-----------------

**Checklist:**
- [ ] Adapted documentation


kshtsk:
retest this please

PR #1530, opened 2019-02-12
force 'adler32' as CRC method

jschmid1:
salt 2019 complains if there is no such setting.

bonus:
got rid of superfluous whitespaces

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -891,6 +891,8 @@ copy-files:
 	# state files - salt
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/salt
 	install -m 644 srv/salt/ceph/salt/*.sls $(DESTDIR)/srv/salt/ceph/salt/
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/salt/crc
+	install -m 644 srv/salt/ceph/salt/crc/*.sls $(DESTDIR)/srv/salt/ceph/salt/crc/
 
 	# state files - orchestrate stage symlinks
 	ln -sf prep		$(DESTDIR)/srv/salt/ceph/stage/0
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ version:
 setup.py:
 	sed "s/DEVVERSION/"$(VERSION)"/" setup.py.in > setup.py
 
-pyc: setup.py 
+pyc: setup.py
 	#make sure to create bytecode with the correct version
 	find srv/ -name '*.py' -exec $(PYTHON) -m py_compile {} \;
 	find cli/ -name '*.py' -exec $(PYTHON) -m py_compile {} \;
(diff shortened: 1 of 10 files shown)


jschmid1:
I also find the warning rather weirdly texted which I claimed in bsc#1122790

Here's the answer from Bo:

```

It is not odd. Problem with Python3+ is that its hashes are salted and
are different between the sessions. This changed since Py3.5. In old
times, Py2 gives you the same hash for particular object everywhere:

$ python2 -c "print hash('foo')"
-4177197833195190597
$ python2 -c "print hash('foo')"
-4177197833195190597
$ python2 -c "print hash('foo')"
-4177197833195190597

...while Py3 will give you different hash for each session:

$ python3 -c "print(hash('foo'))"
8489007209051054531
$ python3 -c "print(hash('foo'))"
3545854823916522463
$ python3 -c "print(hash('foo'))"
-5687362106279541299

This breaks everything inside Salt. To overcome it, there are two ways
of keep it correct:

1. Set environment variable PYTHONHASHSEED to 0 (and let it badly
affect all cryptography-related libraries as well). But at least you
will get the same hash across the sessions and machines:

$ PYTHONHASHSEED=0 python3 -c "print(hash('foo'))"
-2457967226571033580
$ PYTHONHASHSEED=0 python3 -c "print(hash('foo'))"
-2457967226571033580
$ PYTHONHASHSEED=0 python3 -c "print(hash('foo'))"
-2457967226571033580

This is what Salt is using by default. To make sure it is isolated, it
fires another Python interpreter only to get hash for a string value.
For a complex instantiated objects it still not a solution.

2. Do not use hash() for what it was used in Salt, and use different
algorithm, like Adler32 (faster) or CRC32 (a little slower). This is
something that Salt should have used from the very beginning. But well,
they didn't.

This is what Salt is not using by default and asks you to migrate to.

So if you do not use Adler32 or CRC32, Salt will fire an external
Python3 interpreter with PYTHONHASHSEED=0 for *each* hashing function
and only just for that. The warning merely tells you that. So if you
are fine to update existing minions, you should just use either Adler32
(faster but less reliable) or CRC32 (slower but more reliable). To
migrate a minion, remove it from master, reconfigure to use CRC32 (I
would recommend it) and re-add it to the master.

> Is salt saying that the _current_ server_id is not computed with one of the
> new algorithms?


Yes.

> Also, if we don't set it (which we don't) the server_id should be generated
> with Adler32 by default? So.. that seems like unwanted behavior in salt.


Then every time it will fire an external Python to calculate it.
They've implemented it to fear that forcing Adler32 by default will
break things (server_id will be different). But I'd say, it is anyway
different between Python2 and Python3, so it is anyway broken.

```

jan--f:
_sigh_ well fair enough....thx for the info.

jschmid1:
> One minor nit: Maybe we should set that in its own config under `/etc/salt/minion.d/crc.conf` or some other name? Otherwise this might prevent the salt package from installing a new default config.

regarding this, I'm also not convinced to put it under ceph.salt or even how to apply it properly..

etc/salt/minion.d/ sounds more like a proper solution to me..

jschmid1:
@jan--f  well.. actually `/etc/salt/minion.d` gives us only a clean solution on the node where deepsea is installed..

for all minions we have to do a `file.managed` anyways.. That means we have two different ways we apply the file to nodes, which is maybe not ideal. 

I'd suggest we keep the current approach but I'll rename it to:

`ceph.salt.crc`

and add it to `stage.0`

tserong:
Nope, spoke too soon. I'm still getting "/usr/lib/python3.6/site-packages/salt/grains/core.py:2795: DeprecationWarning: This server_id is computed nor by Adler32 neither by CRC32. Please use "server_id_use_crc" option and define algorithm youprefer (default "Adler32"). The server_id will be computed withAdler32 by default." on the master, just not on the minions.

tserong:
Looks like this also needs to be set in `/etc/salt/master` or the master logs that error once per minute. And the master needs to be restarted to pick up the settings change, of course...

tserong:
Ugh...  And I'm *still* getting the warning when running stage 1! WTF am I missing here?

jschmid1:
mhh.. for me it was enough to add the crc line to the minion conf and restart them.

I'll need to double check though, maybe I've missed some messages :/

jschmid1:
@susebot run teuthology

jschmid1:
@susebot run teuthology

jschmid1:
@jan--f @tserong @swiftgist I update the approach. 

It's a state as part of stage.0 which makes sure to append the config to the minion conf.

PR #1533, opened 2019-02-13
add Prometheus and Grafana roles (SES-199)

jan--f:
and install and configure accordingly
[SES-199](https://jira.suse.de/browse/SES-199)


jschmid1:
yummy 6,5k lines deleted :sake:

capri1989:
Yeah we can configure Grafana and Prometheus individually. Thanks a lot!

jschmid1:
Since we have two new roles, we should also add them to `cephprocesses` and add a separate restart logic, right?

jan--f:
> Since we have two new roles, we should also add them to `cephprocesses` and add a separate restart logic, right?

Huh I though I had added them to cephprocesses already. Coming up. Restarts are trivial, things can just be restarted whenever.

srv/salt/ceph/stage/deploy/default.sls
```diff
@@ -1,5 +1,32 @@
+{% set master = salt['master.minion']() %}
+
 include:
   - .core
   - ...restart.mon.lax
   - ...restart.mgr.lax
   - ...restart.osd.lax
+
+enable prometheus module:
```
jschmid1: Maybe we should encapsulate these steps like we did with 'core'
```
.core
...restart.mon.lax
...restart.mon.lax
...restart.mon.lax
.monitoring_deploy
```

also, if we have the restart logic in place, we should add it similar to the ...restart.$role.lax approach

jan--f: hmm point taken. Unfortunately there is more to the monitoring deployment (stage 2 has the majority of it).

kshtsk:
retest this please

jan--f:
Closing in order to test JIRA integration

PR #1545, opened 2019-02-22
populate should overwrite the stack/default tree at all times

jan--f:

### DIFF
--- a/srv/modules/runners/populate.py
+++ b/srv/modules/runners/populate.py
@@ -92,11 +92,11 @@ class SaltWriter(object):
         else:
             self.overwrite = False
 
-    def write(self, filename, contents):
+    def write(self, filename, contents, overwrite=False):
         """
         Write a yaml file in the conventional way
         """
-        if self.overwrite or not os.path.isfile(filename):
+        if self.overwrite or not os.path.isfile(filename) or overwrite:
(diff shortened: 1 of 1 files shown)


jschmid1:
retest this please

jschmid1:
I think we really need to communicate this *before* people upgrade to the new version.

Would you mind doing a write-up and hand that over to Tomas?

jan--f:
@jschmid1 not sure what to communicate. This should be transparent to users. It only overwrites the default tree which is there specificially for our use and customers should have no business there anyway. Pillar data from `/srv/pillar/ceph/stack` will still overwrite the default tree.

jschmid1:
> or our use and customers should have no business there anyway

with an emphasis on the  _should_  :> 

I was under the impression that the global.yaml will also be overwritten, and if I'm not mistaken we reference that a couple of times in our documentation.

jan--f:
> I was under the impression that the global.yaml will also be overwritten, and if I'm not mistaken we reference that a couple of times in our documentation.

That is `/srv/pillar/ceph/stack/default/cluster/global.yml` (maybe silghtly different path but the default tree in any case)

jschmid1:
ah, right. I thought about the /srv/pillar/ceph/global.yaml. No objections then

PR #1548, opened 2019-02-27
monitoring: restarts, avoid "matched no target"

jan--f:

### DIFF
--- a/srv/salt/ceph/restart/grafana/default.sls
+++ b/srv/salt/ceph/restart/grafana/default.sls
@@ -1,6 +1,11 @@
+{% if (salt.saltutil.runner('select.minions', cluster='ceph', roles='grafana') != []) %}
 restarting grafana:
   salt.state:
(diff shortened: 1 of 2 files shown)


srv/salt/ceph/restart/grafana/default.sls
```diff
@@ -1,6 +1,8 @@
+{% if (salt.saltutil.runner('select.minions', cluster='ceph', roles='grafana') != []) %}
 restarting grafana:
   salt.state:
     - tgt: 'I@roles:grafana and I@cluster:ceph'
     - tgt_type: compound
     - sls: ceph.monitoring.grafana.restart
     - failhard: True
+{% endif %}
```
jschmid1: I _think_ this would result in a 'empty rendering' error when there is no role-grafana.

Adding a test.nop is probably needed

jschmid1:
@susebot run teuthology suse

Issue #1550, opened 2019-02-28
[master] DeepSea does not install ceph-mgr-{dashboard, etc.} RPMs


smithfarm:
### Description of Issue/Question

The latest Ceph 14.1.0 ships several `ceph-mgr-*` packages (notably `ceph-mgr-dashboard`, which is needed for the dashboard to work). These are "Recommended" by ceph-mgr.

DeepSea's Stage 3 uses `pkg.installed` to install the "ceph" metapackage, which is expected to bring in everything needed for Ceph server to run. However, this is not the case.

The `pkg.installed` State issues the following zypper command (taken from the zypper log):

```
'/usr/bin/zypper' '--non-interactive' '--no-refresh' 'install' '--auto-agree-with-licenses' '--name' 'ceph'
```

At first glance, one might assume that this will pull in the `ceph-mgr-dashboard` RPM via the Recommends, but it doesn't work this way. Only the Recommends of the package being installed are considered, not those of dependent packages. Since `ceph-mgr` is a dependency of `ceph`, this command does not install any Recommends of `ceph-mgr`.

I verified this on a virgin SLE-15-SP1 system:

```
target192168000059:/home/ubuntu # '/usr/bin/zypper' '--non-interactive' '--no-refresh' 'install' '--auto-agree-with-licenses' '--name' 'ceph'
Loading repository data...
Reading installed packages...
Resolving package dependencies...

The following 23 NEW packages are going to be installed:
  ceph ceph-base ceph-mds ceph-mgr ceph-mon ceph-osd cryptsetup liblvm2app2_2 liblvm2cmd2_02 libpwquality1 lvm2 psmisc
  python3-Mako python3-WebOb python3-WebTest python3-Werkzeug python3-bcrypt python3-beautifulsoup4 python3-logutils
  python3-pecan python3-pyOpenSSL python3-waitress which

23 new packages to install.
Overall download size: 40.4 MiB. Already cached: 0 B. After the operation, additional 159.3 MiB will be used.
Continue? [y/n/...? shows all options] (y): y
Retrieving package liblvm2app2_2-2.02.180-10.3.x86_64                               (1/23), 530.6 KiB (  1.4 MiB unpacked)
Retrieving: liblvm2app2_2-2.02.180-10.3.x86_64.rpm .................................................................[done]
Retrieving package liblvm2cmd2_02-2.02.180-10.3.x86_64                              (2/23), 702.6 KiB (  2.0 MiB unpacked)
[...]

PR #1551, opened 2019-02-28
ceph.packages: Install ceph-mgr-dashboard package

smithfarm:
Along with the rest of the ceph server packages in Stage 3.

Fixes: https://github.com/SUSE/DeepSea/issues/1550


jschmid1:
> This will install the package on all nodes. @swiftgist has a solution in #1558 that only installs it on mgr nodes.

sounds like a more fine grained solution to me.

PR #1555, opened 2019-03-01
Add dashboard orchestration

jschmid1:
I adapted the logic as follows:

* Username will default to 'admin' if not defined in the pillar (which takes precedence)
* Password will be generated with a salt internal function (`random.get_str(10)`) unless defined in the pillar (which takes precedence)

Since only `ceph mgr module enable dashboard` and `ceph dashboard create-self-signed-cert` are idempotent, there needs to be a check for the presence of the `dashboard_user` before actually creating the user.

After a user is created, the password will be stored in a grain structure looking like this:

``` yaml
dasboard_creds:
  username_1:
      password_1
  username_2:
      password_2
```

This structure will not allow to add multiple users at once (maybe we should?) but it will guarantee that switching to a new user, by changing the value in the pillar, will not overwrite the existing user's password.

If a user is present. There will be no action.

Due to how salt works, the password(respective) will be printed to the screen. This (may) be avoided using jinja conditionals, which I'd like to avoid..

Storing the passwords in the master's grain is maybe not the best technique security wise, but better than storing it in the pillar (which we do when specifying the password) as it doesn't leave the node.

The passwords can be queried with:

```
admin:~ # salt -I roles:master grains.get dashboard_creds
admin:
    ----------
    foo:
        ltfCTW33Ak
    bar:
        oVOhYBromI
    baz:
        askdJskaMas
```

Theoretically we can also implement a password changing mechanism controlled via the pillar with ` ceph dashboard ac-user-set-password <username> <password>`

Comments, notes, concerns?

----------------

### DIFF
--- a/srv/salt/ceph/dashboard/default.sls
+++ b/srv/salt/ceph/dashboard/default.sls
@@ -1,0 +1,32 @@
+{% set dashboard_user = salt['pillar.get']('dashboard_user', 'admin') %}
+{% set dashboard_pw = salt['pillar.get']('dashboard_password', salt['grains.get']('dashboard_creds:' ~ dashboard_user , salt['random.get_str'](10))) %}
+
+enable ceph dashboard:
+  cmd.run:
+    - name: ceph mgr module enable dashboard
+    - failhard: True
+
+create self signed certificate:
+  cmd.run:
+    - name: ceph dashboard create-self-signed-cert
+    - failhard: True
+
+dashboard user exists:
+  cmd.run:
+    - name: /bin/true
+    - unless: ceph dashboard ac-user-show -f json | jq -e 'index("{{ dashboard_user }}")'
+
+set username and password:
+  cmd.run:
+    # This command is printed although the 'onchange' statement evaluates as true. This might cause confusion.
+    - name: ceph dashboard ac-user-create {{ dashboard_user }} {{ dashboard_pw }} administrator
+    - onchanges:
+        - cmd: dashboard user exists
+
+set dashboard password grain:
+  module.run:
+    - name: grains.set
+    - key: dashboard_creds:{{ dashboard_user }}
+    - val: {{ dashboard_pw }}
+    - onchanges:
+        - cmd: set username and password
--- a/Makefile
+++ b/Makefile
@@ -310,6 +310,9 @@ copy-files:
 	# state files - macros
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/macros
 	install -m 644 srv/salt/ceph/macros/*.sls $(DESTDIR)/srv/salt/ceph/macros/
+	# state files - dashboard
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/dashboard
+	install -m 644 srv/salt/ceph/dashboard/*.sls $(DESTDIR)/srv/salt/ceph/dashboard/
 	# state files - mds
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/mds
 	install -m 644 srv/salt/ceph/mds/*.sls $(DESTDIR)/srv/salt/ceph/mds/
--- a/deepsea.spec.in
+++ b/deepsea.spec.in
@@ -116,6 +116,7 @@ systemctl try-restart salt-api > /dev/null 2>&1 || :
 %dir /srv/salt/ceph/cephfs/benchmarks
 %dir /srv/salt/ceph/cephfs/benchmarks/files
 %dir /srv/salt/ceph/metapackage
+%dir /srv/salt/ceph/dashboard
 %dir /srv/salt/ceph/salt-api
 %dir /srv/salt/ceph/salt-api/files
 %dir /srv/salt/ceph/configuration
@@ -467,6 +468,7 @@ systemctl try-restart salt-api > /dev/null 2>&1 || :
 %config /srv/salt/ceph/cephfs/benchmarks/*.sls
 %config /srv/salt/ceph/cephfs/benchmarks/files/keyring.j2
 %config /srv/salt/ceph/metapackage/*.sls
+%config /srv/salt/ceph/dashboard/*.sls
 %config /srv/salt/ceph/salt-api/*.sls
 %config /srv/salt/ceph/salt-api/files/*.conf*
 %config /srv/salt/ceph/tools/benchmarks/*.sls
--- a/srv/salt/ceph/dashboard/init.sls
+++ b/srv/salt/ceph/dashboard/init.sls
@@ -1,0 +1,2 @@
+include:
+  - .{{ salt['pillar.get']('dashboard_init', 'default') }}
--- a/srv/salt/ceph/stage/deploy/core/default.sls
+++ b/srv/salt/ceph/stage/deploy/core/default.sls
@@ -119,6 +119,12 @@ wait for mgr to be available:
     - tgt_type: compound
     - kwarg:
         'cmd': 'test "$(ceph mgr dump | jq .available)" = "true"'
+
+dashboard:
+  salt.state:
+    - tgt: {{ master }}
+    - tgt_type: compound
+    - sls: ceph.dashboard
     - failhard: True
 
 osd auth:


jschmid1:
@susebot run teuthology

jschmid1:
Error ENOENT: all mgr daemons do not support module 'dashboard', pass --force to force enablement

smithfarm:
Maybe bring #1551 into this PR, or merge it first?

smithfarm:
(I don't know for sure, but I assume the reason for the failure is the ceph-mgr-dashboard RPM is not installed)

srv/salt/ceph/dashboard/default.sls
```diff
@@ -0,0 +1,15 @@
...
+
+create self signed certificate:
+  cmd.run:
+    - name: ceph dashboard create-self-signed-cert
+    - failhard: True
+
+set username and password:
+  cmd.run:
+    # This command is printed although the 'unless' statement evaluates as true. This might cause confusion.
+    - name: ceph dashboard ac-user-create {{ salt['pillar.get']('dashboard_username', 'admin') }} {{ salt['pillar.get']('dashboard_password', salt['random.get_str'](10)) }} administrator || true
```
smithfarm: We might want to consult with @LenzGr here, since users will want to access the dashboard and this (randomly generated string as password) might make it less-than-trivial to do so?

smithfarm: Nobody is asking me, but if they did, I'd advocate for having a hard-coded default password, like "admin", which is documented along with a big fat box advising folks to change that password as soon as possible after the cluster is deployed.

If someone wants security, they can set the password via the pillar, although not sure how secure that is?

jan--f: I'd be ok with a random password but then we'd need to persist is somewhere so it can be looked up. Ideally this would be handled by the dashboard itself: No requirement to add a password and on first login the user is required to add password (but I guess this is a feature request).

jschmid1: We had this conversation on rocketchat #ceph (Thursday evening ~6pm).

I should've written the conclusion in the PR body, my bad.

* random generated password are to be favoured over a hard-coded default password.
* the random password needs to be printed to the screen (this is a todo, I'll just add a advise runner)
* the dashboard will get a 'change-password-after-first-login' policy
* you should be allowed to set the password in the pillar (implemented)

jschmid1:
We need #1551 for this to work.

srv/salt/ceph/dashboard/default.sls
smithfarm: Hm, then I guess we are going to need documentation on how to reset the dashboard password, since I'm having trouble imagining a 100% success rate on users (a) noticing the random dashboard password printed on the screen and (b) remembering it for long enough that they can use it to log in.

But maybe I'm getting pessimistic in my old age.

Presumably, purging the dashboard role and then redeploying it will reset the password?

jschmid1: CC'ing all parties involved @LenzGr @l-mb

smithfarm: This just seems like we're setting ourselves up for user disappointment. It may seem trivial to us to (a) notice a randomly generated password on the screen, before it scrolls off the top edge, and (b) be capable of entering that password into the dashboard login page, but the real "user experience" might be quite different (and disappointing).

l-mb: Setting up a default password for an admin account on a product is too risky.

That needs to always be user set or unique to a particular install.

jschmid1:
I updated the body of this pullrequest explaining my changes. This is more or less a compromise for all remarks in the review.

l-mb:
I'm OK with this, but I think I'd prefer having it in the pillar (I may be wrong in my understanding of Salt, but I thought that's where credentials are supposed to be).

Since that's where it'd be defined explicitly anyway, can maybe stage 0 (or 2?) notice if no password/user is configured and generate one for admin and put it into the place where it'd normally be in the pillar? So the place where admins would either look up the one they've configured as the initial one or the one we've created would be identical?

(And possibly have "reset" runner in case an admin needs to overwrite what their dashboard users have done, but that's a tangent ...)

Does this question make sense at all?

srv/salt/ceph/dashboard/default.sls
```diff
@@ -0,0 +1,32 @@
+{% set dashboard_user = salt['pillar.get']('dashboard_user', 'admin') %}
+{% set dashboard_pw = salt['pillar.get']('dashboard_password', salt['grains.get']('dashboard_creds:' ~ dashboard_user , salt['random.get_str'](10))) %}
```
jan--f: Why the grains lookup here? Iiuc having a password in the grains means we've created the user already.

jschmid1: There needs to be a password for the previously specified user though. The password being present indicates that this user is already registered.

jschmid1:
> I'm OK with this, but I think I'd prefer having it in the pillar (I may be wrong in my understanding of Salt, but I thought that's where credentials are supposed to be).

I agree that it's weird to have the username/pw in two different places, but one is passive while the other is active. We probably could feed the user/pw in the pillar with weird file manipulation but I doubt that this is make this process any cleaner..

> Since that's where it'd be defined explicitly anyway, can maybe stage 0 (or 2?) notice if no password/user is configured and generate one for admin and put it into the place where it'd normally be in the pillar? So the place where admins would either look up the one they've configured as the initial one or the one we've created would be identical?

Maybe it makes sense to keep the context in the pillar only. I'll have to check if that's feasible at all..

jschmid1:
> We should also prominently document that a user in the Administrator role will be created.

Right, that should be in *bold* somewhere in the wiki/docs

srv/salt/ceph/dashboard/default.sls
jan--f: > There needs to be a password for the previously specified user though.

I don't follow why.

> The password being present indicates that this user is already registered.

Right and that means that both `set username and password` and `set dashboard password grain` won't be executed...?

jschmid1: If you don't do the grain lookup here, a default password will be printed to the screen even though you already chose a password.

```
admin:~ # cat /srv/pillar/ceph/stack/global.yml
dashboard_user: test3
dashboard_password: test3


admin:~ # salt \admin* state.apply ceph.dashboard
admin:
  Name: ceph mgr module enable dashboard - Function: cmd.run - Result: Changed Started: - 09:38:38.273895 Duration: 376.823 ms                                                                
  Name: ceph dashboard create-self-signed-cert - Function: cmd.run - Result: Changed Started: - 09:38:38.650988 Duration: 282.95 ms                                                           
  Name: /bin/true - Function: cmd.run - Result: Clean Started: - 09:38:38.934131 Duration: 252.67 ms                                                                                          
  Name: ceph dashboard ac-user-create test3 test3 administrator - Function: cmd.run - Result: Clean Started: - 09:38:39.187332 Duration: 0.003 ms                                             
  Name: grains.set - Function: module.run - Result: Clean Started: - 09:38:39.188225 Duration: 0.004 ms  
```

if you now change the password in the pillar (or delete it), the grain lookup will still print to correct password to the screen. Removing the grain lookup will generate a new one although it will not be saved in the grains.


Writing this makes me wonder if upon password removal in the pillar, we should react to that.
Overwriting it is possbile with the available `ceph dashboard user foo ` calls..

jschmid1: > Writing this makes me wonder if upon password removal in the pillar, we should react to that.
> Overwriting it is possbile with the available `ceph dashboard user foo ` calls..

Well.. That'd probably be even more confusing when a user changes the password in the web-ui and the password will be reset on the next stage.x invocation. This should probably be a one-shot operation

jschmid1:
> Maybe it makes sense to keep the context in the pillar only. I'll have to check if that's feasible at all..

The main downside of this is that users are allowed to use many different files to manipulate the pillar, and from a sls file we can't just edit 'the pillar' as it's controlled by files living on the master.

Unless we define a policy that only allows the dashboard_user/pw to be set in a certain file, we might be better off with using grains as a store.

jschmid1:
With that said, it probably doesn't get any better with the limitations we have. Grains are the safest bet..

PR #1556, opened 2019-03-01
[SES5] monitoring: update smartmon.sh, fix label dimension inconsistency

jan--f:

### DIFF
--- a/srv/salt/ceph/monitoring/prometheus/exporters/files/smartmon.sh
+++ b/srv/salt/ceph/monitoring/prometheus/exporters/files/smartmon.sh
@@ -6,8 +6,11 @@
 # TODO: This probably needs to be a little more complex.  The raw numbers can have more
 #       data in them than you'd think.
 #       http://arstechnica.com/civis/viewtopic.php?p=22062211
-me=`basename "$0"`
 
+# Formatting done via shfmt -i 2
+# https://github.com/mvdan/sh
+me=`basename "$0"`
+ 
 function clean_up {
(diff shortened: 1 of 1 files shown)


jan--f:
Seems there's a potential issue, waiting on feedback.

jschmid1:
Any reference to that change?

jschmid1:
retest this please

tserong:
It's difficult for me to tell how much of this is reformatting and how much is fixing something (I don't think I actually looked at this script before :-/)

jan--f:
The issue is that the old script potentially causes this log message: `has label dimensions inconsistent with previously collected metrics in the same metric family`. The initial fix didn't deal well with nvme drives (now fixed).
Still waiting on feedback, but I believe this should fix the issue.

Code-wise this is the script from the node_exporter repo plus the fix for nvme drives. Will upstream the nvme fix.

jan--f:
retest this please

jan--f:
Got confirmation from a customer that this fixes their issue. I pushed the fix upstream as well https://github.com/prometheus/node_exporter/pull/1313

jan--f:
> One question: where'd the `clean_up` stuff come from (lines 12-25)? I can't seem to see that in the version in the node_exporter repo, although it certainly seems sensible to have that piece IMO. Maybe that should be added upstream too?

Yeah I added locking via a mkdir and the relevant cleanup. Not really sure if thats necessary, but assume a long running invocation or multiple invocations this should avoid the output file to become corrupted. Will try to upstream it.

PR #1559, opened 2019-03-05
monitoring: defer deployment to stage 3

jan--f:
Otherwise role assignments are not accessible since jinja is evaluated
before anything is run, including push.proposals and pillar sync (and thus
role-based switches get evaluated before roles are in the pillar).


jan--f:
I was hoping we can deploy monitoring in stage 2 but I can't think of a reliable way to get the current role assignments in jinja during stage 2.

jschmid1:
@susebot run teuthology

PR #1543, opened 2019-02-21
stage 3: configure deepsea mgr orchestrator module

tserong:
This enables the deepsea mgr orchestrator module, sets it as the
current orchestrator backend, and configures it to talk to the
salt-api service.

Signed-off-by: Tim Serong <tserong@suse.com>

-----------------

**Checklist:**
- [x] Added unittests and or functional tests
- [ ] Adapted documentation
- [ ] Referenced issues or internal bugtracker
- [ ] Ran integration tests successfully (trigger with "@susebot run teuthology" in a GitHub comment; see the [wiki](https://github.com/SUSE/DeepSea/wiki/Testing) for more information)

PR #1563, opened 2019-03-08
stage 3: wait for mgr to be available after it's initially deployed

tserong:
Immediately after deploying ceph-mgr, it takes a few seconds for the
various modules to become available.  If we don't wait for this, any
subqeuent `ceph` commands that require mgr will fail (for example
`ceph mgr module enable [...]`).

-----------------

**Checklist:**
- [ ] Adapted documentation


tserong:
If we merge this, then #1555 needn't wait for #1543 to be merged. It's cleaner putting the wait where I've got it here now anyway.

tserong:
@susebot run teuthology

PR #1517, opened 2019-01-30
Gradually replace ceph-disk/proposal dependent operations with c-v/drivegroup equivalents

jschmid1:
Update 2019-03-06:

This patch re-implements:

* removal

Got slightly easier due to some convenience methods provided by ceph.
(purge osd, ceph-volume zap) 

* replacement

This is basically the same as the `removal` with the difference of using `ceph osd destroy` instead of `ceph osd purge`, which leaves the OSD in the crushmap and marked 'destroyed'.

* (re)deployment

`disks.deploy` will no compute a map of destroyed osds from `ceph osd tree destroyed` and passes it down to the minion module `dg.py`. This modules reads the map and decides, based on hostnames, if a deployment becomes a 're'-deployement. The difference here is to add a `--osd-ids x x x ` to the ceph-volume call.

* OSD encryption

by adding the `encyrption` flag in the drive_groups.yml (global section). to my knowledge it's not possible to individually encrypt wal/db/data devices.

also Fixes #1557

~working on the unit tests~
done

PR #1567, opened 2019-03-11
fix Partition GUID code in guuid_table

iwakuS:
This patch exchanges partition GUID code in guuid_table.
Partition GUID codes of 'data' and 'journal' are opposite.
There is no problem because key of guuid_table is not used, although.

Signed-off-by: Shingo Iwakura. iwakura.shingo@jp.fujitsu.com

Fixes #

Description:

-----------------

**Checklist:**
- [ ] Adapted documentation


jschmid1:
Thanks for the contribution @iwakuS but with #1517 the `cephdisks` module is going away. The GUID handling, if needed, will be handled by ceph-volume.

Issue #536, opened 2017-08-11
DeepSea creates pools with hardcoded number of PGs, causing non-trivial clusters to come up in HEALTH_WARN due to insufficient PGs per OSD


smithfarm:
As we have discussed before, DeepSea wants to be helpful and create pools for the user, so the user has something working from the get-go. This has been tested on small clusters, and seems to be useful.

The problem is that, when creating a pool, DeepSea must specify a number of PGs.

When a human user creates a pool, they are advised to calculate the number of PGs, taking into account their cluster topology, CRUSH map, and anticipated usage patterns. Contrast this with DeepSea, which does no calculation at all and simply hardcodes the number of PGs.

Since the number of PGs is hardcoded, it is trivial to make DeepSea bring up the cluster in HEALTH_WARN (due to insufficient PGs per OSD) simply by adding OSDs.

This is reproducible in all cases where DeepSea is currently creating pools on behalf of the user.

There are several possibilities for addressing this:

1. fudge by increasing the hardcoded number of PGs, in the hopes that fewer users will get bitten
2. document that the cluster might come up in HEALTH_WARN, so users know they need to re-do the pool creation
3. do some kind of calculation instead of hard-coding the number of PGs, but this means DS has to keep track of which pools are already created since all the pools will presumably be created over all the OSDs
4. leave it to the user to create pools

Options 1 and 2 strike me as unprofessional, and the calculation in 3 seems like it's bound to come up with the wrong value. If we did go with 3, we still need to document which pools DS creates and explain to users how to fix things if the number of PGs guessed by DeepSea turns out to be wrong for their cluster.

As for option "4", well . . . Ceph was auto-creating a pool and stopped doing that, presumably for the same reasons. Also, I heard that ceph-ansible does not create pools, either.

It's as if DS is assuming the user is a Ceph neophyte who knows very little about Ceph and simply wants to push buttons. The current behavior works for this type of user because they're unlikely to have more than a few OSDs. However, it risks annoying users who have knowledge of Ceph, because they will have to re-do the pools when DS is finished, but then if they ever re-run Stage 4 the pools will get re-created . . . (?)

smithfarm:
Users for whom the hard-coded PG values are not right could pre-create the pools (which DS would otherwise create) before running Stage 4. So there is a sane option for dealing with this issue.

[...]

PR #1573, opened 2019-03-14
iscsi: fix deployment of iSCSI gateways 

rjfd:
Fixes the deployment of iSCSI gateways based on ceph-iscsi:
* automatically finds a pool or creates a new one to store ceph-iscsi configuration
* generates `iscsi-gateway.cfg` with correct list of trusted IPs
* automatically configures ceph-dashboard to enable iSCSI managment


rjfd:
labeling this DNM until we enable the iSCSI some test in teuthology.

cc @smithfarm

rjfd:
@jschmid1 in this PR I'm creating a pool called `.configs` to store the ceph-iscsi configuration. The reason why I called it `.configs` is because NFS-Ganesha will also need to store configuration files in a RADOS pool, and I think we can have a single pool for all services that need to store configurations in a RADOS pool.

What do you think?

swiftgist:
@rjfd If you delete the init.sls, then how do you propose that a site would customize the iscsi configuration if needed?  Since the targets of the two state files are different (one is for master and one is for igw), these are normally in separate directories.  Retaining at least

# salt -I roles:igw state.apply ceph.igw.config

would keep the command parallel to the other components.

Martin-Weiss:
Do we really need a separate Pool for a few bytes of configuration files? Can’t we use an existing one the customer
defines for these few bytes?

I see the challenge with the number of PGs in case we have more pools..

With lrbd we also stored the config in the pools that host the RBDs...

Martin

> Am 14.03.2019 um 16:44 schrieb Ricardo Dias <notifications@github.com>:
> 
> @jschmid1 in this PR I'm creating a pool called .configs to store the ceph-iscsi configuration. The reason why I
called it .configs is because NFS-Ganesha will also need to store configuration files in a RADOS pool, and I think we
can have a single pool for all services that need to store configurations in a RADOS pool.

rjfd:
> @rjfd If you delete the init.sls, then how do you propose that a site would customize the iscsi configuration if needed? 

@swiftgist I can change this to match the structure of other parts of DeepSea as you are suggesting but I wonder what exactly needs to be customized and what is the real benefit?

If we are talking about using different values for the options declared in `iscsi-gateway.cfg.j2`, wouldn't make more sense to allow the user to override those values in the salt pillar instead of copy&paste several files, which we will never control in an upgrade scenario?

rjfd:
> Do we really need a separate Pool for a few bytes of configuration files? Can’t we use an existing one the customer defines for these few bytes?

@Martin-Weiss there a few problems with that approach, which I currently know how to solve:
* In a fresh deployment there is no "existing pool created by the user"
* If DeepSea creates a pool for storing the RBD images that will be used in iSCSI targets, how many PGs should be set for that pool?

Besides the above, there will be more services, like nfs-ganesha, that will need to store some RADOS objects with configurations. In what pool should we store nfs-ganesha configuration objects?

What I'm proposing here is to have a dedicated pool to store services configurations. This has the benefit of fully separating the services configuration from the data pools, which will make upgrades a bit safer as we do not need to make changes to objects in pools that also contain user data.

smithfarm:
Pool creation by DeepSea is a real problem to the extent that it uses a hard-coded number of PGs - see https://github.com/SUSE/DeepSea/issues/536 where I already complained about this. 

In luminous, upstream implemented PG overdose (and "underdose") protection, which causes the cluster to very probably come up with health warnings if the number of PGs in the pool(s) is not tailored to the number of OSDs in the cluster. Now, there is also a "hard limit" which can be enabled which will cause the pool creation to simply fail if it would result in an overdose.

I do know that rgw autonomously creates a number of pools when it is first deployed, all with a small number of PGs. These pools are not created by DeepSea, but by rgw itself.

In the CI tests, I get around these problems by pre-creating all pools that DeepSea would otherwise create in Stage 4, with a number of PGs calculated to avoid the warnings.

Martin-Weiss:
>>  Do we really need a separate Pool for a few bytes of configuration files? 
> Can’t we use an existing one the customer defines for these few bytes?
> 
> @Martin-Weiss there a few problems with that approach, which I currently 
> know how to solve:
> * In a fresh deployment there is no "existing pool created by the user"

No pool = no RBD = no iSCSI config required ;-)
The current approach with a default iscsi-images and demo RBD is not bad IMO.

> * If DeepSea creates a pool for storing the RBD images that will be used in 
> iSCSI targets, how many PGs should be set for that pool?

I believe the lowest possible number so that admins can increase it later - so maybe 8?

> 
> Besides the above, there will be more services, like nfs-ganesha, that will 
> need to store some RADOS objects with configurations. In what pool should we 
> store nfs-ganesha configuration objects?

Either NFS exports S3/RGW -> then in one of the RGW pools
Or NFS exports cephFS -> then in the cephfs metadata or data pool

> 
> What I'm proposing here is to have a dedicated pool to store services 
> configurations. This has the benefit of fully separating the services 
> configuration from the data pools, which will make upgrades a bit safer as we 
> do not need to make changes to objects in pools that also contain user data.

It would be great if we could have a sort of config map on the MONs for such things..

Martin

smithfarm:
> I believe the lowest possible number so that admins can increase it later - so maybe 8?

The problem with this is, when a pool with a very low number of PGs is created, the cluster goes into HEALTH_WARN with "too few PGs per OSD" warning.

Somehow, the RGW pools are apparently excluded from this - i.e. they do not trigger the warning. That's why I was wondering if @rjfd 's .config pool could also get such a waiver from upstream.

rjfd:
> > Do we really need a separate Pool for a few bytes of configuration files? Can’t we use an existing one the customer defines for these few bytes? @Martin-Weiss there a few problems with that approach, which I currently know how to solve: * In a fresh deployment there is no "existing pool created by the user"
> No pool = no RBD = no iSCSI config required ;-) The current approach with a default iscsi-images and demo RBD is not bad IMO.

That does not make sense IMO. DeepSea should deploy the iSCSI gateways if `igw-role` is declared in policy.cfg independently if there is a Ceph pool or not.

Creating a demo pool like we did with `iscsi-images` makes sense but it's a pool that can be deleted by the user (as it's a demo).
I would like to store services configuration in a more permanent/persistent location.

> * If DeepSea creates a pool for storing the RBD images that will be used in iSCSI targets, how many PGs should be set for that pool?
> I believe the lowest possible number so that admins can increase it later - so maybe 8?

Changing the number of PGs requires rebalancing, is it worth the pain of data migration because of this?

> Besides the above, there will be more services, like nfs-ganesha, that will need to store some RADOS objects with configurations. In what pool should we store nfs-ganesha configuration objects?
> Either NFS exports S3/RGW -> then in one of the RGW pools Or NFS exports cephFS -> then in the cephfs metadata or data pool

What about when we want exports of both CephFS and RGW types, which pool should we choose?

> What I'm proposing here is to have a dedicated pool to store services configurations. This has the benefit of fully separating the services configuration from the data pools, which will make upgrades a bit safer as we do not need to make changes to objects in pools that also contain user data.
> It would be great if we could have a sort of config map on the MONs for such things..

That's not going to happen. Each service has it's own configuration syntax and structure, also these services use features of RADOS objects, like watch/notify, that would not be available in the mons KV store.

Martin-Weiss:
>>  I believe the lowest possible number so that admins can increase it later - 
> so maybe 8?
> 
> The problem with this is, when a pool with a very low number of PGs is 
> created, the cluster goes into HEALTH_WARN with "too few PGs per OSD" 
> warning.
> 

We only see "too few PGs per OSD" in case this is the only pool which should not be the case in a production setup.
The only way to work around this would be to have a proper calculation for PGs - and that can only be done in case we would know about all required pools.

Martin-Weiss:
>> > Do we really need a separate Pool for a few bytes of configuration files? 
> Can’t we use an existing one the customer defines for these few bytes? 
> @Martin-Weiss there a few problems with that approach, which I currently know 
> how to solve: * In a fresh deployment there is no "existing pool created by 
> the user"
>> No pool = no RBD = no iSCSI config required ;-) The current approach with a 
> default iscsi-images and demo RBD is not bad IMO.
> 
> That does not make sense IMO. DeepSea should deploy the iSCSI gateways if 
> `igw-role` is declared in policy.cfg independently if there is a Ceph pool or 
> not.

From a deepsea point of view I agree - but deepsea is just the screw driver and not the car ;-).
In case I deploy an iSCSI gateway this is 100% useless without a pool for the RBDs.
So what configuration is required to be deployed for the iscsi gateway if no pool and no RBD is there?

> 
> Creating a demo pool like we did with `iscsi-images` makes sense but it's a 
> pool that can be deleted by the user (as it's a demo).
> I would like to store services configuration in a more permanent/persistent 
> location.

Agreed - and the way how lrbd did it was not too bad (store in one object in each pool what is required for the iscsi
config of the RBDs exported out of that pool.

> 
>> * If DeepSea creates a pool for storing the RBD images that will be used in 
> iSCSI targets, how many PGs should be set for that pool?
>> I believe the lowest possible number so that admins can increase it later - 
> so maybe 8?
> 
> Changing the number of PGs requires rebalancing, is it worth the pain of 
> data migration because of this?

Rebalancing is only a challenge in case we have data in the pool - so using a small number for a pool with a few
"bytes" should not cause too many headaches I assume.

> 
>> Besides the above, there will be more services, like nfs-ganesha, that will 
> need to store some RADOS objects with configurations. In what pool should we 
> store nfs-ganesha configuration objects?
>> Either NFS exports S3/RGW -> then in one of the RGW pools Or NFS exports 
> cephFS -> then in the cephfs metadata or data pool
> 
> What about when we want exports of both CephFS and RGW types, which pool 
> should we choose?

Similar to lrbd - store in each pool what is required for the exports out of this pool.

> 
>> What I'm proposing here is to have a dedicated pool to store services 
> configurations. This has the benefit of fully separating the services 
> configuration from the data pools, which will make upgrades a bit safer as we 
> do not need to make changes to objects in pools that also contain user data.

From a deepsea point of view I like this - a simliar requiremen is there for the SBD in case we setup the SMB or NFS
"HA" based gateway.
I am just not sure if we really want to waste the available number of PGs with a small (from a capacity point if view)
config-files pool.

>> It would be great if we could have a sort of config map on the MONs for such 
> things..
> 
> That's not going to happen. Each service has it's own configuration syntax 
> and structure, also these services use features of RADOS objects, like 
> watch/notify, that would not be available in the mons KV store.
> 

We also could use things like corosync and rsync that would not need to store the config files in ceph at all.
Basically storing shared config in ceph only makes sense in case multiple daemons can share single config files and
understand the changes, there - so the daemons using the files in ceph need to be adjusted for that anyway - so maybe a
standard key-value store approach for these things would be the better way?

rjfd:
> From a deepsea point of view I agree - but deepsea is just the screw driver and not the car ;-). In case I deploy an iSCSI gateway this is 100% useless without a pool for the RBDs. So what configuration is required to be deployed for the iscsi gateway if no pool and no RBD is there?

I don't understand why you need DeepSea to create a pool. You already have a simple way to do it in the CLI: `ceph osd pool create rbd 100`, or in Ceph Nautilus you can use Ceph-Dashboard WebUI to create and manage pools.

> > What about when we want exports of both CephFS and RGW types, which pool should we choose?
> Similar to lrbd - store in each pool what is required for the exports out of this pool.
> We also could use things like corosync and rsync that would not need to store the config files in ceph at all. Basically storing shared config in ceph only makes sense in case multiple daemons can share single config files and understand the changes, there - so the daemons using the files in ceph need to be adjusted for that anyway - so maybe a standard key-value store approach for these things would be the better way?

We can't change the way these services (ceph-iscsi, nfs-ganesha, etc...) are storing their configuration files. It is DeepSea that needs to adapt to make sure that it deploys correctly those services.

Martin-Weiss:
>>  From a deepsea point of view I agree - but deepsea is just the screw driver 
> and not the car ;-). In case I deploy an iSCSI gateway this is 100% useless 
> without a pool for the RBDs. So what configuration is required to be deployed 
> for the iscsi gateway if no pool and no RBD is there?
> 
> I don't understand why you need DeepSea to create a pool. You already have a 
> simple way to do it in the CLI: `ceph osd pool create rbd 100`, or in Ceph 
> Nautilus you can use Ceph-Dashboard WebUI to create and manage pools.

True - I do not need deepsea to create a pool. And I do not even want deepsea to create a config pool for iscsi or nfs config things ;-).
So in case I use deepsea to deploy the iscsi gateway deepsea might not even need to configure anything for iSCSI as we might want to do that with the UI..

> 
>> > What about when we want exports of both CephFS and RGW types, which pool 
> should we choose?
>> Similar to lrbd - store in each pool what is required for the exports out of 
> this pool.
>> We also could use things like corosync and rsync that would not need to 
> store the config files in ceph at all. Basically storing shared config in 
> ceph only makes sense in case multiple daemons can share single config files 
> and understand the changes, there - so the daemons using the files in ceph 
> need to be adjusted for that anyway - so maybe a standard key-value store 
> approach for these things would be the better way?
> 
> We can't change the way these services (ceph-iscsi, nfs-ganesha, etc...) are 
> storing their configuration files. It is DeepSea that needs to adapt to make 
> sure that it deploys correctly those services.

Ok - then back to the question - why to we need a ceph config files pool at all?
Ganesha uses /etc/... files for config.. the only service I am aware of is lrbd that is using a config stored in ceph pools.

Martin

rjfd:
> Ok - then back to the question - why to we need a ceph config files pool at all? Ganesha uses /etc/... files for config.. the only service I am aware of is lrbd that is using a config stored in ceph pools.

This is not true anymore. In order to manage iSCSI and NFS-Ganesha gateways from the Ceph-Dashboard, the configuration of these services need to be stored in the Ceph cluster.

For iSCSI, ceph-iscsi stores a single RADOS object with the configuration that is shared across all iSCSI gateways.

For NFS-Ganesha, the structure is a bit more complex. Each gateway daemon must have a corresponding RADOS object with its configuration and each NFS export configuration will be stored in a RADOS object that is shared across all NFS-Ganesha daemons. More details here: http://docs.ceph.com/docs/nautilus/mgr/dashboard/#nfs-ganesha-management

Martin-Weiss:
>>  Ok - then back to the question - why to we need a ceph config files pool at 
> all? Ganesha uses /etc/... files for config.. the only service I am aware of 
> is lrbd that is using a config stored in ceph pools.
> 
> This is not true anymore. In order to manage iSCSI and NFS-Ganesha gateways 
> from the Ceph-Dashboard, the configuration of these services need to be 
> stored in the Ceph cluster.
> 
> For iSCSI, ceph-iscsi stores a single RADOS object with the configuration 
> that is shared across all iSCSI gateways.

Oh - and this is not as "intelligent" as the one we have with lrbd?
If there is only one pool with RBDs I would take that one.. if there are multiple pools available - customer should specify which one should be used.

> 
> For NFS-Ganesha, the structure is a bit more complex. Each gateway daemon 
> must have a corresponding RADOS object with its configuration and each NFS 
> export configuration will be stored in a RADOS object that is shared across 
> all NFS-Ganesha daemons. More details here: 
> http://docs.ceph.com/docs/nautilus/mgr/dashboard/#nfs-ganesha-management 
> 

Hm here I would put the object in the pool that has the data that we want to export.

As said - a single config pool for just 5 or maybe 10 objects seems to be some overkill from my point of view. 
(or we add one that just has 8 PGs and we ignore or can disable any warnings regarding this pool..)

Basically I believe this also needs to be discussed upstream..

jecluis:
Well, can anyone explain in a concise manner what is the hold up here?

Having a pool to store metadata (or config, whatever) makes sense to me. What's the big deal about this? Especially if the objective is to (if I understand correctly) having either deepsea or some other external tool (e.g., the dashboard) creating that pool?

jecluis:
Also, 
> Hm here I would put the object in the pool that has the data that we want to export. As said - a single config pool for just 5 or maybe 10 objects seems to be some overkill from my point of view. (or we add one that just has 8 PGs and we ignore or can disable any warnings regarding this pool..) Basically I believe this also needs to be discussed upstream..

So we're on the same page, that warming will only be triggered if the number of PGs across *all osds* is below a certain threshold. A pool with a few objects, or PGs has nothing to do with that.

Martin-Weiss:
> Am 15.03.2019 um 19:41 schrieb Joao Eduardo Luis <notifications@github.com>:
> 
> Well, can anyone explain in a concise manner what is the hold up here?
> 

Do we want to have a pool for two to five objects only?

How many PGs do you want to use for this pool?

> Having a pool to store metadata (or config, whatever) makes sense to me. What's the big deal about this? Especially if
the objective is to (if I understand correctly) having either deepsea or some other external tool (e.g., the dashboard)
creating that pool?

jecluis:
>> Am 15.03.2019 um 19:41 schrieb Joao Eduardo Luis ***@***.***>: Well, can anyone explain in a concise manner what is the hold up here?

> Do we want to have a pool for two to five objects only? How many PGs do you want to use for this pool?

Yes. One should suffice. We're not expecting it to grow. **IF** it does, we can all go crazy and make it all two pgs and live on the edge.

Martin-Weiss:
> Am 15.03.2019 um 20:44 schrieb Joao Eduardo Luis <notifications@github.com>:
> 
> Am 15.03.2019 um 19:41 schrieb Joao Eduardo Luis ***@***.***>: Well, can anyone explain in a concise manner what is
the hold up here?
> 
> Do we want to have a pool for two to five objects only? How many PGs do you want to use for this pool?
> 
> Yes. One should suffice. We're not expecting it to grow. IF it does, we can all go crazy and make it all two pgs and
live on the edge.
> 
Ok - just make sure we document that the number of PGs for this pool should not be changed and also update the PG
calculator with this info.

(We should to avoid that someone increases the PG number to get rid of the „not enough PG warning“ - so even some sort
of a hidden pool feature would be nice).

jecluis:
> Ok - just make sure we document that the number of PGs for this pool should not be changed and also update the PG calculator with this info. (We should to avoid that someone increases the PG number to get rid of the „not enough PG warning“ - so even some sort of a hidden pool feature would be nice).

I think you are hung up on that warning. The `not enough pgs` warning is a cluster-wide check, considering all PGs across all pools, and it pertains the number of PGs per osd -- on the whole cluster. It's not a 1 PG pool that is going to trigger that warning (well, unless it's the only pool, but in that case, I mean... wat?)

If someone increases the number of pgs for this pool, so what? The worst that could happen is having one object per PG in that pool. How is that a problem, even? Want to document that? Sure, whatever. But that's neither a problem, nor something I foresee being a problem at all.

And, honestly, keep trying to poke holes in this approach seems an exercise in futility. Let the people that have been working on this, and thinking about this, do their jobs.

rjfd:
From the comments above there are two different problems under discussion:

1. if it makes sense to have a single (small) pool to store ceph-iscsi configuration, or other services configuration like nfs-ganesa
2. if DeepSea should create demo data for the services it deploys. Example: RBD data pools plus iSCSI targets for ceph-iscsi service; NFS exports for nfs-ganesha service; S3 buckets for RGW service; etc...

Regarding problem 1. there is no performance penalty for having a pool with just 1 PG, and has the benefit of creating a clear separation between pools for data, and pools for configuration/metadata of services. This pool doesn't need to grow PG-wise and thus the number of PGs should not be changed (it would be nice if we could "protect" a pool for disallowing changes).

Regarding problem 2. I think for POC purposes it makes totally sense to create demo data to show of the several services functionality. What I don't think it makes sense is for creating this demo data in a standard deployment path automatically (as the default). What I suggest is to have an additional stage, called "demo", where we can explicitly create demo data for the several services. For instance, we could create some demo iSCSI targets by running `deepsea stage run ceph.stage.demo.iscsi`. Or we could create demo data for all services by running `deepsea stage run ceph.stage.demo`.

jecluis:
> Regarding problem 1. there is no performance penalty for having a pool with just 1 PG, and has the benefit of creating a clear separation between pools for data, and pools for configuration/metadata of services. This pool doesn't need to grow PG-wise and thus the number of PGs should not be changed (it would be nice if we could "protect" a pool for disallowing changes).

I think this can be roughly an afternoon of work, but I'm not sure it will pass upstream's concerns. I'll try to address it today or tomorrow, depending on progress on other unrelated matters.

LenzGr:
> From the comments above there are two different problems under discussion:
> 
> 1. if it makes sense to have a single (small) pool to store ceph-iscsi configuration, or other services configuration like nfs-ganesa

Since this is how we decided that these services are supposed to be configured, I think we're past the "if it makes sense" phase, aren't we? I think the question is more if that pool should exist (or created) by default and what name it should be given. This is something I would love to have alignment with upstream (and other downstream) vendors, to avoid creating yet another namespace difference.

My proposal would be to name it `ceph-config` (but that's just me), and it should be created by default (not via an individual deployment tool).

> 2. if DeepSea should create demo data for the services it deploys. Example: RBD data pools plus iSCSI targets for ceph-iscsi service; NFS exports for nfs-ganesha service; S3 buckets for RGW service; etc...

> Regarding problem 1. there is no performance penalty for having a pool with just 1 PG, and has the benefit of creating a clear separation between pools for data, and pools for configuration/metadata of services. This pool doesn't need to grow PG-wise and thus the number of PGs should not be changed (it would be nice if we could "protect" a pool for disallowing changes).

Ack.
 
> Regarding problem 2. I think for POC purposes it makes totally sense to create demo data to show of the several services functionality. What I don't think it makes sense is for creating this demo data in a standard deployment path automatically (as the default).

Agreed. While it's nice to be able to have an environment that makes it easy to get going, we should avoid creating all of this by default, but making it possible to set it up, if the users requests it explicitly (security by default).

> What I suggest is to have an additional stage, called "demo", where we can explicitly create demo data for the several services. For instance, we could create some demo iSCSI targets by running `deepsea stage run ceph.stage.demo.iscsi`. Or we could create demo data for all services by running `deepsea stage run ceph.stage.demo`.

Ack.

l-mb:
Having everything in one pool has issues with multi-tenancy when configuration needs to be separated out cleanly for (multiple instances of) different services.

Having everything in just one PG also means that this single PG experiencing problems (up to data corruption) would down or even corrupt all services.

I think storing the configs in the respective service's metadata pool seems more robust.

rjfd:
> Having everything in one pool has issues with multi-tenancy when configuration needs to be separated out cleanly for (multiple instances of) different services.
> 
> Having everything in just one PG also means that this single PG experiencing problems (up to data corruption) would down or even corrupt all services.
> 
> I think storing the configs in the respective service's metadata pool seems more robust.

Those are two valid points that I didn't thought about. I'll revert the implementation.

rjfd:
I've pushed a new commit where we try to find an existing rbd pool, with a preference for a pool called `iscsi-images`, and if we don't find it we create it.

swiftgist:
> > @rjfd If you delete the init.sls, then how do you propose that a site would customize the iscsi configuration if needed?
> 
> @swiftgist I can change this to match the structure of other parts of DeepSea as you are suggesting but I wonder what exactly needs to be customized and what is the real benefit?
> 

When planning for the unknown, the benefit is having a solution for the one or two sites that want to do something completely unexpected (and likely unsupported).  

> If we are talking about using different values for the options declared in `iscsi-gateway.cfg.j2`, wouldn't make more sense to allow the user to override those values in the salt pillar instead of copy&paste several files, which we will never control in an upgrade scenario?

The pillar is fine... just have to worry about collisions.  But having an admin know that their custom files are never touched/modified means not concerning ourselves with a configuration that is arguably drastically different.

If it turns out that some other configuration is a more common use case for some group, we can add it - but the admin/SE or anybody in the field has a workaround and is stuck waiting for a patch.

jschmid1:
@rjfd 

[91minstall: cannot stat 'srv/salt/ceph/igw/restart/reload/*.sls': No such file or directory

rjfd:
@jschmid1 I already fixed the makefile issue, what's the command to re-run the checks?

jschmid1:
retest this please

jschmid1:
huh, this should've triggered a round of tests  @kshtsk ?

smithfarm:
adding a commit or force-pushing should also re-trigger the tests. . . but doesn't.

rjfd:
@susebot run teuthology

srv/salt/ceph/stage/deploy/core/default.sls
```diff
@@ -127,16 +127,17 @@ dashboard:
...
-  salt.state:
-    - tgt: {{ master }}
-    - tgt_type: compound
-    - sls: ceph.mgr.orchestrator
-    - pillar:
-        'salt_api_url': http://{{ master }}:8000/
-        'salt_api_username': admin
-        'salt_api_password': {{ salt.saltutil.runner('sharedsecret.show') }}
-    - failhard: True
+## disabled for now
```
sebastian-philipp: disable for now, cause the ds orchestrator doesn't provide the iscsi url which takes priority in the dashboard

rjfd:
@susebot run teuthology

rjfd:
@jschmid1 tests failed only for CLI enabled and the issue is already tracked in #1587 

So I think we can merge this.

jschmid1:
right, the AA issues.. merging

PR #1584, opened 2019-03-26
cli: fix function args string formating

rjfd:


jschmid1:
@susebot run teuthology

jschmid1:
This issue seems to be caused by the (known) rgw+ipv6 issue

PR #1588, opened 2019-03-26
qa: ensure mgr is available prior to orchestrator tests

tserong:
The orchestrator functests are run immediately after the
terminate functests.  It's possible mgr hasn't quite finished
re-loading at this point, which will cause the orchestrator
functests to fail.

### DIFF
--- a/srv/salt/ceph/tests/orchestrator/init.sls
+++ b/srv/salt/ceph/tests/orchestrator/init.sls
@@ -1,13 +1,23 @@
+ensure mgr is available:
+  module.run:
+    - name: retry.cmd
+    - kwargs:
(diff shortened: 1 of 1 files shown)


tserong:
@susebot run teuthology

tserong:
Well, there goes that theory...

tserong:
@susebot run teuthology

tserong:
Well, that's horrible. Some time during the functests, *before* the orchestrator functests, ceph gets stuck in HEALTH_ERR with the message "7 mgr modules have failed".

tserong:
Turns out the problem is the ceph-mgr apparmor profile (see #1589)

PR #1594, opened 2019-03-28
Move crc_method to master.d/crc_method

### LINKED ISSUE #1593: "DeprecationWarning: This server_id is computed nor by Adler32 neither by CRC32" in salt-run output
#1530 appends "server_id_use_crc: adler32" to /etc/salt/minion, but we also need this appended to /etc/salt/master, otherwise salt-run output is polluted with "[WARNING ] /usr/lib/python3.6/site-packages/salt/grains/core.py:2802: DeprecationWarning: This server_id is computed nor by Adler32 neither by CRC32. Please use "server_id_use_crc" option and define algorithm youprefer (default "Adler32"). The server_id will be computed withAdler32 by default."

jschmid1:
Fixes #1593 
Fixes: aa40d37501600d1dd9c6c2bf650e61e49058c74a

~For me it seems to now suffice to have this entry *only* in the master conf.~

nope, see https://github.com/SUSE/DeepSea/pull/1594#issuecomment-477898215

This PR will add the 'crc_method' to `/etc/salt/minion.d/crc_method.conf` and `/etc/salt/master.d/crc_method.conf` respectively.

resolves: SES-351

-----------------

*Checklist:**

### DIFF
--- a/srv/salt/ceph/salt/crc/default.sls
+++ b/srv/salt/ceph/salt/crc/default.sls
@@ -1,4 +1,0 @@
-/etc/salt/minion:
-  file.append:
-    - text:
-      - "server_id_use_crc: adler32"
--- a/srv/salt/ceph/stage/prep/master/default-no-update-no-reboot.sls
+++ b/srv/salt/ceph/stage/prep/master/default-no-update-no-reboot.sls
@@ -1,4 +1,3 @@
-
 {% set master = salt['master.minion']() %}
 
 {% if salt['saltutil.runner']('validate.setup') == False %}
@@ -10,6 +9,11 @@ validate failed:
     - failhard: True
 
 {% endif %}
+
+crc_method:
+  salt.state:
+    - tgt: {{ master }}
+    - sls: ceph.salt.crc.master
 
 sync master:
   salt.state:
@@ -52,7 +56,3 @@ ready:
   salt.runner:
     - name: minions.ready
     - timeout: {{ salt['pillar.get']('ready_timeout', 300) }}
-
-
-
-
--- a/Makefile
+++ b/Makefile
@@ -915,7 +915,11 @@ copy-files:
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/salt
 	install -m 644 srv/salt/ceph/salt/*.sls $(DESTDIR)/srv/salt/ceph/salt/
 	install -d -m 755 $(DESTDIR)/srv/salt/ceph/salt/crc
-	install -m 644 srv/salt/ceph/salt/crc/*.sls $(DESTDIR)/srv/salt/ceph/salt/crc/
+	install -m 644 srv/salt/ceph/salt/crc/*.conf $(DESTDIR)/srv/salt/ceph/salt/crc/
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/salt/crc/minion
+	install -m 644 srv/salt/ceph/salt/crc/minion/*.sls $(DESTDIR)/srv/salt/ceph/salt/crc/minion/
+	install -d -m 755 $(DESTDIR)/srv/salt/ceph/salt/crc/master
+	install -m 644 srv/salt/ceph/salt/crc/master/*.sls $(DESTDIR)/srv/salt/ceph/salt/crc/master/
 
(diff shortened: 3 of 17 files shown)


jschmid1:
@susebot run teuthology

srv/salt/ceph/stage/prep/master/default-no-update-no-reboot.sls
```diff
@@ -11,6 +10,11 @@ validate failed:
 
 {% endif %}
 
+crc_method:
+  salt.state:
+    - tgt: {{ master }}
+    - sls: ceph.salt.crc
```
smithfarm: Any reason why `failhard: True` is omitted here?

smithfarm:
In upstream Ceph, there is a "best practice" of adding

```
Fixes: aa40d37501600d1dd9c6c2bf650e61e49058c74a
```

to the commit message for follow-on fixes. Maybe worth considering in this case?

srv/salt/ceph/stage/prep/master/default-no-update-no-reboot.sls
jschmid1: I think it shouldn't failhard here. It's more or less just suppressing a warning (at least that's the surfacing result when not applying this config option).

tserong:
> For me it seems to now suffice to have this entry _only_ in the master conf.

I'm not sure, I tried only having it on the master, and removing it from a minion, then when i restart the minion, I get these warnings in the minion logs. Will test a bit more tomorrow.

srv/salt/ceph/stage/prep/master/default-no-update-no-reboot.sls
smithfarm: If failure of this state doesn't cause the whole Stage to fail, then obviously there's no reason to add failhard.

I guess I have a testing mentality - whenever a program continues running after doing something that will cause the test to fail, it makes debugging the failure harder.

tserong:
Confirmed. Here's my test:

1) Make sure "server_id_use_crc" is *not* set in either master or minion config. Reboot just to be sure.
2) Check /var/log/salt/master and /var/log/salt/minion. Both contain the deprecation warnings.
3) Add "server_id_use_crc: adler32" to /etc/salt/master. Restart the master, check the log file. No warnings. Restart the minion, check the log file. The warnings are still present.
4) Add "server_id_use_crc: adler32" to /etc/salt/minion. Restart the minion, check the log file. No warnings.

srv/salt/ceph/salt/crc/default.sls
```diff
@@ -1,4 +1,4 @@
-/etc/salt/minion:
+/etc/salt/master.d/crc_method:
```
jan--f: Nit: all our config fragments in [master,minion].d have a `.conf` extension. Maybe we should keep things consistent?

srv/salt/ceph/salt/crc/default.sls
```diff
@@ -1,4 +1,4 @@
...
   file.append:
```
jan--f: Whoops...that should've been part of the review:
`file.append` seems like the wrong method here. Conceptually this file is fully under our control so `file.managed` with a `content:` section would make more sense. Not sure what file.append does if someone appends a line and re-runs the salt state...append the line again? Certainly a bit hair-split-y...

srv/salt/ceph/salt/crc/default.sls
jschmid1: ack

srv/salt/ceph/salt/crc/default.sls
jschmid1: good point, I'll adapt it.

jschmid1:
@susebot run teuthology

jschmid1:
> Commit [d3c2bfc](https://github.com/SUSE/DeepSea/commit/d3c2bfc4868711e77d5163e536f233a7317e859f) is _NOT OK_ for suite deepsea:tier2.
> Check tests results in the Jenkins job: http://storage-ci.suse.de:8080/job/pr-deepsea/136/

yikes, forgot to add the .conf

jschmid1:
@susebot run teuthology

PR #1596, opened 2019-03-29
iscsi: use iSCSi gateway public IP address in dashboard configuration

rjfd:
This PR fixes a bug that occurs when the cluster DNS configuration maps the same hostname to different IP addresses. We will now make sure the dashboard uses the correct IP address where the iSCSI gateway is listening.


rjfd:
@susebot run teuthology

Issue #1580, opened 2019-03-25
[DeepSea CI] Stage 4 reproducibly fails at RGW bucket creation


smithfarm:
### Description of Issue/Question

The DeepSea PR CI (for PRs targeting master) recently started to see failures like this:

CLI:

```
019-03-25T10:15:16.911 INFO:teuthology.orchestra.run.target192168000018.stdout:Failures summary:
2019-03-25T10:15:16.911 INFO:teuthology.orchestra.run.target192168000018.stdout:ceph.rgw.buckets (/srv/salt/ceph/rgw/buckets):
2019-03-25T10:15:16.912 INFO:teuthology.orchestra.run.target192168000018.stdout:  target192168000018.teuthology:
2019-03-25T10:15:16.912 INFO:teuthology.orchestra.run.target192168000018.stdout:    create demo bucket for demo: Module function rgw.create_bucket threw an exception. Exception: [Errno 111] Connection refused
2019-03-25T10:15:16.912 INFO:teuthology.orchestra.run.target192168000018.stdout:    create demo bucket for demo1: Module function rgw.create_bucket threw an exception. Exception: [Errno 111] Connection refused
```

non-CLI:

```
2019-03-25T10:25:20.163 INFO:teuthology.orchestra.run.target192168000004.stdout:----------
2019-03-25T10:25:20.163 INFO:teuthology.orchestra.run.target192168000004.stdout:          ID: rgw demo buckets
2019-03-25T10:25:20.163 INFO:teuthology.orchestra.run.target192168000004.stdout:    Function: salt.state
2019-03-25T10:25:20.163 INFO:teuthology.orchestra.run.target192168000004.stdout:      Result: False
2019-03-25T10:25:20.163 INFO:teuthology.orchestra.run.target192168000004.stdout:     Comment: Run failed on minions: target192168000004.teuthology
2019-03-25T10:25:20.164 INFO:teuthology.orchestra.run.target192168000004.stdout:     Started: 10:18:24.512374
2019-03-25T10:25:20.164 INFO:teuthology.orchestra.run.target192168000004.stdout:    Duration: 140975.18 ms
2019-03-25T10:25:20.164 INFO:teuthology.orchestra.run.target192168000004.stdout:     Changes:
2019-03-25T10:25:20.164 INFO:teuthology.orchestra.run.target192168000004.stdout:              target192168000004.teuthology:
[...]

PR #1600, opened 2019-03-29
Remove RGW demo bucket creation

swiftgist:
With the support of IPv6, the parsing in rgw.py would need to be much
more sophisticated to handle civetweb, beast and the various keyword
parameters for either.

After discussion with Abhi and Ricardo D., the demo buckets were only
necessary for the standalone openATTIC and ganesha.  With oA as the Ceph
dashboard and a newer implementation for configuring ganesha from the
Ceph dashboard, these buckets are not needed.

-----------------

**Checklist:**
- [ ] Adapted documentation


swiftgist:
@susebot run teuthology

smithfarm:
Note: this should probably reference #1580

PR #1585, opened 2019-03-26
nfs-ganesha: deploy ganesha configuration in RADOS pool

rjfd:
This PR changes the deployment of nfs-ganesha to match the requirements of ceph-dashboard.

Signed-off-by: Ricardo Dias <rdias@suse.com>

PR #1606, opened 2019-04-03
mgr_orch: return ganesha and iscsi endpoint URLs

tserong:
This changes describe_service() so it will return a dict for each
service, which includes at minimum service_instance (i.e. the daemon
ID), and also rados_config_location and service_url for ganesha
and igw instances respectively.

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/modules/runners/mgr_orch.py
+++ b/srv/modules/runners/mgr_orch.py
@@ -7,12 +7,26 @@ Runner endpoints for integration specifically with the ceph-mgr orchestrator
 
 from __future__ import absolute_import
 import logging
+import ipaddress
 import os
 import sys
 # pylint: disable=import-error,3rd-party-module-not-gated,redefined-builtin
 import salt.client
 
 log = logging.getLogger(__name__)
+
+
+def _run_master_module_function(function, *args, **kwargs):
+    """
+    Used internally by this module to run master.find_pool.  Could also be
+    used to run master.minion if it turns out we need that.
+    """
+    __opts__ = salt.config.client_config('/etc/salt/master')
+    __grains__ = salt.loader.grains(__opts__)
+    __opts__['grains'] = __grains__
+    __utils__ = salt.loader.utils(__opts__)
+    __salt__ = salt.loader.minion_mods(__opts__, utils=__utils__)
+    return __salt__[function](*args, **kwargs)
 
 
 def get_inventory(nodes=None, roles=None):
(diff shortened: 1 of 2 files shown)


tserong:
@rjfd this exposes the iscsi and ganesha endpoint URLs so they can be picked up by the deepsea orchestrator module in https://github.com/ceph/ceph/pull/27336. The URLs are constructed in the same form as used in https://github.com/SUSE/DeepSea/pull/1596 and https://github.com/SUSE/DeepSea/pull/1585. Once we've got the orchestrator piece doing the right thing, should we then remove the calls to `ceph dashboard set-ganesha-clusters-rados-pool-namespace` and `ceph dashboard iscsi-gateway-add`?

jschmid1:
looks good but would look even better with tests :)

tserong:
> looks good but would look even better with tests :)

you're right of course, I'll do something about that :-)

tserong:
Tests updated to match, and also fixed so it correctly takes into account ceph_iscsi_{username,password,port,ssl} pillar variables.

tserong:
Test failures are unrelated (ENOSPC)

jschmid1:
retest this please

srv/modules/runners/mgr_orch.py
```diff
@@ -125,12 +140,33 @@ def describe_service(role=None, service_id=None, node=None):
...
+                                         ['ceph_iscsi_username'])[minion] or 'admin'
+                igw_password = local.cmd(minion, 'pillar.get',
+                                         ['ceph_iscsi_password'])[minion] or 'admin'
+                igw_port = local.cmd(minion, 'pillar.get',
+                                     ['ceph_iscsi_port'])[minion] or '5000'
+                igw_ssl = local.cmd(minion,
+                                    'pillar.get', ['ceph_iscsi_ssl'])[minion] or False
+                protocol = 'https' if igw_ssl else 'http'
+                minion_result[minion_role]['service_url'] = (
+                    '{}://{}:{}@{}:{}'.format(protocol, igw_username, igw_password,
```
rjfd: If `igw_address` is an IPv6 address it must be enclosed in square brackets in the URL

rjfd:
@tserong I don't think we should start using the orchestrator to get ganesha and iscsi info before improving the latency of that operation. But the PR as it is looks good to me.

srv/modules/runners/mgr_orch.py
tserong: Thanks, will fix.

tserong:
> I don't think we should start using the orchestrator to get ganesha and iscsi info before improving the latency of that operation

Agreed, but we should be able to fix the latency with caching in the deepsea mgr module, so I'll do that separately to this PR.

jschmid1:
```
14:02:34 ************* Module srv.modules.runners.mgr_orch
14:02:34 C:159,19: Using type() instead of isinstance() for a typecheck. (unidiomatic-typecheck)
```

tserong:
Oops... :-/

tserong:
That's better.

PR #1607, opened 2019-04-03
IPv6 missing changes

swiftgist:
The ceph.conf.j2 and module changes were missing.

-----------------

**Checklist:**
- [ ] Adapted documentation


swiftgist:
@susebot run teuthology

srv/salt/_modules/public.py
```diff
@@ -51,3 +51,43 @@ def address():
...
+    if 'public_network' not in __pillar__:
+        return ""
+    return _ipnetwork(6)
+
+
+def ipv4():
+    """
+    Return if any network is IPv4
+    """
+    if 'public_network' not in __pillar__:
```
jan--f: nit picking but maybe push this down to `_ipnetwork` so we only need it once?

srv/salt/ceph/configuration/files/ceph.conf.j2
```diff
@@ -18,6 +18,13 @@ ms_bind_msgr2 = false
 mon_health_preluminous_compat = true
 mon health preluminous compat warning = false
 
+{% set ipv6=salt['public.ipv6']() %}
+{% if ipv6 %}
+ms bind ipv4 = {{ salt['public.ipv4']() | lower }}
```
jan--f: I assume this means ceph can now run with both ipv4 and ipv6? Otherwise I'm confused why ipv6 is hardcoded and ipv4 is queried.

swiftgist: @jan--f Yes, it was for future compatibility.  The validation prevents IPv4 and IPv6 today, but Ceph upstream is planning on supporting both.  When that happens, only the validation will need correcting.  If an admin uses both IPv4 and IPv6 networks, then the ceph.conf will be correct.

As far as hardcoding the IPv6 setting, it will always match the conditional.  So, no need to query a second time.

srv/salt/_modules/public.py
swiftgist: Both ipv6 and ipv4 return strings back to the Salt user.  The _ipnetwork returns bool.

jan--f: But don't `ipv4` and `ipv6` now return either an empty string or a bool? Seems like the same return could be in `_ipnetwork`, reducing code duplication without a change in the public interface.

PR #1611, opened 2019-04-10
fix crc method minion

jan--f:
If deepsea_minions is compound target (e.g. `G@deepsea:*`) stage 0 will only succeed if salt is told about the compound target type.


jan--f:
Somehting went wrong with my branches.

PR #1612, opened 2019-04-10
crc_method: deepsea_minions can be a compound target

jan--f:
If deepsea_minions is compound target (e.g. G@deepsea:*) stage 0 will only succeed if salt is told about the compound target type.


jan--f:
> LGTM, but also needs to be applied to default-no-update-no-reboot.sls, default-no-update-reboot.sls
> and default-update-reboot.sls

Thx, good catch!

Issue #1614, opened 2019-04-16
Deepsea pushes out chrony.conf with no allows in it


smithfarm:
quoting from https://bugzilla.suse.com/show_bug.cgi?id=1132632

Deepsea pushes out a chrony.conf file to all the nodes with no allow statements in it which causes errors in Syslog as below

with no allow statements Chrony doesnt have anything running on port 123.

```
: read udp 127.0.0.1:60117->127.0.0.1:123: read: connection refused" source="collector.go:132"
2019-04-16T05:29:14.054093-04:00 node4 node_exporter[1633]: time="2019-04-16T05:29:14-04:00" level=error msg="ERROR: ntp collector failed after 0.000156s: couldn't get SNTP reply: read udp 127.0.0.1:37524->127.0.0.1:123: read: connection refused" source="collector.go:132"
2019-04-16T05:29:20.093455-04:00 node4 node_exporter[1633]: time="2019-04-16T05:29:20-04:00" level=error msg="ERROR: ntp collector failed after 0.000164s: couldn't get SNTP reply: read udp 127.0.0.1:33941->127.0.0.1:123: read: connection refused" source="collector.go:132"
2019-04-16T05:29:24.054571-04:00 node4 node_exporter[1633]: time="2019-04-16T05:29:24-04:00" level=error msg="ERROR: ntp collector failed after 0.000202s: couldn't get SNTP reply: read udp 127.0.0.1:36014->127.0.0.1:123: read: connection refused" source="collector.go:132"
2019-04-16T05:29:30.094069-04:00 node4 node_exporter[1633]: time="2019-04-16T05:29:30-04:00" level=error msg="ERROR: ntp collector failed after 0.000308s: couldn't get SNTP reply: read udp 127.0.0.1:59926->127.0.0.1:123: read: connection refused" source="collector.go:132"
```

This is on master/SES6.

smithfarm:
Now my understanding is that DeepSea intentionally avoids starting any time server (chrony or otherwise). So this bug and the corresponding downstream bug are probably invalid.

PR #1615, opened 2019-04-16
chrony: allow NTP access from local network

smithfarm:
Without an allow statement, Chrony does not listen on port 123

Fixes: https://github.com/SUSE/DeepSea/issues/1614


smithfarm:
not sure how well this dovetails with the "Support IPv6" feature. . .

swiftgist:
The long ago agreement is that DeepSea would not configure any time server, but only clients.  So, chrony mirrored the original behavior of the ntp setup.  

That said, I'm fine either way.  I do not know if this addresses what Darren is after though.

Issue #1616, opened 2019-04-17
Makefile: "make rpm" gets the version number wrong sometimes


smithfarm:
There are at least three cases when "make rpm" can get the version number wrong:

1. when the latest tag is something like "foobar" instead of "v0.9.17"
2. when the latest tag is a lightweight (as opposed to annotated) tag
~3. when there are zero additional commits following the latest tag~ (already fixed)

smithfarm:
For example, the recently-added `v0.9.17` tag is lightweight whereas previous tags were annotated:

```
smithfarm@localhost:~/src/DeepSea> git for-each-ref refs/tags | grep -P 'v0.9.1\d'
6a058d452ba02092fd628f459e2c974b34e4ad92 tag    refs/tags/v0.9.10
3a8778e9980eff7598eddf379d868410a3e182ab tag    refs/tags/v0.9.11
6e0ac5d4e8b82bcfc0482921e10c45ab0d05b7cc tag    refs/tags/v0.9.12
3f28e9fc6409752540323a63bb4e33a7829b89a5 tag    refs/tags/v0.9.13
e241900a9608769868b3047fa9650bc63e5c1df9 tag    refs/tags/v0.9.14
8d0e63a964e108654cb6f5c36f83f179026c0d98 tag    refs/tags/v0.9.15
56a5ade672f96ff67d257e3ab1aea212e4f068dd tag    refs/tags/v0.9.16
d90e228ff87fd52a03fbdba3a02d13cdfb8f5d35 commit refs/tags/v0.9.17
```

As a result, when you do `git describe` on the tip of master, you get a v0.9.16 version number instead of the expected v0.9.17:

```
$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
$ git describe
v0.9.16-28-ge8e9783d
```

We can fix that by adding `--tags`.

However, the tip of master is currently pointing directly to this `v0.9.17` lightweight tag. The output of `git describe` is a version number, but not in the desired format:

```
$ git describe --tags
v0.9.17
```

jan--f:
TIL that there are two different tags. @jschmid1 is the release process documented somewhere?

jschmid1:
> TIL that there are two different tags. @jschmid1 is the release process documented somewhere?

Yeah, I created [this](https://github.com/SUSE/DeepSea/wiki/Branches-&-Releases) document some time back.

jschmid1:
[...]

PR #1617, opened 2019-04-17
Makefile: make "make rpm" less likely to produce a wrong version number

smithfarm:
Fixes: https://github.com/SUSE/DeepSea/issues/1616


jan--f:
Maybe it worth reverting this once we have a) the release process clearly documented and b) a newer tag? I think it makes sense to rely on annotated tags for releases.

smithfarm:
@jan--f I agree the `--tags` could be removed once we are certain that only annotated tags are used for releases. Then "miscellaneous" (non-version-number) lightweight tags could be used in the main repo without disrupting the version numbering.

PR #1621, opened 2019-04-18
rpm: declare runtime dependencies in the spec file

smithfarm:
DeepSea installs a number of RPMs on minions in order to be able to do its work.
Declaring these as runtime dependencies in the deepsea spec file does not mean
DeepSea no longer needs to use salt to install these RPMs on the minions.

However, IMO it makes sense to declare all of these in the spec file. The
rationale for this is: if any of these packages can not be installed along with
DeepSea, there is a high probability DeepSea will fail to install them on the
minions as well. Instead of obscure salt error messages when running Stages,
we'll get a (hopefully clearer) unfulfilled dependency error from zypper when 
installing the deepsea RPM.

Also, by declaring the dependency in the spec file we can include version guards
like "> 0.14" if needed.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1129711


jschmid1:
shall we put 'ceph' in there aswell?

smithfarm:
> shall we put 'ceph' in there aswell?

No :-)

(Rationale: the whole purpose of DeepSea is to deploy Ceph clusters, so the need for repos containing ceph packages is obvious. The runtime dependencies being declared in this PR are **not** obvious.)

PR #1625, opened 2019-04-24
Avoid running drivegroup based osd deployment post-upgrade

jschmid1:
In a post-upgrade scenario the user still has old-style profiles defined in the
pillar.

Since we don't force the user to move to the new drive-group implementation
right after the upgrade, we have to detect that they are running in 'legacy'
mode and stop.

Fixes #

Description:

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/salt/_modules/dg.py
+++ b/srv/salt/_modules/dg.py
@@ -823,6 +823,14 @@ def c_v_commands(**kwargs):
 
 def deploy(**kwargs):
     """ Execute the generated ceph-volume commands """
+    # do not run this when there is still ceph:storage in the pillar
+    # this indicates that we are in a post-upgrade scenario and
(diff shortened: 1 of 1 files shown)


jschmid1:
@susebot run teuthology

jschmid1:
> Makes sense, this should also avoid having the old definitions hanging around. Maybe add a hint to remove any `profile-*` lines from policy.cfg?

I added a placeholder for future references.. Only telling users to remove the profile line from the policy will probably result in a mess.

We have to have a proper document ready that describes the profiles -> drive-group migration process

jschmid1:
@susebot run teuthology

jschmid1:
From here: https://github.com/SUSE/DeepSea/issues/1629

Issue #1623, opened 2019-04-23
Enhance handling of 'Unknown host'


jschmid1:
### Description of Issue/Question

Bug: https://bugzilla.suse.com/show_bug.cgi?id=1132271 

~This gives a stacktrace due to ' [ERROR   ] Exception during resolving address: [Errno 1] Unknown host'.~

In disks.py:207

``` python
def destroyed():
    """ List destroyed (about to be replaced) disks
    """
    # This can't be solved with Popen since the salt-master is running as salt:salt
    local_client = salt.client.LocalClient()
    ret: str = local_client.cmd(
        "roles:master",  # <--- If this is not assigned, we get {} back
        'cmd.shell', ['ceph osd tree destroyed --format json'],
        tgt_type='pillar')

    ret = list(ret.values())[0]     <-- improve this
    tree = json.loads(ret).get('nodes')
```

This runs into an IndexError when there is a bad return.

edit: This happens due to an unassigned master role

jan--f:
Yep saw this too. Happens when the disks runner is ran before roles are assigned. I have a fix in the pipe.

jschmid1:
was about to write that :dagger:

jschmid1:
On 19/04/24 04:08, Joshua Schmid wrote:
> was about to write that :dagger: 
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/SUSE/DeepSea/issues/1623#issuecomment-486178252

Corrected the issue.

-- 
Joshua Schmid
Software Engineer
SUSE Enterprise Storage

PR #1626, opened 2019-04-24
disks runner: guard destroyed() against roles:master being absent

jan--f:
Fixes: #1623


jschmid1:
the `TODO` is not accepted by the linter

jan--f:
oh that is also not needed anymore...fixing

jschmid1:
@susebot run teuthology

PR #1628, opened 2019-04-25
core: do not skip disks that are not used by ceph

jschmid1:
This affects disks.details and disks.report/deploy
We have a function that explicitly asks for disks
that are used by ceph. The default should be 'None'.

bsc:1131161

-----------------

**Checklist:**

### DIFF
--- a/srv/salt/_modules/cephdisks.py
+++ b/srv/salt/_modules/cephdisks.py
@@ -26,14 +26,14 @@ class Inventory(object):
     @property
     def available_filter(self) -> bool:
         """ The available filter """
-        return self.kwargs.get('available', None)
+        return self.kwargs.get('available', False)
 
     @property
     def used_by_ceph_filter(self) -> bool:
         """ The used_by_ceph filter """
         # This also returns disks that are marked as
         # 'destroyed' is that valid?
-        return self.kwargs.get('used_by_ceph', True)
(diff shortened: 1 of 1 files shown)


jschmid1:
@susebot run teuthology

jschmid1:
> Hmm I'm a bit confused by how this is used. So this change just sets the default of cephdisks to ignore devices used by ceph, right?

Ahm, not quite. This un-sets the filter completely, which was previsously enabled by default and had the following side-effect.

Imagine a cluster with 10 disks. 8 of them are used by ceph. 2 of them are not. 
You now want to get the details of those disks as you are planning to create a drive-group for them. (salt-run disks.details). With the `used_by_ceph` flag being true we fall into the branch of

``` python
    elif self.used_by_ceph_filter:
        if dev.used_by_ceph:
            devs.append(dev)
```

since the two bare disks actually don't have this flag, they are not appended to the disks list (which is used by `disks.details`).

-------

I do agree that 

``` python
elif self.used_by_ceph_filter:
    if dev.used_by_ceph:
        devs.append(dev)
``` 

is the better and less confusing solution :>

let me adapt the implementation accordingly.

jschmid1:
pushed @jan--f

jschmid1:
@susebot run teuthology

PR #1630, opened 2019-04-30
ceph-volume: use re to extract json string from c-v output

jan--f:
This awful crutch should make life easier until
http://tracker.ceph.com/issues/38548 is fixed.

Signed-off-by: Jan Fajerski <jfajerski@suse.com>

PR #1631, opened 2019-04-30
do no parse ceph output when no valid json

jschmid1:
disks.details/deploy is used before a cluster is deployed
and no json will be returned

-----------------

**Checklist:**

### DIFF
--- a/srv/modules/runners/disks.py
+++ b/srv/modules/runners/disks.py
@@ -210,7 +210,12 @@ def destroyed():
 
     tree = {}
     if ret:
-        tree = json.loads(list(ret.values())[0]).get('nodes')
+        try:
+            tree = json.loads(list(ret.values())[0]).get('nodes')
+        except json.decoder.JSONDecodeError:
+            log.info(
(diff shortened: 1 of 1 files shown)


jschmid1:
@susebot run teuthology

jan--f:
I think #1630 addresses the same issue?

jschmid1:
while #1630 tries to handle dirty json output from c-v, we _know_ that we get no json from an unconfigured ceph here. I think it's overkill to regex over this.

jschmid1:
one issue looks unrelated..(ganesha) 
the other _may_ indicate an issue.. re-running to be sure

jschmid1:
@susebot run teuthology

jan--f:
Oh right...that json is from ceph osd. Will remove that from #1630 then.

jschmid1:
> Oh right...that json is from ceph osd. Will remove that from #1630 then.

my patch is handling disks.destroyed. I think that's not touched by your PR

jan--f:
well not anymore ;)

PR #1632, opened 2019-05-02
core: osd.remove to accept compound targets

jschmid1:
This simplifies ceph.purge osd handling and avoids jinja based
osd_id iterations in sls files.

bsc#1133706

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/modules/runners/osd.py
+++ b/srv/modules/runners/osd.py
@@ -42,13 +42,44 @@ class NotOkToStop(Exception):
     pass
 
 
-class OSDUtil(object):
+class Util(object):
+    """
+    Util helper class
+    """
+
+    local = salt.client.LocalClient()
+
+    @staticmethod
+    def join_list(inp):
+        """ Join a list of integers to a list of strings """
+        return " ".join([str(x) for x in list(inp)])
+
+    @classmethod
+    def get_osd_list_for(cls, target):
+        """ call salt's LocalClient for osd.list with target """
+        return cls.local.cmd(target, "osd.list", tgt_type="compound")
+
+    @staticmethod
+    def master_minion():
+        """
+        Load the master modules
+        """
+        __master_opts__ = salt.config.client_config("/etc/salt/master")
+        __master_utils__ = salt.loader.utils(__master_opts__)
+        __salt_master__ = salt.loader.minion_mods(
+            __master_opts__, utils=__master_utils__)
+
+        return __salt_master__["master.minion"]()
+
+
+class OSDUtil(Util):
     """ Util class for OSD handling """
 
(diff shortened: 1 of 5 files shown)


jschmid1:
@susebot run teuthology

srv/modules/runners/osd.py
```diff
@@ -410,13 +408,44 @@ def pre_check(osd_list, force):
...
     """ User facing osd.remove function """
     results = dict()
     kwargs.update({'operation': 'remove'})
-    osd_list = " ".join([str(x) for x in list(args)])
-    pre_check(osd_list, kwargs.get('force', False))
-    for osd_id in args:
+    osd_list, osd_list_string = _target_lookup(args)
+    if not osd_list:
+        return False
+    pre_check(osd_list_string, kwargs.get('force', False))
```
tserong: Rather than having `_target_lookup` return both a list and a string, it might simplify things to have it return only a list, then `.join` that into a string here, which seems to be the only place the string is needed.

jschmid1: Right, I have a comment about that somewhere in my notes.. I think it'd be even better to move the list->str conversion `pre_check`. I'll update this shortly

srv/modules/runners/osd.py
```diff
@@ -403,20 +419,48 @@ def pre_check(osd_list, force):
...
     if force:
         print("The 'force' flag is set. ok-to-stop checks are disabled."
               " Please use with caution.")
     else:
         ok_to_stop_osds(osd_list)
 
 
+def _target_lookup(inp):
+    """
+    This allows to specify a compound taret on the commandline
```
jan--f:
```suggestion
    This allows to specify a compound target on the command line
```

jschmid1:
> I like the idea. [The _target_lookup function](https://github.com/SUSE/DeepSea/pull/1632/files#diff-6388310932c2fa32de25d711a368c5f6R430) however gives me some headaches.
> 
> If the first argument `isinstance(int)` the whole list is assumed to be OSD IDs. What prevents a user from specifying some OSD IDs and a compound target?
> If one argument is identified as a compound target but it matches nothing the whole operation is aborted, whereas is seems generally possible to specify _multiple_ compound targets. As long as they all match something this seems to work too (as long as there're no ID duplicates...didn't check).

to be honest, the lack of parsing is mainly due to lazyness and priority. The main purpose of this addition is to simplify our salt/jinja for `ceph.purge`. I'm not sure if people ever want to remove entire nodes outside of a _purge_ operation. If we however find a valid point why this may be useful, I'd love to adjust this function to be more accurate.

PR #1638, opened 2019-05-06
core: add include_unavailable flag for disks.py

jschmid1:
This serves mainly the purpose of allowing to show matching disks
in a already deployed cluster.
There needs to be such a functionality in order to make migration
from profile-based to drive-group-based deployments easier.

Docs will point people towards:

`salt-run disks.list include_unavailable`

-----------------

**Checklist:**
- [ ] Adapted documentation #TODO
~- [ ] Referenced issues or internal bugtracker~

### DIFF
--- a/srv/salt/_modules/dg.py
+++ b/srv/salt/_modules/dg.py
@@ -538,9 +538,10 @@ class DriveGroup(object):
     """
 
     # pylint: disable=too-many-instance-attributes
-    def __init__(self, filter_args: dict) -> None:
+    def __init__(self, filter_args: dict, include_unavailable=False) -> None:
         self.filter_args: dict = filter_args
         self._check_filter_support()
+        self._include_unavailable = include_unavailable
         self._data_devices = None
         self._wal_devices = None
         self._db_devices = None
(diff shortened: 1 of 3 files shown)


jschmid1:
@susebot run teuthology

jschmid1:
The downside of this:

There is no _easy_ way (besides checking mountpoints etc) to exclude root disks from this list.

I'd love to put this check in ceph-volume, specifically improve on the 'rejected_reason' field to include a 'root_disk' cause or similar, which can be filtered for.

srv/salt/_modules/dg.py
```diff
@@ -741,7 +741,7 @@ def _reduce_inventory(disk: dict) -> dict:
         # maybe this can and should be dropped when the fix is public
         if not disk:
             return {}
-        if disk["available"] is True:
+        if disk.get("available", False) is not self._include_unavailable:
```
tserong: This `if` is hurting my brain, but unless I'm reading it wrong, there's a problem. As written it will do this:

* `disk["available"] is False` and `self._include_unavailable` is False, condition fails, disk is not included (ok)
* `disk["available"] is False` and `self._include_unavailable` is True, condition succeeds, disk is included (ok)
* `disk["available"] is True` and `self._include_unavailable` is False, condition succeeds, disk is included (ok)
* `disk["available"] is True` and `self._include_unavailable` is True, condition fails, disk is not included (not ok)

Maybe `if disk.get("available", False) or self._include_unavailable:` ?

jschmid1: good catch, updating

PR #1640, opened 2019-05-07
NFS-Ganesha and iSCSI upgrade

rjfd:
This PR implements several things besides the upgrades of iSCSI and NFS-Ganesha, here's the list:
* NFS-Ganesha upgrade
* iSCSI upgrade (lrbd -> ceph-iscsi)
* Improved RGW dashboard configuration
* Automatic Grafana dashboard configuration

NFS-Ganesha and iSCSI upgrades are part of stage 4. The upgrade process will detect if the upgrade is needed and performs the upgrade, otherwise it does nothing.


rjfd:
@ricardoasmarques can you review the iSCSI upgrade script in `srv/salt/_modules/iscsi.py`?

srv/modules/runners/iscsi_upgrade.py
```diff
@@ -0,0 +1,187 @@
+# -*- coding: utf-8 -*-
+"""
+NFS-Ganesha Upgrade script
```
Martin-Weiss: Seems that this text "NFS-Ganesha Upgrade script" needs to be changed to "iSCSI Upgrade script" as this is not the iSCSI upgrade script I understand ;-)

rjfd: good catch! fixed

jschmid1:
@susebot run teuthology

rjfd:
@susebot run teuthology

rjfd:
@susebot run teuthology

rjfd:
@susebot run teuthology

rjfd:
@susebot run teuthology

rjfd:
@susebot run teuthology

rjfd:
@jschmid1 integration tests look good. I know that tests are always important but due to time constraints and because the benefit adding unit tests for this particular upgrade process is not that advantageous comparing to run integration tests and real upgrade tests, I did not add any new unit tests.

srv/salt/ceph/ganesha/restart/controlled/default.sls
```diff
@@ -1,20 +1,17 @@
-{% if salt['cephprocesses.need_restart'](role='ganesha') == True %}
-
 restart nfs-ganesha:
-  module.run:
-    - name: service.restart
-    - m_name: nfs-ganesha
+  deepsea.state_apply_if:
```
jschmid1: What's the reasoning behind changing the current theme?

tests/unit/_modules/test_cephprocesses.py
```diff
@@ -203,10 +195,9 @@ def test_add_rgw(self, mc):
     def test_add_igw(self, mc):
         mc.up = []
         role_name = 'igw'
-        proc_name = 'rbd-target-gw'
+        proc_name = 'rbd-target-api'
         mc.add(self.build_proc(role_name, proc_name), role_name)
-        assert len(mc.up) == 1
-        assert mc.up[0].name == proc_name
+        assert len(mc.up) == 0
```
jschmid1: that's due to the deleted entry in `cephprocesses.py`

srv/salt/_modules/cephprocesses.py
```diff
@@ -29,7 +29,7 @@
              'mgr': ['ceph-mgr'],
              'storage': ['ceph-osd'],
              'mds': ['ceph-mds'],
-             'igw': ['rbd-target-gw', 'rbd-target-api'],
```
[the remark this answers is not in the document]
rjfd: those processes don't exist in SES5, and during an upgrade the cephprocesses.wait will fail because those processes don't exist.
We can add those processes in the next version of DeepSea.

srv/salt/ceph/stage/iscsi/default.sls
```diff
@@ -1,4 +1,4 @@
 
 include:
-  - .core
-  - ...restart.igw.lax
```
jschmid1: I might have overlooked it but I can't see any restart logic for igw.

srv/salt/ceph/ganesha/restart/controlled/default.sls
rjfd: when running the stage using deepsea-cli the the restart check made in jinja runs before the code that actually sets the grain value, which causes deepsea-cli to think that that state will not be executed.

This changes moves the check to the "runtime" phase instread of the "parsing" phase.

srv/salt/ceph/stage/iscsi/default.sls
rjfd: I removed it because I thought we would not need it. But now that I'm thinking again about this, we may need it for the case where we have updated the ceph-iscsi packages and need to restart the services. I'll add it back

srv/salt/_modules/cephprocesses.py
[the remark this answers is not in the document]
jschmid1: well, but how can you guarantee in which version people will upgrade. some people will do it in one month and some in a year..

ricardoasmarques:
@rjfd the latest `ceph-iscsi` config version is v9, so we should generate config v9 instead of v8.
Please adapt you code according to the following path to generate the correct config version - https://github.com/ricardoasmarques/lrbd-to-ceph-iscsi/commit/2ba93b4fb51a1e691305b42098b88c8bab006d0a

rjfd:
@susebot run teuthology

rjfd:
@susebot run teuthology

rjfd:
@smithfarm with the recent changes that I've pushed to this PR the iscsi suite will fail because it depends on the new ceph-iscsi package. Can you fix this somehow?

smithfarm:
> @smithfarm with the recent changes that I've pushed to this PR the iscsi suite will fail because it depends on the new ceph-iscsi package. Can you fix this somehow?

@rjfd AFAIK the set of packages used by the DeepSea PR integration CI is fixed, and ATM @kshtsk may be the only one who knows how to update it (?)

ricardoasmarques:
@rjfd `srv/salt/_modules/iscsi.py` lgtm

srv/salt/ceph/ganesha/restart/controlled/default.sls
jschmid1: this makes sense, as discussed. We should however be consistent across our codebase and adapt all restart/* to the new implementation.

rjfd: Yes, I agree

smithfarm:
Testing this requires a newer version of the ceph-iscsi RPM than is in the repos currently used by the DeepSea master PR CI.

I ran the `deepsea/tier2/igw` test manually, and it passed: http://10.86.0.135:8081/ubuntu-2019-05-10_11:08:23-deepsea:tier2:igw-ses6---basic-openstack/

srv/salt/ceph/configuration/files/rgw.conf
```diff
@@ -1,4 +1,4 @@
 [client.{{ client }}]
-rgw frontends = "beast endpoint=[::]:80"
+rgw frontends = "beast endpoint=[::]:{{ pillar.get('rgw_frontend_port', 80) }}"
```
theanalyst: `"beast port={{ pillar.get('rgw_frontend_port', 80) }}"`

In wake of https://github.com/SUSE/ceph/pull/295 we need to use port as that will bind to both ipv4 and v6

srv/salt/ceph/configuration/files/rgw-ssl.conf
```diff
@@ -1,4 +1,4 @@
 [client.{{ client }}]
-rgw frontends = "beast ssl_endpoint=[::]:443 ssl_certificate=/etc/ceph/rgw.pem"
+rgw frontends = "beast ssl_endpoint=[::]:{{ pillar.get('rgw_frontend_ssl_port', 443) }} ssl_certificate=/etc/ceph/rgw.pem"
```
theanalyst: `ssl_port={{ pillar.get('rgw_frontend_ssl_port', 443) }}`

srv/salt/ceph/configuration/files/rgw.conf
rjfd: done

srv/salt/ceph/configuration/files/rgw-ssl.conf
rjfd: done

rjfd:
@jschmid1 I think this is ready to be merged

jschmid1:
https://github.com/SUSE/DeepSea/pull/1640/files#diff-fe001a31b195bc6e5c54315ec36ee883L32 is not correct still.

We discussed adding a mechanism to query the pillar in order to judge which igw service is used.

rjfd:
> https://github.com/SUSE/DeepSea/pull/1640/files#diff-fe001a31b195bc6e5c54315ec36ee883L32 is not correct still.
> 
> We discussed adding a mechanism to query the pillar in order to judge which igw service is used.

It's fixed in this commit: https://github.com/SUSE/DeepSea/pull/1640/commits/55c0a83b4e72e601f6f7e5c7ede54ab43c8d0b0c

jschmid1:
> > https://github.com/SUSE/DeepSea/pull/1640/files#diff-fe001a31b195bc6e5c54315ec36ee883L32 is not correct still.
> > We discussed adding a mechanism to query the pillar in order to judge which igw service is used.
> 
> It's fixed in this commit: [55c0a83](https://github.com/SUSE/DeepSea/commit/55c0a83b4e72e601f6f7e5c7ede54ab43c8d0b0c)

Ah, there it is. looks good

PR #1642, opened 2019-05-07
Implement rebuild runner

swiftgist:

Description:
Serially removes and redeploys OSDs.

```
salt-run rebuild.node TARGET...
salt-run rebuild.aggressive TARGET...
salt-run rebuild.identical TARGET...
```
The difference betwen `node` and `aggressive` is an `active+clean` is performed for each OSD with node and is only performed between the nodes with aggressive.  The Ceph cluster must recover in either case before the rebuild proceeds.

One caveat is that the Ceph cluster will reach HEALTH_ERR and recover.  That is still under investigation, but no data loss occurs.

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/modules/runners/rebuild.py
+++ b/srv/modules/runners/rebuild.py
@@ -1,0 +1,257 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Runner to remove and deploy OSDs
+"""
+
+from __future__ import absolute_import
+from __future__ import print_function
+import logging
+import pprint
+import time
+# pylint: disable=import-error,3rd-party-module-not-gated,redefined-builtin
+import salt.client
+import salt.runner
+
+log = logging.getLogger(__name__)
+
+
+def help_():
+    """
+    Usage
+    """
+    usage = ('salt-run rebuild.node [TARGET...]\n'
+             'salt-run rebuild.nodes [TARGET...]:\n\n'
+             '    Removes and deploys OSDs\n'
+             'salt-run rebuild.identical [TARGET...]:\n\n'
+             '    Removes and deploys OSDs with same IDs\n'
+             'salt-run rebuild.check [TARGET...]:\n\n'
+             '    Checks available space\n'
+             '\n\n'
+             'Examples:\n'
+             '    salt-run rebuild.node data1.ceph\n'
+             '    salt-run rebuild.node data1.ceph data2.ceph\n'
+             '    salt-run rebuild.node data[12].ceph\n'
+             '    salt-run rebuild.node I@roles:storage\n'
+             '    salt-run rebuild.node data*.ceph storage*.ceph\n'
+             '\n')
+    print(usage)
+    return ""
+
+
+def master_minion():
+    """
+    Load the master modules
+    """
+    __master_opts__ = salt.config.client_config("/etc/salt/master")
+    __master_utils__ = salt.loader.utils(__master_opts__)
+    __salt_master__ = salt.loader.minion_mods(
+        __master_opts__, utils=__master_utils__)
+
+    return __salt_master__["master.minion"]()
+
+
+# pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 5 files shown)


swiftgist:
@susebot run teuthology

srv/modules/runners/rebuild.py
```diff
@@ -0,0 +1,257 @@
...
+        self.minions = self._minions(*args)
+        self.operation = kwargs.get('operation', 'osd.remove')
+        self.aggressive = kwargs.get('aggressive', False)
+        self.skipped = []
+
+    def _minions(self, *args):
+        """ Return expanded Salt target """
+        minions = set()
+        for arg in args:
+            results = self.local.cmd(arg, 'grains.get', ['id'], tgt_type="compound")
```
jschmid1: learning from #1647 we might consider using grains.get('host') instead of `id`.

in `L:84` we are checking for `minion in osd_ret` where it can be beneficial to have the short hostname.

This needs to be confirmed in a virtualized environment

srv/modules/runners/rebuild.py
```diff
@@ -0,0 +1,257 @@
...
+        """ Wait until OSDs are safe to stop """
+        delay = 3
+        while True:
+            quiescent = self.local.cmd(self.master_minion,
+                                       'osd.ceph_quiescent', [], tgt_type="compound")
+            log.debug("quiescent: {}".format(quiescent))
+            if self.master_minion in quiescent and quiescent[self.master_minion] is True:
+                break
+            print("Waiting for PGs to recover...")
+            time.sleep(delay)
```
jschmid1: This sleeps potentially forever. I think we should add a timeout (a very long though).

srv/modules/runners/rebuild.py
```diff
@@ -0,0 +1,257 @@
...
+           Verify that the storage minion is safe to rebuild
+           Remove the OSDs
+           Deploy the minion
+        """
+        if not self._disengaged() and not checkonly:
+            log.error('Safety engaged...run "salt-run disengage.safety"')
+            return ""
+
+        for minion in self.minions:
+            log.info("minion: {}".format(minion))
```
jschmid1: I think we should improve on the log.info output.

srv/modules/runners/rebuild.py
```diff
@@ -0,0 +1,257 @@
...
+            osds = self._osd_list(minion)
+            if osds:
+                log.info("osds: {}".format(osds))
```
jschmid1: here aswell (I think we should improve on the log.info output.)

jschmid1:
Also, I'd like to see the functests re-enabled. There should be all bits and pieces in place to just assemble a quick test

srv/modules/runners/rebuild.py
swiftgist: It does, but the runner is interactive and printing to the terminal.  For the complaints we had in the past regarding any timeout, this would address that.  A Ctrl-C is enough to end it when the admin is tired of waiting and finally decides to investigate.

srv/modules/runners/rebuild.py
swiftgist: The info is simply a bit more verbose of which minion is processed.

srv/modules/runners/rebuild.py
jschmid1: I'm also fine with that

srv/modules/runners/rebuild.py
swiftgist: The id vs host is not what is used.  The keys of the results are Salt minions.

srv/modules/runners/rebuild.py
jschmid1: my comment was addressing more the _format_ of the log output. reading just one word in the logs is not very informative imho.

Something like

> Processing minion: {} for foo/bar

is more pleasing to the eyes :>

PR #1644, opened 2019-05-08
By default teardown apparmor

jan--f:
SLE15 brings apparmor by default. Tear that down in a default installation.

### DIFF
--- a/srv/salt/ceph/apparmor/default-teardown.sls
+++ b/srv/salt/ceph/apparmor/default-teardown.sls
@@ -1,0 +1,13 @@
+
+aa-teardown:
+  cmd.run
+
+apparmor:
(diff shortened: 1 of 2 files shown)


Martin-Weiss:
Stage.3 does not run with apparmor enabled and running. Stopping apparmor via systemd does not do teardown.

Martin

jan--f:
> I assume everything breaks by default if we don't do this?

Thats right.

jan--f:
> Stage.3 does not run with apparmor enabled and running. Stopping apparmor via systemd does not do teardown. Martin

Hence the `aa-teardown` command.

jschmid1:
@susebot run teuthology

PR #1635, opened 2019-05-03
core: add wal/db size flag, enable limit: for wal/db devices

jschmid1:
$title

-----------------

**Checklist:**
- [x] Added unittests and or functional tests
- [x] Adapted documentation ~#TODO~ adapted the wiki
~- [ ] Referenced issues or internal bugtracker~
- [x] Ran integration tests successfully (trigger with "@susebot run teuthology" in a GitHub comment; see the [wiki](https://github.com/SUSE/DeepSea/wiki/Testing) for more information)

PR #1646, opened 2019-05-08
core: workaround ceph-volume's wal/db VG spanning issue

jschmid1:
* Work around c-v's issue described here: http://tracker.ceph.com/issues/37502
  We now create separate calls for each wal/db.

Bonus:
* Improve access time of disks(dg).c_v_commands (and successive .deploy) by 30%
  by reducing the amount of dynamic data/db/wal_devices invocations.

* Add log.debug statements in _filter_devices

* Add mechanism to prevent the listing of the same devices under different
  columns (wal/db/data).
  This is achieved by tracking the assigned devices internally.
  This also reduces the amount of checks we have to perform

Testing:

* Re-work 'inventory' generation by implementing a InventoryFactory.
* Added tests for c_v_commands

~depended on #1635 (hence https://github.com/SUSE/DeepSea/commit/611d6d98742811450f9d1f3de839a3929a993ef5)~ merged

-----------------

**Checklist:**
~- [ ] Adapted documentation~

### DIFF
--- a/srv/salt/_modules/dg.py
+++ b/srv/salt/_modules/dg.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+# pylint: disable=incompatible-py3-code
 """ This module will match disks based on applied filter rules
 
 Internally this will be called 'DriveGroups'
@@ -9,7 +10,7 @@ from __future__ import absolute_import
 import json
 import re
 import logging
-from typing import Set, Tuple
+from typing import Tuple
 
 log = logging.getLogger(__name__)
 
@@ -545,6 +546,7 @@ class DriveGroup(object):
         self._data_devices = None
         self._wal_devices = None
         self._db_devices = None
+        self.disks = self.inventory()
 
     @property
     def db_slots(self) -> dict:
@@ -623,13 +625,6 @@ class DriveGroup(object):
         return self.filter_args.get("journal_size", 0)
 
     @property
-    def inventory(self) -> dict:
-        """
-        Disks found in the inventory
-        """
-        return Inventory().disks
-
-    @property
     def data_devices(self) -> list:
         """ Filter for (bluestore/filestore) DATA devices
         """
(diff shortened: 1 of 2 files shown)


jschmid1:
@susebot run teuthology

jschmid1:
One thing I discovered:

Let's say we have a cluster with 7 disks (5HDDs and 2SSDs). We create a drive_group like this
``` yaml
<snip>
data_devices:
  rotational: 1
  limit: 4
wal_devices:
  rotational: 0
  limit: 2
</snip>
```
On the first iteration, we get a proper setup with 4 OSDs backed by 2 WALs.

On the next iteration DeepSea will create a standalone OSD for the last (7th) disk that wasn't used since limit was in place.

I don't think we should adapt the logic as it's technically a flaw in the users drive_group specification. If there is the request to specifically exclude certain disks from the deployment, we have to find a solution for it.. 

1) Add a 'exclude' flag in drive_groups
2) Add a mechanism in Deepsea to mark ignored disks (`salt-run disks.ignore $disk`)
3) Add a mechanism to ceph-volume to mark ignored disks (`ceph-volume ignore $disk`)

ideas?

tserong:
> ideas?

I'd wait for someone to make the request :-) but if we have to do this, I think I prefer adding an exclude flag in drive_groups, rather than something deeper down the stack. That way all the disk configuration is still in one place.

jschmid1:
@susebot run teuthology

srv/salt/_modules/dg.py
```diff
@@ -845,7 +900,11 @@ def deploy(**kwargs):
                 "Will not deploy using Drive-Groups."
                 "Please consult <insert doc/man> for guidance"
                 "on how to migrate to Drive-Groups")
-    return __salt__['helper.run'](c_v_commands(**kwargs))
+    c_v_command_list = c_v_commands(**kwargs)
```
jschmid1: add log.debug statements

srv/salt/_modules/dg.py
```diff
@@ -845,7 +900,11 @@ def deploy(**kwargs):
...
+    rets = []
+    for cmd in c_v_command_list:
```
jschmid1: @susebot run teuthology

jschmid1:
@susebot run teuthology

jschmid1:
failing tests are unrelated to this PR

PR #1653, opened 2019-05-10
rgw: use the port syntax

theanalyst:
Beast frontend's port will bind to both v4 and v6 addresses by default. Failure
to bind if the interface doesn't support the protocol will not be treated as an
error and we warn and continue in this case unless we have no ports to bind to.

Related to: https://github.com/SUSE/ceph/pull/295

Description:

-----------------

**Checklist:**
- [ ] Adapted documentation


rjfd:
FYI, https://github.com/SUSE/DeepSea/pull/1640/commits/9b01db331243b87160e2de07e8079aea9434b7f9

theanalyst:
Changes are incorporated in #1640, so if that merges we can drop this. I'll cross check everything and update hte thread

theanalyst:
closed in favor of #1640

PR #1650, opened 2019-05-09
Some fixes for new monitoring deployment

jan--f:
Forgot a few things.

Fixes bsc#1134467

PR #1658, opened 2019-05-13
rebuild: add functest, fix help message

swiftgist:

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/modules/runners/rebuild.py
+++ b/srv/modules/runners/rebuild.py
@@ -21,12 +21,12 @@ def help_():
     Usage
     """
     usage = ('salt-run rebuild.node [TARGET...]\n'
-             'salt-run rebuild.nodes [TARGET...]:\n\n'
-             '    Removes and deploys OSDs\n'
-             'salt-run rebuild.identical [TARGET...]:\n\n'
-             '    Removes and deploys OSDs with same IDs\n'
-             'salt-run rebuild.check [TARGET...]:\n\n'
-             '    Checks available space\n'
+             'salt-run rebuild.nodes [TARGET...]:\n'
+             '    Removes and deploys OSDs\n\n'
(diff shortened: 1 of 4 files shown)


swiftgist:
@susebot run teuthology

jschmid1:
1) linter

```
21:26:45 ************* Module srv.modules.runners.rebuild
21:26:45 W:254,17: Duplicate key 'node' in dictionary (duplicate-key)
```

2) functests

It looks like the current 'functest' testrun fails due to the missing alertmanager rpm (from #1650) 
Only the functest test fails with this error, which is rather weird..

3) Code

By the looks of it, the rebuild functest is not enabled. See https://github.com/SUSE/DeepSea/pull/1658/files#diff-9d0ae1f9554b0c101b9817e7366e68b2R15

smithfarm:
jschmid1: Is the linter failure something we can ignore?

srv/modules/runners/rebuild.py
```diff
@@ -254,4 +254,5 @@ def check(*args):
 __func_alias__ = {
                  'help_': 'help',
                  'node': 'nodes',
+                 'node': 'node',
```
smithfarm: linter caught this -> the key `'node'` is duplicated

smithfarm:
> It looks like the current 'functest' testrun fails due to the missing alertmanager rpm (from #1650)

This has since been fixed.

> Only the functest test fails with this error, which is rather weird..

`deepsea/tier2/functional` is the only test case in tier2 which has prometheus and grafana roles enabled in policy.cfg

swiftgist:
I added the .rebuild to the init.sls, but it's commented out.  When I saw the other entries commented, I thought there may be a current issue.  So, I don't see how this could even get called at the moment.

jschmid1:
rebuild is still disabled.. leaving that for later

PR #1663, opened 2019-05-16
Deploy Prometheus Alertmanager SNMP receiver

votdev:
Install the prometheus-webhook-snmp Prometheus Alertmanager reveiver if pillar ``monitoring:alertmanager:receiver:snmp:enabled`` is set to True.

To enable and configure the Prometheus Alertmanager SNMP receiver the pillar data might look like:

```
monitoring:
  alertmanager:
    receiver:
      snmp:
        enabled: True
        config:
          host: localhost
          port: 9099
          snmp_host: snmp.foo-bar.com
          snmp_community: private
          metrics: True
```

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/salt/ceph/monitoring/alertmanager/default.sls
+++ b/srv/salt/ceph/monitoring/alertmanager/default.sls
@@ -4,8 +4,43 @@ install alertmanager:
     - fire_event: True
     - refresh: True
 
+{% set receiver_snmp_enabled = salt['pillar.get']('monitoring:alertmanager_receiver_snmp:enabled', False) %}
+{% if receiver_snmp_enabled | to_bool %}
+
+prometheus_webhook_snmp_install:
+  pkg.installed:
+    - name: prometheus-webhook-snmp
+    - fire_event: True
+    - refresh: True
+
+prometheus_webhook_snmp_configure:
+  file.managed:
+    - name: "/etc/prometheus-webhook-snmp.conf"
+    - contents: |
+        {{ salt['pillar.get']('monitoring:alertmanager_receiver_snmp:config', {}) | yaml(False) | indent(8) }}
+    - user: root
+    - group: root
+    - mode: 644
+    - fire_event: True
+
+prometheus_webhook_snmp_start:
+  service.running:
+    - name: prometheus-webhook-snmp
+    - enable: True
+    - restart: True
+
+{% else %}
+
+prometheus_webhook_snmp_stop:
+  service.dead:
+    - name: prometheus-webhook-snmp
+    - enable: False
+
+{% endif %}
+
 {% set alertmanager_config = salt['pillar.get']('monitoring:alertmanager:config', '') %}
 {% if alertmanager_config == '' %}
+
 /etc/prometheus/alertmanager.yml:
   file.exists
 
--- a/doc/pillar/alertmanager_receiver_snmp.yml
+++ b/doc/pillar/alertmanager_receiver_snmp.yml
@@ -1,0 +1,9 @@
+monitoring:
+  alertmanager_receiver_snmp:
+    enabled: True
+    config:
+      host: localhost
+      port: 9099
+      snmp_host: snmp.foo-bar.com
+      snmp_community: private
+      metrics: True
--- a/srv/salt/ceph/rescind/alertmanager/default.sls
+++ b/srv/salt/ceph/rescind/alertmanager/default.sls
@@ -18,5 +18,14 @@ golang-github-prometheus-alertmanager:
 
 /etc/sysconfig/prometheus-alertmanager:
   file.absent
-{% endif %}
 
+prometheus_webhook_snmp_stop:
+  service.dead:
+    - name: prometheus-webhook-snmp
+    - enable: False
+
+prometheus_webhook_snmp_uninstall:
+  pkg.removed:
+    - name: prometheus-webhook-snmp
+    - fire_event: True
+{% endif %}


jschmid1:
retest this please

jschmid1:
@susebot run teuthology

jschmid1:
@susebot run teuthology

jschmid1:
transient issue over transient issue..

jschmid1:
@susebot run teuthology

srv/salt/ceph/monitoring/alertmanager/files/prometheus-webhook-snmp.conf.j2
```diff
@@ -0,0 +1,3 @@
+{%- for key, value in config.items() -%}
+{{ key }}: {{ value }}
```
jan--f: No need to iterate over a yaml dict to output a yaml dict.
```
{{config|yaml}}
```
will achieve the same.

srv/salt/ceph/monitoring/alertmanager/default.sls
```diff
@@ -4,8 +4,45 @@ install alertmanager:
...
+    - fire_event: True
+    - refresh: True
+
+prometheus_webhook_snmp_configure:
+  file.managed:
+    - name: "/etc/prometheus-webhook-snmp.conf"
+    - source: salt://ceph/monitoring/alertmanager/files/prometheus-webhook-snmp.conf.j2
+    - template: jinja
+    - context:
+        config: {{ salt['pillar.get']('monitoring:alertmanager:receiver:snmp:config', {}) | json }}
```
jan--f: Alternatively to having a template file, you could also use the pillar straight as the content:
```
content: |
    {{salt['pillar.get']('monitoring:alertmanager:receiver:snmp:config', {}) | json}}
```

jan--f: also any reason you use json here when the config file is in yaml?

doc/pillar/prometheus_alertmgr_snmp.yml
```diff
@@ -0,0 +1,11 @@
+monitoring:
+  alertmanager:
+    receiver:
```
jan--f: Please put this under its own key under monitoring. How about
```
monitoring:
  snmp-webhook-receiver:
    enabled: ...
```

The `monitoring:alertmanager` namespace should be reserved for configuring the alertmanager. This is about configuring the webhook recevier.

srv/salt/ceph/monitoring/alertmanager/default.sls
votdev: That was my first approach:
```
contents: {{ salt['pillar.get']('monitoring:alertmanager_receiver_snmp:config', {}) | yaml }}
```
which did not produce the expected output:
```
OrderedDict([('foo', 'bar'), ('metrics', True)])
```
Using 
```
prometheus_webhook_snmp_configure:
  file.managed:
    - name: "/etc/prometheus-webhook-snmp.conf"
    - contents: |
        {{ salt['pillar.get']('monitoring:alertmanager_receiver_snmp:config', {}) | yaml }}
```
also does not output what i want to have:
```
{foo: bar, metrics: true}
```
This looks much better, but i want to have
```
foo: bar
metrics: true
```
That's why i use the Jinja template to iterate over the key/values.

doc/pillar/prometheus_alertmgr_snmp.yml
votdev: Done

srv/salt/ceph/monitoring/alertmanager/files/prometheus-webhook-snmp.conf.j2
votdev: The output will be
```
{metrics: true, foo: bar}
```
but should be
```
metrics: true
foo: bar
```

srv/salt/ceph/monitoring/alertmanager/default.sls
jan--f: sigh...you right. I guess it would have been surprising if something was easy in salt.
The `yaml` filter does work but only with the correct indentation. Like so:
```
- contents: |
    {{ salt['pillar.get']('monitoring:alertmanager:receiver:snmp:config', {}) | yaml(False) | indent(8) }}
```
The value indent gets depends on the indent level in the file. I think 8 should be right in `srv/salt/ceph/monitoring/alertmanager/default.sls`

jan--f: eh I think that might work but only for this case...feel free to ignore while I quietly hate salt.

srv/salt/ceph/monitoring/alertmanager/files/prometheus-webhook-snmp.conf.j2
jan--f: yeah fair...my mistake. _Should_ work though :wink:

srv/salt/ceph/monitoring/alertmanager/default.sls
votdev: Fixed. I've chosen this solution because it requires one file less.

jschmid1:
retest this please

jschmid1:
everyone happy with the current version of this patch? @votdev @jan--f ?

jschmid1:
retest this please

PR #1665, opened 2019-05-16
dg: Add parameter to bypass the pillar_check of dg.py

jschmid1:
This parameter is useful for using disks.report and a forceful
disks.deploy.

The command is already mentioned in https://bugzilla.suse.com/show_bug.cgi?id=1135340

-----------------

**Checklist:**
- [x] Adapted documentation

### DIFF
--- a/srv/salt/_modules/dg.py
+++ b/srv/salt/_modules/dg.py
@@ -961,7 +961,8 @@ def deploy(**kwargs):
     # do not run this when there is still ceph:storage in the pillar
     # this indicates that we are in a post-upgrade scenario and
     # the drive assignment was not ported to drive-groups yet.
-    if __pillar__.get('ceph', {}).get('storage'):
+    if not kwargs.get('bypass_pillar', False) and ('storage' in __pillar__.get(
+            'ceph', {})):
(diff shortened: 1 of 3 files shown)


jschmid1:
@susebot run teuthology

srv/salt/_modules/dg.py
```diff
@@ -961,7 +961,8 @@ def deploy(**kwargs):
     # do not run this when there is still ceph:storage in the pillar
     # this indicates that we are in a post-upgrade scenario and
     # the drive assignment was not ported to drive-groups yet.
-    if __pillar__.get('ceph', {}).get('storage'):
+    if __pillar__.get('ceph', {}).get('storage') and not kwargs.get(
```
rjfd: shouldn't you write condition like this: `not kwargs.get('bypass_pillar', False) and ('storage' in __pillar__.get('ceph', {}))`? This way you will only check the pillar if `bypass_pillar is False` and will avoid a `KeyError` exception when 'storage' is not a key in the `__pillar__['ceph']`

jschmid1: right, that's better.

PR #1667, opened 2019-05-17
osd: add independent osd discovery method

jschmid1:
This is a utility function for the profile->drivegroup migration.

-----------------

**Checklist:**

### DIFF
--- a/srv/modules/runners/disks.py
+++ b/srv/modules/runners/disks.py
@@ -244,6 +244,33 @@ def report(**kwargs):
     return DriveGroups(**kwargs).call_out('deploy', alias='report')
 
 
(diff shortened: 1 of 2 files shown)


jschmid1:
> Maybe a comment to the new OSD classe which files are actually read? This part seems a bit nebulous to me, though I assume its the OSD metadata under `/var/lib/ceph/osd/osd-<id>/...`. Otherwise LGTM

you're correct with that assumption. I'll add a comment though

jschmid1:
Added a docstring for `__init__`

jschmid1:
retest this please

jschmid1:
@susebot run teuthology

Issue #1590, opened 2019-03-27
ceph-mgr apparmor profile is broken


tserong:
Set apparmor to enforcing mode (i.e. apply ceph.apparmor.default-enforce.sls), restart ceph-mgr, all the "always on" modules will be broken, and no other modules will be loadable. This will immediately manifest as:

```
# ceph status
  cluster:
    id:     ff3e2f55-0f55-4d87-8a5a-752b79b68b33
    health: HEALTH_ERR
            7 mgr modules have failed
```

The following gets things partially running again:

```
diff --git a/srv/salt/ceph/apparmor/files/usr.bin.ceph-mgr b/srv/salt/ceph/apparmor/files/usr.bin.ceph-mgr
index 1910df2f..c5d6405c 100644
--- a/srv/salt/ceph/apparmor/files/usr.bin.ceph-mgr
+++ b/srv/salt/ceph/apparmor/files/usr.bin.ceph-mgr
@@ -36,6 +36,7 @@
   /proc/*/fd** r,
   /proc/*/mounts r,
 
-  /usr/lib64/ceph/mgr/** rw,
+  /usr/share/ceph/mgr/ rw,
+  /usr/share/ceph/mgr/** rw,
 
 }
```

...but is not quite enough, because then we get to:

```
  cluster:
    id:     158a34a6-1807-4858-af38-4e4d8f9c3962
    health: HEALTH_ERR
            Module 'dashboard' has failed: [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/']
```

There's loads of failures in /var/log/audit/audit.log at this point (/tmp, /dev/shm/, ...), so this probably needs some detailed review.

Presumably this wasn't noticed previously because the apparmor functests set enforcing mode, then check the cluster status. There's a bunch of stuff ceph-mgr (and, presumably the other daemons) do on startup, which is prevented by the current profile, but of course if the cluster is already running when you set enforcing mode, those things have already happened. The apparmor functest should be updated to set enforcing mode, then restart the entire cluster, then check the cluster status (see also #1589)

PR #1675, opened 2019-05-28
apparmor: fix mgr profile after nautilus changes

jan--f:
Fixes: bsc#1130630
Fixes: #1590

Also removes the oA profile.


jschmid1:
@susebot run teuthology

Issue #284, opened 2017-05-24
status.report unit tests


swiftgist:
The status report runner works fine, but the unit tests seem to randomly succeed/fail after successive runs.

Here is an example of the output from test_report_6_of_9 from four separate runs with no changes:

```
tests/unit/runners/test_status.py::TestStatusReport::test_report_6_of_9   salt: 2015.0.1
  os: SUSE Linux Enterprise Server 12 SP2
  ceph: ceph version 10.0.2-269-g9148e53 (9148e530e27266064881e628ccadecb4975b0904)

  mon1.ceph:
    os: SUSE Linux Enterprise Server 12 SP3
  data5.ceph:
    os: SUSE Linux Enterprise Server 12 SP3
    salt: 2016.11.4
    ceph: ceph version 12.0.2-269-g9148e53 (9148e530e27266064877e632ccadecb4979b0904)
  data3.ceph:
    os: SUSE Linux Enterprise Server 12 SP3
    salt: 2016.11.4
    ceph: ceph version 12.0.2-269-g9148e53 (9148e530e27266064877e632ccadecb4979b0904)
  data2.ceph:
    salt: 2016.11.4
    ceph: ceph version 12.0.2-269-g9148e53 (9148e530e27266064877e632ccadecb4979b0904)
FAILED


ests/unit/runners/test_status.py::TestStatusReport::test_report_6_of_9   salt: 2015.0.1
  ceph: ceph version 10.0.2-269-g9148e53 (9148e530e27266064881e628ccadecb4975b0904)
  os: SUSE Linux Enterprise Server 12 SP2

  admin.ceph:
    os: SUSE Linux Enterprise Server 12 SP3
    ceph: ceph version 12.0.2-269-g9148e53 (9148e530e27266064877e632ccadecb4979b0904)
    salt: 2016.11.4
  mon1.ceph:
    os: SUSE Linux Enterprise Server 12 SP3
    ceph: ceph version 12.0.2-269-g9148e53 (9148e530e27266064877e632ccadecb4979b0904)
    salt: 2016.11.4
  mon2.ceph:
    os: SUSE Linux Enterprise Server 12 SP3
    ceph: ceph version 12.0.2-269-g9148e53 (9148e530e27266064877e632ccadecb4979b0904)
    salt: 2016.11.4
PASSED

tests/unit/runners/test_status.py::TestStatusReport::test_report_6_of_9   salt: 2015.0.1
  os: SUSE Linux Enterprise Server 12 SP2
[...]

PR #1684, opened 2019-06-06
re-enable TestStatusReport tests

azraelxyz:
These unit tests has two issues:
1. python3 compatibility
2. the unit tests seem to randomly fail after successive runs.

This commit fixed python3 compatibility issue using six module.
Also, the second issue for successive runs are tested by the
following script.

  for i in `seq 100`;
    do echo $i;
    make test;
    if [ $? -ne 0 ]
    then
      exit 1
    fi
  done

This commit also introduced parameterized test framework to
reduce redundant code.

Fixes #
https://github.com/SUSE/DeepSea/issues/284

Description:

-----------------

**Checklist:**
- [*] Added unittests and or functional tests
- [] Adapted documentation
- [*] Referenced issues or internal bugtracker

### DIFF
--- a/tests/unit/runners/test_status.py
+++ b/tests/unit/runners/test_status.py
@@ -2,6 +2,7 @@ from mock import patch
 import pytest
 from srv.modules.runners import status
 from tests.unit.helper.output import OutputHelper
+import six
 
 
 class TestStatusReport():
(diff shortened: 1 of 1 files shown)


jschmid1:
Great, thanks for the fix!

PR #1693, opened 2019-06-20
SES5: fix mds numerical names

jan--f:
This fixes the mds names for new deployments. daemon names that would start with a digit (i.e. when a hostname starts with a digit) get a `mds.` prefix.
This also adds an orchestration to migrate from the invalid names to the new naming scheme. Starting the migration with no invalid names present should not change anything, i.e. be safe to run at any time with an mds cluster present.

### DIFF
--- a/srv/salt/ceph/mds/migrate-numerical-names.sls
+++ b/srv/salt/ceph/mds/migrate-numerical-names.sls
@@ -1,0 +1,55 @@
+{% set master = salt['pillar.get']('master_minion) %}
+
+{% if salt.saltutil.runner('select.minions', cluster='ceph', roles='mds') %}
+
+{% if salt['saltutil.runner']('validate.discovery', cluster='ceph') == False %}
+
+validate failed:
+  salt.state:
+    - name: just.exit
+    - tgt: {{ master }}
+    - failhard: True
+
+{% endif %}
+
+refresh_pillar1:
+  salt.state:
+    - tgt: '{{ salt['pillar.get']('deepsea_minions') }}'
+    - tgt_type: compound
+    - sls: ceph.refresh
+
+create renamed mds keys:
+  salt.state:
+    - tgt: {{ master }}
+    - tgt_type: compound
+    - sls: ceph.mds.key
+    - failhard: True
+
+push mds keys:
+  salt.state:
+    - tgt: {{ master }}
+    - tgt_type: compound
+    - sls: ceph.mds.key
+    - failhard: True
+
+auth new keys:
+  salt.state:
+    - tgt: {{ master }}
+    - sls: ceph.mds.auth
+    - failhard: True
+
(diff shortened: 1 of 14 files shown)


jan--f:
This is partly the backport for #1694 witht he addition of the migration orchestration.

srv/salt/ceph/mds/migrate-numerical-names.sls
```diff
@@ -0,0 +1,55 @@
...
+    - failhard: True
+
+remove old mds:
+  salt.state:
+    - tgt: "I@roles:mds and I@cluster:ceph"
+    - tgt_type: compound
+    - sls: ceph.mds.remove-mds-with-numeric-names
+
+{% endif %}
+
```
tserong: Does this need a test.nop thingy in case there's no MDS at all?

No, wait, ignore me, nobody will run this if they don't have MDSes.

srv/salt/ceph/mds/remove-mds-with-numeric-names.sls
```diff
@@ -0,0 +1,19 @@
...
+{% set name = grains['host'] %}
+{% if name != salt['mds.get_name'](name) %}
+stop mds {{ name }}:
+  service.dead:
+    - name: ceph-mds@{{ name }}
+    - enable: False
+
+stop mds:
+  service.dead:
+    - name: ceph-mds@mds
```
tserong: What's this stanza for?

srv/salt/ceph/mds/shutdown.sls
```diff
@@ -1,3 +1,3 @@
 shutdown daemon:
   service.dead:
-    - name: ceph-mds@{{ grains['host'] }}
+    - name: ceph-mds@mds.{{ grains['host'] }}
```
tserong: Shouldn't this be using `mds.get_name`? (same comment as in #1694)

srv/salt/ceph/mds/migrate-numerical-names.sls
jan--f: and if they do it'll fail. I think this is acceptable, as this is not run automatically.

srv/salt/ceph/mds/shutdown.sls
jan--f: Right, good catch. I also fixed the restart states.

srv/salt/ceph/mds/remove-mds-with-numeric-names.sls
jan--f: Left over from copy-pasting.

jan--f:
I have a documentation bug open. However we should note in the update release news or somewhere prominent, that users with problematic mds names need to run this first thing after updating DeepSea. See bsc#1138804 for doc input.

PR #1694, opened 2019-06-20
mds: numerical mds names are forbidden (have been deprecated since 2014)

jan--f:
This adds a "mds." prefix the the daemon name in case a mds hostname starts
with a digit.

Fixes: bsc#1138442

This fixes the mds names for new deployments. daemon names that would start with a digit (i.e. when a hostname starts with a digit) get a `mds.` prefix.

### DIFF
--- a/srv/salt/ceph/mds/shutdown.sls
+++ b/srv/salt/ceph/mds/shutdown.sls
@@ -1,3 +1,5 @@
+{% set name = salt['mds.get_name'](grains['host']) %}
+
 shutdown daemon:
   service.dead:
-    - name: ceph-mds@{{ grains['host'] }}
+    - name: ceph-mds@{{ name }}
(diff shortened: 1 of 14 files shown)


jan--f:
The backport #1693 also adds the orchestration to rename MDS daemons. The backport needs to be shipped asap as users that are impacted (have MDS names starting with a digit) will have to run this before upgrading. Doc bug will follow.

srv/salt/ceph/mds/shutdown.sls
```diff
@@ -1,3 +1,3 @@
 shutdown daemon:
   service.dead:
-    - name: ceph-mds@{{ grains['host'] }}
+    - name: ceph-mds@mds.{{ grains['host'] }}
```
tserong: Shouldn't this be using `mds.get_name`?

jan--f: Right, good catch. I also fixed the restart states.

jschmid1:
@jan--f please fix linter complaints. besides that - ready to merge

jan--f:
retest this please

jan--f:
retest this please

PR #1701, opened 2019-07-03
Fix subvolume mount for generic device or label

swiftgist:
bnc:#1139379
-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/salt/_modules/subvolume.py
+++ b/srv/salt/_modules/subvolume.py
@@ -53,3 +53,16 @@ def _subvol():
         if line.endswith("/var/lib/ceph"):
             found = True
     return found
+
+
+def device():
+    """
+    Print the device (i.e. first column) of the root filesystem for btrfs
+    """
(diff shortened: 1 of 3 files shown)


jschmid1:
Having the `awk` expr inside the .sls files makes it really hard to verify if it's correct. Could we offload this to a module and have this unit-tested?

swiftgist:
I added a function, but I did think it was overkill for "print first column for the root filesystem when it's btrfs".

jschmid1:
> I added a function, but I did think it was overkill for "print first column for the root filesystem when it's btrfs".

Reads may better, imo. Thanks!

jschmid1:
We received a request to include this in SES6. Would you mind creating a backport of this @swiftgist ?

swiftgist:
It's already there in commit e673363dfca4707b2a704a98622fddefa6c48dd3

PR #1690, opened 2019-06-18
disks: improve `disks.list` and various other things

jschmid1:
This PR does quite a bit:

* remove the (standalone) discovery functionality (merged with `.list`)
   This was confusing a only used in a migration context
* Move disk gathering logic to cephdisks.py (where it belongs)
* exclude rbd, sr[0-9], and root disks from the output cephdisks.py
* re-arrange public methods of cephdisks.py
* Fix bsc#1137587 (by excluding rbd,sr devices _and_ checking for key presence)
* Disk discovery for LVM and NON-LVM disks
* Return meaningful error messages when users configured weird/wrong drivegroups bsc#1135179
* Generate annotated return (default with `disks.list`) that makes it easier to identify disks and associated OSDs and about to be created OSDs
* add and adapt tests (There could be more, admittedly)

<details><summary> Example outputs </summary> 
<p>

```
admin:/deepsea # salt-run disks.list                    
Found DriveGroup <default>                                                                                                              
Calling dg.report on compound target I@roles:storage                                                                                                                                                  
|_                                                                                                                              
  ----------                                                                                                                                                                                          
  data1:                                                                                                                        
      ----------                                                                                                 
      data_devices:                                                                                                             
          ----------                                                                                                                           
          /dev/vdb:                                                                                                                                            
              ----------                                                                                                                              
              osds:                                                                                                                                                   
                  - 1                                                                                         
          /dev/vdc:                                                                                                                     
              ----------                                                                                                                                
              osds:                                                                                                                     
                  - 8                                                                                     
          /dev/vdd:                                                                                                                                                                                                  
              ----------                                                                                                                                                                                        
              osds:                                                                                                                                     
                  - 2                                                                                                                   
          /dev/vde:                                                                                                                                     
              ----------                                                                                                                
              osds:                                                                                                                                     
                  - 15                                                                                                                  
          /dev/vdf:                                                                                                                                     
              ----------                                                                                                                                
              osds:                                                                                                             
                  - 17                                                                                           
          /dev/vdg:                                                                                                             
              ----------                                                                                         
              osds:                                                                                
                  - 1
          /dev/vdh:
              ----------
              message:
                  No OSD detected (Will be created in the next deployment run)
      db_devices:
          ----------
      errors:
          ----------
      journal_devices:
          ----------
      wal_devices:
          ----------
[..snip]
```

</p>
</details>
<details><summary> Or the output when the DriveGroups change (in case of a rebuild) </summary>

<p>

```

[...]

PR #1715, opened 2019-07-26
remove/replace: add path to status report for better identification

jschmid1:

before:

`salt-run osd.remove(replace) $osd_id`
[..snip..]     
```                                                                                                                                                                                                                                                                                                                                              
14:                                                                                                                                                                                                                                            
    True 
```

after:

```
14:                                                                                                                                                                                      
    ----------                                                                                                                                                                                                                                 
    path:                                                                                                                                                                                                                                      
        /dev/vdb                                                                                                                                                                                                                               
    returncode:                                                                                                                                                                                                                                
        True               
```

suse_internal: bnc#1142663

-----------------

**Checklist:**
~- [ ] Added unittests and or functional tests~
~- [ ] Adapted documentation~
~- [ ] Referenced issues or internal bugtracker~

### DIFF
--- a/srv/modules/runners/osd.py
+++ b/srv/modules/runners/osd.py
@@ -71,6 +71,20 @@ class Util(object):
 
         return __salt_master__["master.minion"]()
 
+    def find_by_osd_id(self, target: str, osd_id: int) -> dict:
+        """ Find by osd_id and return details about disk """
+        osd_metadata: dict = self.local.cmd(
+            target, 'cephdisks.find_by_osd_id', [osd_id],
+            tgt_type="compound")[target]
+        if isinstance(osd_metadata, str):
+            if osd_metadata.startswith(
+                    'The minion function caused an exception'):
+                log.error(osd_metadata)
+                return dict()
+        if isinstance(osd_metadata, list):
+            return osd_metadata[0]
+        return dict()
+
 
 class OSDUtil(Util):
     """ Util class for OSD handling """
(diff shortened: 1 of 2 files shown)


jschmid1:
@susebot run teuthology

swiftgist:
Is this for bnc#1142663?  It's the request for the model and serial number.

jschmid1:
> Is this for bnc#1142663? It's the request for the model and serial number.

You're right. I wasn't aware that there is a bug for this. Adding to the PR body.

tserong:
Shouldn't this then print the model and serial number in addition to the device path?

jschmid1:
> Shouldn't this then print the model and serial number in addition to the device path?

while the user could technically retrieve this by searching for the device path, this would probably save some time :+1: 

In a perfect world we should even hook a `disk.led` call to a `replace/remove` call. Would that be useful right now?

tserong:
> In a perfect world we should even hook a `disk.led` call to a `replace/remove` call. Would that be useful right now?

Maybe? Although I'm kinda inclined to leave that until someone explicitly asks for it to happen automatically :-)

jschmid1:
> > In a perfect world we should even hook a `disk.led` call to a `replace/remove` call. Would that be useful right now?
> 
> Maybe? Although I'm kinda inclined to leave that until someone explicitly asks for it to happen automatically :-)

adapted the return to: 

* path
* returncode
* model


`serial` is postponed until present in `ceph-volume inventory` json output, as discussed.

jschmid1:
Added a fix dc5e1ef for a bug that was introduced with #1690

Issue #842, opened 2017-12-04
stop setting rgw dns name


theanalyst:
We set the `rgw dns name` configurable as the node's fqdn. However most deployments seem to have multiple rgw endpoints and the fqdn not set up correctly in the node to match the dns name. Without this set, all of the S3 and swift apis should still work correctly except when a bucket style access is used which s3 clients like s3cmd default to. But for these clients even after an fqdn setting they would require dnsmasq configured in the client side correctly, so people making this deployment step anyway cannot just use the default out of the box yet, so we can probably drop this from the default?

swiftgist:
Should this just become an alternate default?  Then, it would serve as an example if somebody is experimenting with bucket style access with one client.

theanalyst:
It could be an alternate default. Don't want to have it as the default default

swiftgist:
Recently ran across setups that use load balancers and want a static `rgw dns name`.  While supported, the .../ceph.conf.d/rgw.conf is technically unique in needing a label.  All other *.conf sections have the label in the ceph.conf.j2 (e.g. mon, osd, mgr, etc.)

This may need an improvement in the README, a default-loadbalancer example or both.

PR #1717, opened 2019-07-31
conf: rgw: drop rgw dns name configurable

theanalyst:
This is only useful for s3 clients which don't support path style access, and
even in those cases there needs to be an additional dnsmasq setting or the like
that can correctly redirect `<bucketname>.<rgw-endpoint>` to `<rgw-endpoint>`.
For the remainder of use cases; if this is configured incorrectly it always leads
to bugs for all s3 clients and it is hard to get this right, as many rgws may be
terminated with a load balancer in which case we need the endpoint to be that.
So moving forward, we'd make this option manually configurable for those sites
that really need the vhost style access, as this requires some knowledge on the
s3 clients and their site dns setting which is hard to predict correctly anyway

Fixes: #842

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/salt/ceph/configuration/files/rgw-ssl.conf
+++ b/srv/salt/ceph/configuration/files/rgw-ssl.conf
@@ -1,4 +1,3 @@
 [client.{{ client }}]
 rgw frontends = "beast ssl_port={{ pillar.get('rgw_frontend_ssl_port', 443) }} ssl_certificate=/etc/ceph/rgw.pem"
-rgw dns name = {{ fqdn }}
 rgw enable usage log = true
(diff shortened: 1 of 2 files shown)


theanalyst:
I'll see what needs to be done on the testing front here, but otherwise should be good to go

jschmid1:
> I'll see what needs to be done on the testing front here, but otherwise should be good to go

any news here :)?

theanalyst:
this should be ready to go, i was just making sure that deepsea ci tests pass for this

jschmid1:
@susebot run teuthology

PR #1724, opened 2019-08-16
Accommodate Salt integration tests

swiftgist:
I hate changing code to accommodate tests, but I don't see another path
for this one.  The structure of a Salt orchestration file provides `arg`
as a keyword to the runner internally.  Changing from using `osd_list`
to `args` broke the tests, but matches the user requirement.

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/modules/runners/remove.py
+++ b/srv/modules/runners/remove.py
@@ -26,6 +26,16 @@ def osd(*args, **kwargs):
     Remove an OSD gracefully or forcefully.  Always attempt to remove
     ID from Ceph even if OSD has been removed from the minion.
     """
+    # Specifically for Salt integration testing
+    # The 'arg' keyword is passed as a kwarg to the salt.runner
+    # See delay.sls and multiple.sls for examples
+    if not args:
+        if 'arg' in kwargs and kwargs['arg']:
+            args = kwargs['arg']
+
+    if not args:
+        help()
+        return ""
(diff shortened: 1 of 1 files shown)


swiftgist:
@susebot run teuthology

srv/modules/runners/remove.py
```diff
@@ -36,6 +36,13 @@ def osd(*args, **kwargs):
 
     local = salt.client.LocalClient()
 
+    # Specifically for Salt integration testing
+    # The 'arg' keyword is passed as a kwarg to the salt.runner
+    # See delay.sls and multiple.sls for examples
+    if not args:
```
jschmid1: In a non-testing environment, wouldn't this mean

`salt-run remove.osd`  # <- without passing any OSDs?

I'd assume this is here we should drop a help message

smithfarm:
I guess there is a good reason why this test is failing in SES5 only, and not in master, and I'm just not aware of it.

srv/modules/runners/remove.py
swiftgist: Addressed... had to move the Salt check up as well.

swiftgist:
@susebot run teuthology

swiftgist:
> I guess there is a good reason why this test is failing in SES5 only, and not in master, and I'm just not aware of it.

The runners aren't the same.

PR #1726, opened 2019-08-20
[next] Simpler configuration

swiftgist:
The original configuration uses stack.py and supports individual minion
customizations.  With a default and user pillar directory tree, this led
to confusion for many more than once.

This strategy abandons using stack.py and the custom trees.  Instead of
file globbing, values of policy.cfg become Salt targets.  The roles are
generated the same without the intermittent files.  Lastly, the
global.yml moves out of the stack directory and is only initialized.
The global.yml is still a user maintained file.

Fixes #

Description:

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/modules/runners/config.py
+++ b/srv/modules/runners/config.py
@@ -1,0 +1,282 @@
+# -*- coding: utf-8 -*-
+'''
+Salt runner for reading the policy.cfg, creating the role assignments and
+global configuration files.
+'''
+from __future__ import absolute_import
+import errno
+import os
+import logging
+import pprint
+import uuid
+import sys
+import yaml
+
+import salt.key
+import salt.client
+import salt.runner
+from ext_lib.utils import master_minion, evaluate_state_return
+from ext_lib.network import DeepSeaNetwork
+
+log = logging.getLogger(__name__)
+
+MINIONS_DIR = "/srv/pillar/ceph/minions"
+POLICY = "/srv/pillar/ceph/proposals/policy.cfg"
+GLOBAL_YML = "/srv/pillar/ceph/global.yml"
+CUSTOM_YML = "/srv/pillar/ceph/custom.yml"
+
+
+class DeepSeaRoles(object):
+    ''' Drives creation of /srv/pillar/ceph/minions '''
+
+    def __init__(self, roles, minions_dir=MINIONS_DIR):
+        self.dir = minions_dir
+        self.roles = roles
+        self.error = ""
+        self.minions = {}
+        self.dumper = yaml.SafeDumper
+        self.dumper.ignore_aliases = lambda self, data: True
+
+    def invert(self):
+        ''' Swaps keys and values '''
+        for key, value in self.roles.items():
+            for item in value:
+                self.minions.setdefault(item, []).append(key)
+        log.info(f"Minions:\n{pprint.pformat(self.minions)}")
+
+    def write(self):
+        ''' Save each minion configuration '''
+        if self._mkdir(self.dir):
+            for minion in self.minions:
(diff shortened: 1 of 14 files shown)


swiftgist:
If everyone agrees on the change to policy.cfg and this overall direction, I'll look to complete the calls to the network parts of the DeepSeaGlobal class.

swiftgist:
An example policy.cfg

```
role-master: admin*
role-admin: mon*
role-igw: igw*
role-rgw: rgw*
role-mon: mon*
role-mds: mds*
role-mgr: mon*
rgw:
  role-red: rgw*
  role-blue: rgw*
ganesha:
  role-alpha: ganesha1*
  role-beta: ganesha2*

role-client-cephfs: client1*
role-client-radosgw: client1*
role-client-iscsi: client1*
role-client-nfs: client1*

role-benchmark-rbd: client2*
role-benchmark-blockdev: client2*
role-benchmark-fs: client2*
```

jschmid1:
> If everyone agrees on the change to policy.cfg and this overall direction, I'll look to complete the calls to the network parts of the DeepSeaGlobal class.

This sounds like the right direction to me!

srv/modules/runners/config.py
```diff
@@ -0,0 +1,186 @@
...
+                filename = f"{self.dir}/{minion}.sls"
+
+                contents = {'roles': self._lstrip(self.minions[minion])}
+                log.info(f"Writing {filename}")
+                log.info(f"Contents {contents}")
+                with open(filename, "w") as yml:
+                    yml.write(yaml.dump(contents, Dumper=self.dumper,
+                              default_flow_style=False))
+            return True
+        return False
```
mgfritch: Invert the `if` statement for readability?

```
    def write(self):
        '''  '''
        if not self._mkdir(self.dir):
            return False

        for minion in self.minions:
            filename = f"{self.dir}/{minion}.sls"

            contents = {'roles': self._lstrip(self.minions[minion])}
            log.info(f"Writing {filename}")
            log.info(f"Contents {contents}")
            with open(filename, "w") as yml:
                yml.write(yaml.dump(contents, Dumper=self.dumper,
                          default_flow_style=False))
        return True
```

tserong:
> This sounds like the right direction to me!

Likewise :-)

srv/modules/runners/ext_lib/utils.py
```diff
@@ -1,4 +1,4 @@
-from salt.client import LocalClient
+from salt.client import LocalClient, salt
```
jschmid1: the bare salt import is not needed, right? Let's be explicit about our imports

srv/modules/runners/config.py
```diff
@@ -0,0 +1,186 @@
...
+
+log = logging.getLogger(__name__)
+
+class DeepSeaRoles(object):
+    '''
+    Drives creation of /srv/pillar/ceph/minions and /srv/pillar/ceph/global.yml
+    '''
+
+    def __init__(self, roles):
+        self.dir = "/srv/pillar/ceph/minions"
```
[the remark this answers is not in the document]
jschmid1: I also think that removing 'cluster' is a good thing. +1 for `..ceph/minions`

jschmid1:
besides the linter complaints, shall we merge this and #1728 together?

swiftgist:
Fine with me... the last couple of linter complaints I was undecided on.

jschmid1:
I fear we have to rebase here

swiftgist:
okay.... give me a bit... still resolving some other issues.

swiftgist:
rebased... we need to chat about Salt config vs. Ceph config

PR #1732, opened 2019-08-30
Fixup rpmbuild in Makefile

mgfritch:
Description:

The 'make tarball' and 'make rpm' targets were failing in various ways when run as user 'root' and also when run as a 'non-root' user.

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 setup.py
-deepsea.spec
 
 *.sw[pon]
 *~
@@ -35,7 +34,6 @@ wheels/
 #  Usually these files are written by a python script from a template
 #  before PyInstaller builds the exe, so as to inject date/other infos into it.
 *.manifest
-*.spec
 
 # Installer logs
 pip-log.txt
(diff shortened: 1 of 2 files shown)


jschmid1:
adding @smithfarm for Make/spec file sorcery

smithfarm:
This PR doesn't change the spec file, so nothing for me to review, but I can say the following:

I vaguely recall that "make rpm" did not work when run as root.

The proof is in the pudding - the deepsea-make-rpm Jenkins test passed. We should also run teuthology on this, since (IIRC) the Jenkins job which runs teuthology first builds RPMs using "make rpm".

kshtsk:
Yes I think running teuthology is reasonable.

mgfritch:
@susebot run teuthology

jschmid1:
we need a rebase here, but I **think** this looks reasonable. I can't 100% tell if the changes will cause any inconvenience due to my lack of knowledge in the spec/make land :/

mgfritch:
@susebot run teuthology

PR #1737, opened 2019-09-04
Adapt logic for global.conf changes

jschmid1:

* changed logic for dependency resolution
* global.conf changes now result in core service restart(flags) (mon, mgr, osd)

~Ignore whitespace and indentation fixes. [auto]~

suse_internal: bsc#1149400

-----------------

**Checklist:**
~- [ ] Added unittests and or functional tests~
~- [ ] Adapted documentation~

### DIFF
--- a/srv/modules/runners/changed.py
+++ b/srv/modules/runners/changed.py
@@ -39,7 +39,6 @@ class Role(object):
         self._conf_files = [self.conf_dir + self.conf_filename + self.conf_extension]
         self._depends = []
         self._set_depends()
-        self.rgw_configurations()
 
     @property
     def name(self):
@@ -97,7 +96,7 @@ class Role(object):
             self.add_dependencies(Role(role_name='mon'))
             self.add_dependencies(Role(role_name='mgr'))
             self.add_dependencies(Role(role_name='storage'))
-            # also stage restarts on gateways (igw is not maintained here)
+            # also stage restarts on gateways (igw and genesha is not maintained here)
             self.add_dependencies(Role(role_name='mds'))
             self.add_dependencies(Role(role_name='rgw'))
         else:
@@ -121,28 +120,6 @@ class Role(object):
         DEV/DEBUG
         """
         return [dep.name for dep in self.dependencies]
-
-    def rgw_configurations(self):
-        """
-        RadosGW allows custom configurations.  Include these roles with a
-        dependency on the global.conf.  Default to 'rgw' if not set.
-        """
-        # pylint: disable=redefined-outer-name
-        local = salt.client.LocalClient()
-        roles = []
-        try:
-            roles = list(local.cmd("I@roles:master", 'pillar.get',
-                                   ['rgw_configurations'],
-                                   tgt_type="compound").values())[0]
-            log.debug("Querying pillar for rgw_configurations")
-        # pylint: disable=bare-except
-        except:
-            pass
-        if not roles:
-            roles = ['rgw']
-        for role in roles:
-            if role == self.name:
-                self.add_dependencies(Role(role_name='global'))
 
 
 class Config(object):
@@ -265,8 +242,7 @@ def requires_conf_change(**kwargs):
                 tgt_type="compound")
             log.info(f"Set restart grain for role:{dep.name}")
         return True
-    else:
-        return False
+    return False
 
 
 def rgw():
--- a/srv/modules/runners/changed.py
+++ b/srv/modules/runners/changed.py
@@ -37,7 +37,8 @@ class Role(object):
         self.conf_filename = kwargs.get('conf_filename', self._role_name)
         self.conf_extension = kwargs.get('conf_extension', '.conf')
         self._conf_files = [self.conf_dir + self.conf_filename + self.conf_extension]
-        self._depends = [self]
+        self._depends = []
+        self._set_depends()
         self.rgw_configurations()
 
     @property
@@ -73,6 +74,34 @@ class Role(object):
         Returns dependency list
         """
         return self._depends
+
+    @property
+    def depends(self):
+        """
+        Getter for name attr
+        """
+        return self._depends
+
+    @depends.setter
+    def depends(self, role):
+        """
+        Setter for name attr
+        """
+        self._depends = role
+
+    def _set_depends(self):
+        """ Set dependency to self if not global """
+        if self._role_name == 'global':
+            # if the global config changed
+            # stage restart on the core services
+            self.add_dependencies(Role(role_name='mon'))
+            self.add_dependencies(Role(role_name='mgr'))
+            self.add_dependencies(Role(role_name='storage'))
+            # also stage restarts on gateways (igw is not maintained here)
+            self.add_dependencies(Role(role_name='mds'))
+            self.add_dependencies(Role(role_name='rgw'))
+        else:
+            self.depends = [self]
 
     def add_dependencies(self, role):
         """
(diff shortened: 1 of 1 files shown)


srv/modules/runners/changed.py
```diff
@@ -74,6 +78,31 @@ def dependencies(self):
...
+        self._depends = role
+
+    def _set_depends(self):
+        """ Set dependency to self if not global """
+        if self._role_name == 'global':
+            # if the global config changed
+            # stage restart on the core services.
+            self.add_dependencies(Role(role_name='mon'))
+            self.add_dependencies(Role(role_name='mgr'))
+            self.add_dependencies(Role(role_name='storage'))
```
jan--f: `mds` too?

jschmid1: I limited the restarts to the core services, since the mds should have it's own section (mds.conf)

swiftgist: While that's true, I don't know if any of us know of all the Ceph variables that could impact both mds and another component.  If an admin updates the global.conf not quite understanding that it's more suited for mds.conf, I would still expect DeepSea to restart everything.  Currently, the mon, mgr and osds will restart, but not the mds.  So, the admin will still have suffer the penalty of the serial restart, but still have to investigate what went wrong.

srv/modules/runners/changed.py
```diff
@@ -102,9 +131,11 @@ def rgw_configurations(self):
...
         roles = []
         try:
-            roles = list(local.cmd("I@roles:master", 'pillar.get',
-                                   ['rgw_configurations'],
-                                   tgt_type="compound").values())[0]
+            roles = list(
+                local.cmd(
+                    "I@roles:master",
+                    'pillar.get', ['rgw_configurations'],
+                    tgt_type="compound").values())[0]
```
mgfritch: Why the change in whitespace?

srv/modules/runners/changed.py
```diff
@@ -129,8 +160,8 @@ def __init__(self, **kwargs):
         self.base_dir = '/srv/salt/ceph/configuration/files/'
         self.checksum_dir = self.base_dir + 'ceph.conf.checksum/'
         self.checksum_file = self.checksum_dir + self.role.conf_filename + self.role.conf_extension
-        log.debug("dependencies of role {}: {}".format(self.role.name,
-                                                       self.role.dependencies_unwrapped()))
+        log.debug("dependencies of role {}: {}".format(
+            self.role.name, self.role.dependencies_unwrapped()))
```
mgfritch: Change in whitespace to stay under char limit per line?

srv/modules/runners/changed.py
jschmid1: I get the point, but do we also stage 'rgws' for restarts then?

swiftgist: I think so, which then gets us into `rgw_configurations`.  sigh... Considering that's not a one-liner, should that just be a separate PR.  I don't know the urgency of this patch currently.

jan--f: Yeah I think the issue is that _any_ config option can potentially be in the global section. Techincally we'd have to restart the whole cluster on a change here.

This is certainly very heavy-handed. Not sure what the reasonable way is.

swiftgist: It is, but I swear that's where we were with this originally.  Since Ceph literally has 1K+ parameters (and maybe users shouldn't use them, but that doesn't stop them), we have no map to say `global X` impacts components A and B.  And even if we did, our maintained version would likely be wrong the minute we shipped.

Restarting the whole cluster gracefully is expensive, but anything less results in a bug like this one.

jschmid1:
note to self: split up commits in

* whitespace changes
* actual code changes

srv/modules/runners/changed.py
jschmid1: Changed it to also restart `rgw` and `mds` - `igw` is not maintained in that runner. However I'd argue this is the right interface to stage restarts.

srv/modules/runners/changed.py
jschmid1: I reverted the style changes

srv/modules/runners/changed.py
jschmid1: I reverted the style changes

jschmid1:
> note to self: split up commits in
> 
~> * whitespace changes~
> * actual code changes

For now, let's stick to the actual code changes until we reached consensus on the style-guide

jschmid1:
@susebot run teuthology

jschmid1:
> Commit [edbb5ad](https://github.com/SUSE/DeepSea/commit/edbb5ad2381332027aac7b9268bbd8c768fbd773) is _NOT OK_ for suite deepsea:tier2.
> Check tests results in the Jenkins job: http://ci.ses.suse.de:8080/job/pr-deepsea/282/

the CI caught a bug in this patch, fixing

srv/modules/runners/changed.py
```diff
@@ -74,6 +75,34 @@ def dependencies(self):
...
+        """ Set dependency to self if not global """
+        if self._role_name == 'global':
+            # if the global config changed
+            # stage restart on the core services
+            self.add_dependencies(Role(role_name='mon'))
+            self.add_dependencies(Role(role_name='mgr'))
+            self.add_dependencies(Role(role_name='storage'))
+            # also stage restarts on gateways (igw is not maintained here)
+            self.add_dependencies(Role(role_name='mds'))
+            self.add_dependencies(Role(role_name='rgw'))
```
jschmid1: This causes a loop due to this https://github.com/SUSE/DeepSea/blob/edbb5ad2381332027aac7b9268bbd8c768fbd773/srv/modules/runners/changed.py#L125-L145

I'm not sure why we set the dependency for `global` on a rgw_configuration change?. I'd argue for getting rid of this

jschmid1: by the looks of it, we don't need `rgw_configurations` at all. According to 

https://github.com/SUSE/DeepSea/blob/master/srv/salt/ceph/rgw/restart/controlled/default.sls#L1-L10

we restart all `rgw_configurations` if a `restart_rgw` grain is set. No need to check rgw_configrations inside the `changed` runner at all.

jschmid1:
ready for a second round of reviews

jschmid1:
@susebot run teuthology

jschmid1:
unrelated to *this* pr, but it seems we have a failure in stage.4.. re-running to verify

jschmid1:
@susebot run teuthology

jschmid1:
@smithfarm It seems that the CI is currently not able to deploy nfs-ganesha. Did we update the build artifacts already?

jschmid1:
@susebot run teuthology

smithfarm:
@jschmid1 I don't see where the CI is unable to deploy nfs-ganesha. What I see is this:

* Stage 4 completes successfully
* ganesha smoke test fails with:

```
2019-09-16T14:00:38.503 INFO:teuthology.orchestra.run.target192168000079.stderr:scripts/ganesha_smoke_test.sh: line 225: /root/mnt/rgw/bucket1/hello.txt: No such file or directory
```

jschmid1:
The failure is weird though. All tests before and after were successful.. This branch is clearly not changing anything in the NFS-ganesha area.. re-triggering to be extra-extra-sure

jschmid1:
@susebot run teuthology

jschmid1:
apparently we have two nfs-ganesha tests:

1) with mds backend
2) with rgw backed

only 1) fails. Maybe the mds state fails while checking for changed files somehow?
Why wouldn't the standard(non-ganesha) mds test fail?..

jschmid1:
I'm puzzled..

jschmid1:
Closing.. Reopen if needed

PR #1742, opened 2019-09-09
Add remove option to the ping to remove multiple ip interface

AvengerMoJo:
Signed-off-by: Alex Lau (AvengerMoJo) <alau@suse.com>

Fixes #
Exclude doesn't work the way it should when multiple ip apply to a single node. 

Description:
When net.ping with node get mulitple ip, "exclude" option will take out the minion completely with all the interfaces ip, only ip can be left to filter. So adding the "remove" option is to allow exclude actually take out specific subnet ip instead of the minion completely. 

-----------------

**Checklist:**
it is more a runner that help reporting the network environment that didn't mean to be called by external. Didn't design a unittest for this. 
- [*] Adapted documentation
command and code are doc accordingly
no offical bug reported yet, it is more like an enhancement for the ping filter 
Not related to the stage of deployment but pre-network test for better understanding the physical condition of the network. So virtual test shouldn't do much help here.

### DIFF
--- a/srv/modules/runners/net.py
+++ b/srv/modules/runners/net.py
@@ -45,6 +45,7 @@ def help_():
              'salt-run net.ping ceph:\n'
              'salt-run net.ping cluster=ceph:\n'
              'salt-run net.ping exclude=target:\n\n'
+             'salt-run net.ping remove=192.168.128.0/24:\n\n'
              '    Summarizes network connectivity between minion interfaces\n'
              '\n\n'
              'salt-run net.jumbo_ping:\n\n'
@@ -54,6 +55,7 @@ def help_():
              'salt-run net.iperf ceph:\n'
              'salt-run net.iperf cluster=ceph:\n'
              'salt-run net.iperf exclude=target:\n\n'
+             'salt-run net.ping remove=192.168.128.0/24:\n\n'
              '    Summarizes bandwidth throughput between minion interfaces\n'
              '\n\n')
     print(usage)
@@ -71,7 +73,7 @@ def get_cpu_count(server):
     return cpu_core
 
 
-def iperf(cluster=None, exclude=None, output=None, **kwargs):
+def iperf(cluster=None, exclude=None, remove=None, output=None, **kwargs):
     """
     iperf server created from the each minions and then clients are created
     base on the server's cpu count and request that number of other minions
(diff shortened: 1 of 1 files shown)


AvengerMoJo:
@jschmid1 sure here are the list of possible options after adding remove in the following 5 nodes example, each of the nodes have multiple interfaces and ip
subnet="192.168.2.0/24" are the control 1G link
subnet="192.168.128.0/24" are the public network 10G link
subnet="192.168.228.0/24" are the private network 10G link
as following
```
salt "*" grains.get ipv4
salt-master:
    - 10.0.2.32
    - 127.0.0.1
    - 192.168.2.200
    - 192.168.128.200
    - 192.168.228.200
mon-osd-3:
    - 127.0.0.1
    - 192.168.2.203
    - 192.168.128.203
    - 192.168.228.203
osd-4:
    - 127.0.0.1
    - 192.168.2.204
    - 192.168.128.204
    - 192.168.228.204
mon-osd-2:
    - 127.0.0.1
    - 192.168.2.202
    - 192.168.128.202
    - 192.168.228.202
mon-osd-1:
    - 127.0.0.1
    - 192.168.2.201
    - 192.168.128.201
    - 192.168.228.201
```

```
sudo salt-run net.ping exclude="S@192.168.128.0/24"
No minions matched the target. No command was sent, no jid was assigned.
No minions matched the target. No command was sent, no jid was assigned.
Succeeded: 0 addresses from 0 minions average rtt 0.00 ms
```
As you can see removing the 128 usbnet removing all result which is not easy to remove all 128 subnet interface ip by ip one by one. 
So with remove option, you can do the following 

```
Succeeded: 11 addresses from 1 minions average rtt 1.07 ms
Errored:
    10.0.2.32 from mon-osd-1
    10.0.2.32 from mon-osd-2
    10.0.2.32 from mon-osd-3
    10.0.2.32 from osd-4

```
As you can see 11 address only 1 get success because of failing with the 10.0.2.32 internal interface in one of the node which bring us to excluding that ip in the following
```
sudo salt-run net.ping exclude="10.0.2.32" remove="192.168.128.0/24"
Succeeded: 10 addresses from 5 minions average rtt 1.14 ms
```
Once excluding the 10.0.2.32 interface now you can see all 5 minion both private interface and control interface are inter-ping-able with each other. If user try to just test public or private network interface which are not inter-ping-able. Then we can easily use remove to do remove="192.168.2.0/24,192.168.128.0/24" for the private network and 
```
sudo salt-run net.ping exclude="10.0.2.32" remove="192.168.2.0/24,192.168.128.0/24"
Succeeded: 5 addresses from 5 minions average rtt 0.63 ms
```
remove="192.168.2.0/24,192.168.228.0/24" for the public network and 
```
sudo salt-run net.ping exclude="10.0.2.32" remove="192.168.2.0/24,192.168.228.0/24"
Succeeded: 5 addresses from 5 minions average rtt 0.66 ms
```

jschmid1:
@AvengerMoJo If you'd like to see this functionality in SES6, please create a forwardport. Otherwise this is good to be merged

PR #1750, opened 2019-09-16
cephfs: fix pool creation prerequisite (bsc#1149557)

jan--f:
With unless the state is executed when one condition returns false. For
example it'll try to create the cephfs_data pool if the pool is present
but the cephfs instance is not yet.
Negating the condition and using onlyif yields the expected behaviour.

### DIFF
--- a/srv/salt/ceph/mds/pools/default.sls
+++ b/srv/salt/ceph/mds/pools/default.sls
@@ -7,27 +7,29 @@ prevent empty rendering:
 
 {% else %}
 
+{# create cephfs_data pool iif it doesn't exist and no fs instance exists #}
 cephfs data:
   cmd.run:
     - name: "ceph osd pool create cephfs_data 256"
-    - unless:
-      - "rados lspools | grep -q cephfs_data"
-      - "ceph fs ls | grep -q ^name"
+    - onlyif:
+      - 'test -z "$(rados lspools | grep cephfs_data)"'
+      - 'test -z "$(ceph fs ls | grep ^name)"'
 
 cephfs data pool enable application:
   cmd.run:
-    - name: "ceph osd pool application enable cephfs cephfs_data || :"
+    - name: "ceph osd pool application enable cephfs_data cephfs || :"
 
+{# create cephfs_metadata pool iif it doesn't exist and no fs instance exists #}
 cephfs metadata:
   cmd.run:
     - name: "ceph osd pool create cephfs_metadata 64"
-    - unless:
-      - "rados lspools | grep -q cephfs_metadata"
-      - "ceph fs ls | grep -q ^name"
+    - onlyif:
+      - 'test -z "$(rados lspools | grep cephfs_metadata)"'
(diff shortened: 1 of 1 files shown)


swiftgist:
I think it's still a bit awkward to read with the number of negations.  Wouldn't 

```
{% if salt['cmd.run']('ceph fs ls') == "" %}
```
around all the steps be a bit clearer.  If ceph fs returns nothing, then create/enable both pools, etc.

That would also address the issue from the bug where `cephfs_data` had been manually created as an EC and the unless/ifonly conditional would have exactly one test.

srv/salt/ceph/mds/pools/default.sls
```diff
@@ -7,27 +7,29 @@ prevent empty rendering:
...
+{# create cephfs_data pool iif it doesn't exist and no fs instance exists #}
 cephfs data:
   cmd.run:
     - name: "ceph osd pool create cephfs_data 256"
-    - unless:
-      - "rados lspools | grep -q cephfs_data"
-      - "ceph fs ls | grep -q ^name"
+    - onlyif:
+      - 'test -z "$(rados lspools | grep cephfs_data)"'
+      - 'test -z "$(ceph fs ls | grep ^name)"'
```
Martin-Weiss: Is this "unless" an "or" or an "and"? 
It would be nice if the "ceph osd pool create cephfs_data" is NOT executed in case the pool exists but the ceph fs filesystem is not.

jan--f: We use `onlyif` here and that only executes the command if both tests return true. See https://docs.saltstack.com/en/latest/ref/states/requisites.html#onlyif

jan--f:
> I think it's still a bit awkward to read with the number of negations. Wouldn't
> 
> ```
> {% if salt['cmd.run']('ceph fs ls') == "" %}
> ```
> 
> around all the steps be a bit clearer. If ceph fs returns nothing, then create/enable both pools, etc.
> 
> That would also address the issue from the bug where `cephfs_data` had been manually created as an EC and the unless/ifonly conditional would have exactly one test.

Tbh I'm not a big fan of adding Jinja to states. The different execution contexts and times have often made things very awkward.
Do you think this is to confusing even with the comments?

swiftgist:
> Tbh I'm not a big fan of adding Jinja to states. The different execution contexts and times have often made things very awkward.
> Do you think this is too confusing even with the comments?

I understand the reluctance to add more Jinja. :)  Considering Martin's comment about the union/intersection of the two commands and a bit of the mental gymnastics to understand when the command is called, my thinking was that reducing each test to a single line would be more obvious.

Is it too confusing even with the comments?  I don't know honestly.  I know what it does.  Considering this achieves the goal and fixes the original issue, I'd say go ahead and merge.

jschmid1:
@jan--f Please create a backport so that we can merge those two together

jschmid1:
> @jan--f Please create a backport so that we can merge those two together

ignore this. I should've checked the next tab :>

PR #1753, opened 2019-09-17
cephfs: use ceph fs fail command to bring down MDS daemons before remove (bsc#1149530)

jan--f:
This new command fails all MDS daemons attached to a file system. Also
rename remove/mds to remove/cephfs and add it to stage 5.

Seems like no one ever tried to remove a cephfs instance...


jschmid1:
@susebot run teuthology

jschmid1:
@jan--f Would you be so kind to create a backport for this?

jan--f:
> @jan--f Would you be so kind to create a backport for this?

nah ;) Hardly any of this backports. I'll provide a PR against SES5 though.

PR #1755, opened 2019-09-18
stage 5: actually call remove.mds in stage 5

jan--f:

Can be considered a backport of #1753


jschmid1:
@susebot run teuthology

PR #1762, opened 2019-09-19
[next] introduce new return structure and refactor modules

jschmid1:
This is a dummy PR to publicly track our effort towards $subject.


jschmid1:
I fixed-up and split my commits properly. This should be ready for review now

PR #1768, opened 2019-09-23
ganesha: update ceph.conf in stage 3 instead of 4

rjfd:
Fixes: bsc#1150643


jschmid1:
@susebot run teuthology

PR #1777, opened 2019-10-08
runners/upgrade: enable automated upgrade if cluster is on nautilus

tserong:
This re-allows ceph.maintenance.upgrade, provided all nodes are running
Nautilus, so that function can be used to apply minor updates again.

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/modules/runners/upgrade.py
+++ b/srv/modules/runners/upgrade.py
@@ -77,12 +77,47 @@ class UpgradeValidation(object):
         """
         Check if the automated upgrade is supported
         """
-        msg = """
-                ************** PLEASE READ ***************
-                The automated upgrade is currently not supported.
-                Please refer to the official documentation.
-                ******************************************"""
-        return False, msg
+        __opts__ = salt.config.client_config('/etc/salt/master')
+        __grains__ = salt.loader.grains(__opts__)
+        __opts__['grains'] = __grains__
+        __utils__ = salt.loader.utils(__opts__)
+        __salt__ = salt.loader.minion_mods(__opts__, utils=__utils__)
+
+        os_codename, _, ceph_version = __utils__['status.get_sys_versions']()
+
+        os_codenames = set(os_codename.values())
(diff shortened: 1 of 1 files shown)


tserong:
I've labelled this DNM, as I haven't actually tested ceph.maintenance.upgrade yet on Nautilus, and I suspect nobody else has either. I mostly just want to get feedback on two things:

1) Is this check sensible?

2) What's up with https://github.com/SUSE/DeepSea/blob/079ad63d6593a9db6cf55a60987a2472aa18351d/srv/modules/runners/upgrade.py#L125-L126 ? It looks like the colocated_services check was never enabled, but the comment at https://github.com/SUSE/DeepSea/blob/079ad63d6593a9db6cf55a60987a2472aa18351d/srv/modules/runners/upgrade.py#L17-L26 seems to suggest it should be...

tserong:
Good point. Also I just realised that multiple Ceph versions are fine, provided they're within the same major release (imagine someone somehow gets half way through applying minor updates, then something breaks, then they need to run the update again). I'll rework this a bit.

tserong:
While poking around further I discovered that f2e85228 changed the logic in /srv/salt/ceph/maintenance/upgrade/master/default.sls. Previously, the upgrade would only run if minions.ready, upgrade.check and validate.setup all returned True. Now, since that commit, if minions.ready returns False, the upgrade proceeds anyway, and does *not* call upgrade.check or validate.setpu. @jschmid1 (or anyone else), can you recall the reason for this change?

tserong:
OK, that's a better check and error messages.

Still need to figure out what's up with f2e8522. I don't think the behaviour introduced there is correct. Possibly the intention was to wait a while if some nodes are down in case they come back, but then to proceed with the upgrade anyway? If that's the case, surely it should still run upgrade.check and validate.setup?

jschmid1:
Closing.. Reopen if needed

PR #1787, opened 2019-10-23
[next] add help() func to test runner

jschmid1:
* rely on a doc_template living in doc.py
  This will make sure to have a consistent way of writing help texts
  for all our runners.

* use pydoc's pager() functionality to use the systems default pager


jschmid1:
This is an easier approach as opposed to writing .rst documentation that will be exported to .pdf, .man, .html and other useful formats.

PR #1790, opened 2019-10-31
Remove python dependency on host when building local rpm

swiftgist:

Description:
The pyc files require build host to have a compatible python3 to build an rpm.  

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/deepsea.spec.in
+++ b/deepsea.spec.in
@@ -65,7 +65,7 @@ A collection of Salt files providing a deployment of Ceph as a series of stages.
 %setup
 
 %build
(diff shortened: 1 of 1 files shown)


swiftgist:
@susebot run teuthology

swiftgist:
> /lgtm
> 
> Maybe we should also remove the `pyc` target from the Makefile too ?

Possibly... somebody may want still want this for the `make install` or maybe in QA.  It's not a development environment requirement.  If nobody pipes up, I'm okay with removing it altogether in another PR.

PR #1801, opened 2019-11-21
Workaround for python-rados package version

swiftgist:
Zypper cannot successfully resolve both ceph and python-rados in the
same path, but will pick the previous version of librados when
python-rados is separated.  This allows Stage 0 to complete. Manual
intervention may remain on the Salt master due to the deepsea rpm.

-----------------

**Checklist:**
- [ ] Adapted documentation


jschmid1:
@susebot run teuthology

swiftgist:
Looks like this is not needed for the CI.

PR #1804, opened 2019-11-26
ADD doc/examples/policy.cfg-full: full example using all roles

johanneskastl:
Description:
Add a full example with all roles that currently exist.

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/doc/examples/policy.cfg-full
+++ b/doc/examples/policy.cfg-full
@@ -1,0 +1,61 @@
+## Cluster Assignment
+cluster-ceph/cluster/*.sls
+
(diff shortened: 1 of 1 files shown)


johanneskastl:
@susebot run teuthology

johanneskastl:
I would have liked to find something like this when starting with the deployment. Not sure if it is really complete, or if there are some roles missing?

Those are the roles I had when deploying with deepsea-0.9.23+git.0.6a24f24a0-3.3.2.noarch.

PR #1807, opened 2019-11-28
osd/replace: only access metadata if present

jschmid1:
When a OSD is not actively deployed anymore
but is still present in the crushmap you want to be
able to replace it.
This patch extends the condition for such a case.

suse_internal: bsc#1157469


jschmid1:
@susebot run teuthology

PR #1821, opened 2020-02-14
Fix iSCSI default values

ricardoasmarques:
This PR fixes some iSCSI default values to match the ones configured in `ceph-iscsi`:

https://github.com/SUSE/ceph-iscsi/blob/ses6/ceph_iscsi_config/gateway_setting.py#L212

Signed-off-by: Ricardo Marques <rimarques@suse.com>

PR #1824, opened 2020-02-18
[ses6][backport] Fix iSCSI default values

ricardoasmarques:
Backport of https://github.com/SUSE/DeepSea/pull/1821


jschmid1:
I cannot replicate the linter issues locally, maybe something has changed in the jenkins worker @gekios ?

gekios:
Hello,
Last week @toabctl  upgraded the server so I guess that it has newer version of pylint that's catching more things.. I compared the logs and see that from Friday it starts complaining about `3rd-party-module-not-gated` which if I am not mistaken is `W8410` that we can disable.

gekios:
Opened https://github.com/SUSE/DeepSea/pull/1825 to disable this check.

PR #1825, opened 2020-02-18
pylintrc: Disable W8410 check

gekios:
I propose disabling W8410 - 3rd-party-module-not-gated, in order to fix the CI
failures that got introduced by server upgrade to Leap 15.1.

Signed-off-by: Georgios Kyratsas <gkyratsas@suse.com>

Description:

-----------------

**Checklist:**
- [ ] Adapted documentation


jan--f:
huh any idea why pylint suddenly thinks `os` is a 3rd party module?

gekios:
Yeah actually I noticed this as well and after googling a bit it seems to be missing from all the lists that make`known_std_modules` https://github.com/saltstack/salt-pylint/blob/37d7a63ea808fd42c68378ddf6a5429570c800ac/saltpylint/thirdparty.py#L68

gekios:
Hmmm, I added every module that it was complaining about in `allowed-3rd-party-modules` and it passed but is this the right approach?

jschmid1:
probably not, but it lets us move ahead :>

PR #1840, opened 2020-06-10
runners/upgrade: Add ceph_salt_config function

tserong:
This adds minimal JSON config generation suitable for import into ceph-salt when switching from DeepSea to cephadm, so the admin doesn't have to manually re-add all the minions.  To use:

- Run `salt-run upgrade.ceph_salt_config --out=json > config.json`
- Uninstall DeepSea
- Install ceph-salt
- Run `ceph-salt import config.json`

The admin still needs to use `ceph-salt config` to generate ssh keys, set the container image, and set time server subnet and external servers.

### DIFF
--- a/srv/modules/runners/upgrade.py
+++ b/srv/modules/runners/upgrade.py
@@ -239,6 +239,63 @@ def status():
     return ""
 
 
+def ceph_salt_config():
+    """
+    Generate minimal JSON suitable for import into ceph-salt, based on DeepSea
+    configration.  Call via `salt-run upgrade.ceph_salt_config --out=json`
+    """
+    config = {
+        'minions': {
+            'all': [],
+            'admin': [],
+            'cephadm': []
+        },
+        'bootstrap_minion': None,
+        'bootstrap_mon_ip': None,
+        'dashboard': {
+            # These values are just so the display gives an indication of what's
+            # really going to happen when the user runs `ceph-salt config ls`
+            # (the dashboard won't be reconfigured during an upgrade)
+            'username': 'USE EXISTING',
+            'password': 'USE EXISTING',
+            'password_update_required': False
+        },
+        'time_server': {
+            'enabled': True,
+            'external_time_servers': ['pool.ntp.org'],
+            'server_host': '',
+            'subnet': None
+        }
+    }
+
+    __opts__ = salt.config.client_config('/etc/salt/master')
+    __grains__ = salt.loader.grains(__opts__)
+    __opts__['grains'] = __grains__
+    __utils__ = salt.loader.utils(__opts__)
+    __salt__ = salt.loader.minion_mods(__opts__, utils=__utils__)
+    master_minion = __salt__['master.minion']()
+
+    runner = salt.runner.RunnerClient(__opts__)
+    mon_ips = dict(runner.cmd('select.public_addresses',
+                              ['cluster=ceph', 'roles=mon', 'tuples=True'], print_event=False))
+
+    local = salt.client.LocalClient()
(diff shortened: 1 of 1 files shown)


tserong:
If someone can come up with a better name for this function, I welcome it, hence the DNM label :-)

srv/modules/runners/upgrade.py
```diff
@@ -239,6 +239,63 @@ def status():
...
+            # really going to happen when the user runs `ceph-salt config ls`
+            # (the dashboard won't be reconfigured during an upgrade)
+            'username': 'USE EXISTING',
+            'password': 'USE EXISTING',
+            'password_update_required': False
+        },
+        'time_server': {
+            'enabled': True,
+            'external_time_servers': ['pool.ntp.org'],
+            'server_host': '',
```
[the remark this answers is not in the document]
tserong: Yeah, that gets set on line 294 (is there an easier way to get pillar data in a runner than what I'm doing there?)

tserong:
> When or how would this get called?

The admin would need to run it per the instructions in the commit message, before removing DeepSea and installing ceph-salt.

Martin-Weiss:
In my setup I have "time_init: disabled"  in /srv/pillar/ceph/stack/global.yml however the json has this:

    "time_server": {
        "enabled": true,
        "external_time_servers": [
            "pool.ntp.org"
        ],
        "server_host": "",
        "subnet": null
    }

I assume this this wrong? I would have expected that if I have time_init disabled it will create a json for ceph-salt that has the same config?

srv/modules/runners/upgrade.py
```diff
@@ -239,6 +239,63 @@ def status():
...
+
+    local = salt.client.LocalClient()
+    roles = local.cmd("I@cluster:ceph", 'pillar.get', ['roles'], tgt_type="compound")
+
+    for node in sorted(roles):
+        config['minions']['all'].append(node)
+        config['minions']['cephadm'].append(node)
+        if 'master' in roles[node] or 'admin' in roles[node]:
+            config['minions']['admin'].append(node)
+        if not config['bootstrap_minion'] and 'mon' in roles[node]:
```
ricardoasmarques, line 290: In https://github.com/ceph/ceph-salt/pull/315 we are making Bootstrap Minion optional, so we will not need to set it in an upgrade scenario.

srv/modules/runners/upgrade.py
```diff
@@ -239,6 +239,63 @@ def status():
...
+    mon_ips = dict(runner.cmd('select.public_addresses',
+                              ['cluster=ceph', 'roles=mon', 'tuples=True'], print_event=False))
+
+    local = salt.client.LocalClient()
+    roles = local.cmd("I@cluster:ceph", 'pillar.get', ['roles'], tgt_type="compound")
+
+    for node in sorted(roles):
+        config['minions']['all'].append(node)
+        config['minions']['cephadm'].append(node)
+        if 'master' in roles[node] or 'admin' in roles[node]:
```
ricardoasmarques, line 288: Since https://github.com/ceph/ceph-salt/pull/361 and https://github.com/ceph/ceph-salt/pull/364, `ceph-salt` introduced `tuned/latency` and `tuned/throughput` roles, you may want to include those roles in the generated config.

tserong:
Closing in favour of #1852 - sorry to break the flow of comments like that, but the new PR adds a whole bunch of other stuff too.

I didn't add `tuned/latency` and `tuned/throughput` roles, as I couldn't see a way of deciding what minions to apply those to, based on DS config.

PR #1846, opened 2020-07-28
aa-teardown can fail if apparmor is disabled on the boot command line

swiftgist:
bnc: 1174536

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/salt/ceph/apparmor/default-teardown.sls
+++ b/srv/salt/ceph/apparmor/default-teardown.sls
@@ -7,6 +7,8 @@ aa-teardown:
   cmd.run:
     - onlyif:
       - which aa-teardown
+    - onfail:
+      - test: apparmor
 
 stop apparmor:
   service.dead:


swiftgist:
The onfail is like a goto... so adding the failure condition for the binary being present but disabled in the kernel command line.  This causes the state to succeed in that case, but leaves the original behavior intact.

jan--f:
huh...is there a naming conflict or version issue? Cause https://docs.saltstack.com/en/latest/ref/states/requisites.html#onfail reads differently to me. Or at least to goto-like aspect isn't mentioned.

swiftgist:
Take a look at the whole file, it's the same behavior with the previous step.  Also, I did have Martin test in his environment and the state succeeded.  Would you have another solution?

jan--f:
Well who am I to argue with successful testing :)
Only alternative might be adding jinja and I'm always up for avoiding that.

jschmid1:
@susebot run teuthology

PR #1848, opened 2020-08-18
validate: check if domain grain is set

jschmid1:

If a user misconfigured their fqdn like so:

```
# hostname --fqdn
target-00
```

[salt can't determine the `domain`](https://github.com/saltstack/salt/blob/7c0b30d9a423a4dca21c9dcaa5a35849417f5597/salt/grains/core.py#L1633-L1654
)

```python
In [2]: 'target-00'.partition('.')[::2]                                         
Out[2]: ('target-00', '')

```

[And we fail here ](https://github.com/SUSE/DeepSea/blob/04335830b8ba1630071a19fbfe400d089a139fa2/srv/salt/_modules/deepsea.py#L94-L99)

with:

`Exception: According to the 'domain' grain, the cluster does not have a DNS domain configured`

This is probably too late. Also the error message is not very descriptive.

My concern is that we break existing (working) installations that don't use SSL (yet).

Maybe raising a warning instead of an error is also fine?

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/modules/runners/validate.py
+++ b/srv/modules/runners/validate.py
@@ -706,7 +706,25 @@ class Validate(Preparation):
 
         self._set_pass_status('fqdn')
 
-    # pylint: disable=line-too-long
+    def domain(self):
+        """
+        Verify that the domain grain exists
+        """
+        for minion_id in self.grains.keys():
+            domain = self.grains[minion_id]['domain']
+            fqdn = self.grains[minion_id]['fqdn']
+            msg = f"""
+In order to determine a `domain` which is required for some features in DeepSea you need to set-up the FQDN properly.
+Example:
+A device with the hostname `myhost` in the parent domain example.com has the fully qualified domain name myhost.example.com. The FQDN uniquely distinguishes the device from any other hosts called myhost in other domains.
+
+Please make sure 'hostname --fqdn' returns an FQDN, for example, myhost.example.com
+Your currently configured FQDN is: <$fqdn>
+ """
+            if not domain:
+                self.errors.setdefault('domain', []).append(msg)
+        self._set_pass_status('domain')
+
     def openattic(self):
         """
         Check for incompatible issues for openATTIC
@@ -1416,6 +1434,7 @@ def pillar(cluster=None, printer=None, **kwargs):
     valid.pool_creation()
     valid.time_server()
     valid.fqdn()
+    valid.domain()
     valid.report()
 
     if valid.errors:


kshtsk:
Related ticket: https://jira.suse.com/browse/SES-2122

srv/modules/runners/validate.py
```diff
@@ -706,7 +706,18 @@ def fqdn(self):
...
 
-    # pylint: disable=line-too-long
+    def domain(self):
+        """
+        Verify that the domain grain exists
+        """
+        for minion_id in self.grains.keys():
+            domain = self.grains[minion_id]['domain']
+            fqdn = self.grains[minion_id]['fqdn']
+            msg = f"The domain grain is not set. The reason might be a missconfigured FQDN. Yours: <{fqdn}>"
```
swiftgist, line 716: Just one 's' in misconfigured.  Did you want "Yours" to actually be minion_id.  If the fqdn is blank, they would still need to track it down.

kshtsk: Could you please add here a message to the user, how user could verify if they have FQDN configured correctly from the Salt point of view.

jschmid1: What about:

```
In order to determine a `domain` which is required for some features in DeepSea you need to set-up the FQDN properly.
Example:
A device with the hostname `myhost` in the parent domain example.com has the fully qualified domain name myhost.example.com. The FQDN uniquely distinguishes the device from any other hosts called myhost in other domains.
```

kshtsk: @jschmid1 it does not say what exactly user should verify to make DeepSea happy, maybe include text like:
```Please make sure 'hostname --fqdn' returns FQDN, for example, host.example.com```?

Comment (rbeldin) [reply srv/modules/runners/validate.py:716]: A message to the user would be quite nice.  I ran into this issue and had to dig around in deepsea to figure it out.

Comment (jschmid1) [reply srv/modules/runners/validate.py:716]:
```
In order to determine a `domain` which is required for some features in DeepSea you need to set-up the FQDN properly.
Example:
A device with the hostname `myhost` in the parent domain example.com has the fully qualified domain name myhost.example.com. The FQDN uniquely distinguishes the device from any other hosts called myhost in other domains.
Please make sure 'hostname --fqdn' returns FQDN, for example, host.example.com
Your currently configured FQDN is: <$fqdn>
```

Alright. Is this fine with everyone?

swiftgist: sounds good.

kshtsk: (sating hello sitting on vacation) Looks fine to me, just minor thing, since you're already using 'myhost' in the example and the example itself, probably it would be more appropriate:

> *Please make sure 'hostname --fqdn' returns **myhost.example.com** such as above*.

jschmid1: ack

jschmid1:
@susebot run teuthology

PR #1849, opened 2020-09-07
fix many-to-many not allowed in Grafana dashboard

p-se:
One part of the issue is that the metrics from the ceph mgr contain labels for `instance`, though, the `instance` label is also automatically populated by Prometheus. When Prometheus does that, the instance label becomes the hostname of the Ceph manager node, which is wrong. By setting `honor_labels` in the scrape configuration of `ceph-mgr` to `true`, the instances labels exported by the Prometheus manager module are preserved.

Partly fixes bsc#1175585

-----------------

**Checklist:**
- [ ] Adapted documentation


jschmid1:
@susebot run teuthology

PR #1856, opened 2020-10-14
Add multi MDS per node feature

jan--f:
Description:
This pr enables deploying multiple MDS daemons per node. This is configured by adding `mds_daemons_per_node: $n` to the pillar. This can be done either globally (via ` /srv/pillar/ceph/stack/global.yml`) or only on specific MDS nodes (via one of the files in `/srv/pillar/ceph/stack/ceph/minions/`). Rescind, restart and the shirnk to update functionality can handle multiple daemons per node.

-----------------

### DIFF
--- a/srv/salt/_modules/mds.py
+++ b/srv/salt/_modules/mds.py
@@ -11,12 +11,15 @@ import logging
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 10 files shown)


jschmid1:
@susebot run teuthology

PR #1857, opened 2020-10-15
Pass --container-init to `cephadm adopt` (bsc#1177588)

tserong:
Note that while this causes the adopted containers to be run with --init, the user must still run `ceph config set mgr mgr/cephadm/container_init true` at some point, so that any new containers created in future are also run with --init.

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1177588


jschmid1:
@susebot run teuthology

smithfarm (approved):

PR #1862, opened 2020-11-27
monitoring: fix node_exporter overrides in prometheus.yml (bsc#1179029)

tserong:
f24acb4b changed the name of the node exporter pillar variables from "node" to "node_exporter", but missed making a matching change to the prometheus.yml config file.

This commit also fixes application of custom prometheus rule_files (there was a missing space when invoking `basename rule_file`).

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1179029


jan--f (approved):
:+1:

p-se (approved):

PR #1863, opened 2020-12-03
Fix setting custom Grafana certificates and Prometheus CLI flags

p-se:

Fixes issues in using pillar values for setting custom Grafana certificates and providing additional command line flags for Prometheus (to increase retention time, for instance).

The template inside these conditions has probably never worked, as it cannot be compiled. The conditional code is only compiled if the pillar values are specified, so it didn't reveal an issue if those features weren't used, which seems to have always been the case.

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/salt/ceph/monitoring/grafana/create_configs.sls
+++ b/srv/salt/ceph/monitoring/grafana/create_configs.sls
@@ -10,7 +10,7 @@
 
 /srv/salt/ceph/monitoring/grafana/cache/tls/certs/grafana.crt:
   file.managed:
-    - source: {{ pillar['monitoring:grafana:ssl_cert'] }}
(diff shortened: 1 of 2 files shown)


smithfarm:
@susebot run teuthology

smithfarm:
@tserong It is my understanding that this PR fixes https://bugzilla.suse.com/show_bug.cgi?id=1179581

Therefore, I have added it to https://confluence.suse.com/display/SUSEEnterpriseStorage/SES6+DeepSea+Maintenance+Updates

It passed `susebot run teuthology` - shall we go ahead and merge it now so it makes it into 0.9.34? (It is already circulating as a PTF.)

tserong (approved):
Fine by me

PR #1865, opened 2020-12-17
osd: add method to zap simple osds

jan--f:
This parses the c-v simple scan result in /etc/ceph/osd to retrieve the
devices the osd.$i uses and then passes them to c-v lvm zap. The zap
subcommand will zap devices no problem, the trick is really just to
figure out which devices to zap.

Fixes bsc#1178657

Description:

-----------------

**Checklist:**
- [ ] Adapted documentation


tserong (approved):
LGTM, and tested OK according to https://bugzilla.suse.com/show_bug.cgi?id=1178657#c20

jan--f:
:+1: I also confirmed that existing running OSDs won't be accidentally zapped by the new script (in case the files in `/etc/ceph/osd` are outdated or wrong) as an unmount failes with `Device busy`.

PR #1868, opened 2021-01-19
monitoring: enable custom scrape configurations

p-se:
Scrape configurations cannot be loaded externally by using a statement
like `include` inside the Prometheus configuration file. The
configuration management system is supposed to create the Prometheus
configuration file appropriately. By using service discovery the user
wouldn't be able to configure some details like the `scrape_interval` or
those would need to be configured separately.

This implementation enables to provide a whole block to
`/srv/pillar/ceph/stack/global.yml` under the
`monitoring:prometheus:custom_scrape_configs` hierarchy and does so by
iterating the keys and values for least implementation effort.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1179029

### DIFF
--- a/srv/salt/ceph/monitoring/prometheus/files/prometheus.yml.j2
+++ b/srv/salt/ceph/monitoring/prometheus/files/prometheus.yml.j2
@@ -13,6 +13,11 @@ rule_files:
 
 scrape_configs:
   # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
+  {% for config in salt['pillar.get']('monitoring:prometheus:custom_scrape_configs', {}) %}
+  {%- for key, value in config.items() %}
+  {%- if loop.first %}- {% else %}  {% endif %}{{- key }}: {{ value | json}}
+  {% endfor %}
+  {%- endfor -%}
   - job_name: 'ceph-mgr'
     honor_labels: true
     scrape_interval: {{ salt['pillar.get']('monitoring:prometheus:scrape_interval:ceph', '10s')|yaml }}


smithfarm:
@p-se Could you include (ideally, in the commit message itself) a brief explanation why this can't be done upstream?

smithfarm (approved):
In my experience, Jinja's behavior when it comes to newlines (whether they get added or not) is odd and difficult to control.

If this patch has been tested and shown to work, then I don't have any problem with it.

p-se:
> In my experience, Jinja's behavior when it comes to newlines (whether they get added or not) is odd and difficult to control.
> 
> If this patch has been tested and shown to work, then I don't have any problem with it.

Yes, it has been tested and proven to work.

The resulting YAML snippet does not look as readable as YAML is used to, as it is exported in JSON format (which is luckily 100% compatible with YAML), but it enables to simply write all kinds of values (even those with nested structures) into the file without any issues.

It would have to be evaluated, but maybe this implementation could even be used as basis for clean-ups which could potentially  reduce the complexity and size of many of the SLS files in the project.

srv/salt/ceph/monitoring/prometheus/files/prometheus.yml.j2
```diff
@@ -13,6 +13,11 @@ rule_files:
 
 scrape_configs:
   # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
+  {% for config in salt['pillar.get']('monitoring:prometheus:custom_scrape_configs', {}) %}
+  {%- for key, value in config.items() %}
+  {%- if loop.first %}- {% else %}  {% endif %}{{- key }}: {{ value | json}}
+  {% endfor %}
+  {%- endfor -%}
```
jan--f, line 20: What is the function of the `{%-` and `-%}` delimeters?

tserong: It's a Jinja thing which strips whitespace from the output

jan--f (approved):

PR #1871, opened 2021-02-02
upgrade to cephadm: fix Drive Group generation

sebastian-philipp:
Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1181665
Signed-off-by: Sebastian Wagner <sebastian.wagner@suse.com>

Description:

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/modules/runners/upgrade.py
+++ b/srv/modules/runners/upgrade.py
@@ -7,12 +7,14 @@ Verify that an automated upgrade is possible
 from __future__ import absolute_import
 from __future__ import print_function
 import re
+from typing import Iterable, List
 # pylint: disable=import-error,3rd-party-module-not-gated,redefined-builtin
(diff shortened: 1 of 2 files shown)


tserong (approved):
Thanks! (FTR, I figured out the problem with the original implementation was it incorrectly assumed DeepSea drive groups were multi doc yaml -- see https://gist.github.com/jschmid1/7e3b6a8b11a17e893f7bb0fd030ea5f3)

sebastian-philipp:
I think we have to cope wit both: multi-document and single-document with many drive groups

tserong:
> I think we have to cope wit both: multi-document and single-document with many drive groups

Makes sense -- this PR seems to handle both (or even a mix), so I think we're good now.

PR #1872, opened 2021-02-10
Rework config change detection to handle global.conf correctly (bsc#1181183)

tserong:
This rather large change makes DeepSea schedule all daemons for restart if global.conf has changed.  The original (unimplemented) idea is that there was going to be a tree of roles, so each individual role could depend on global.conf, so if global.conf changed, it would propagate through this tree.  Unfortunately, that was never going to work due to the way the checksum verification/generation is done.  The original implementation called changed.mon, then changed.osd, then changed.mgr in that order.  If each of those in turn were to check if the global config had changed, then that'd be true on the first call, so the mons would get their restart flag set, but would no longer be true for subsequent calls (the mgr and osd wouldn't be restarted).  So I got rid of that entire tree concept altogether, and collapsed everything back to one function which checks for all possible changes and sets the restart grains appropriately.
    
Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1181183

### DIFF
--- a/man/deepsea-commands.7
+++ b/man/deepsea-commands.7
@@ -1087,29 +1087,6 @@ salt-run validate.deploy
 .RE
 salt-run validate.saltapi
 .PP
-salt-run changed.rgw
-.RS
-.RE
-salt-run changed.mds
-.RS
-.RE
-salt-run changed.osd
-.RS
-.RE
-salt-run changed.mon
-.RS
-.RE
-salt-run changed.igw
-.RS
-.RE
-salt-run changed.global
-.RS
-.RE
-salt-run changed.client
-.RS
-.RE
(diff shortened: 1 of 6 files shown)


tserong:
@jan--f do you mind eyeballing this? It'll make a lot more sense if you a side-by-side/split diff, rather than a unified diff, given the magnitude of the change.

jan--f (approved):
Code change looks good.

Should we add a documentation bug that points this behavior out specifically? I think adding a note about this and the recommendation to use runtime config changes for temporary/testing changes to avoid unnecessary restarts of all daemons, as this can take quite a while.

tserong:
Yeah, good idea, I'll sort something out for docs

PR #1873, opened 2021-03-18
Use -i to pass credentials to "ceph dashboard" commands (bsc#1183600)

tserong:
Ceph Nautilus 14.2.17+ requires that user, rgw and iscsi credentials passed to `ceph dashboard` commands come via a file, rather than being specified directly as literal parameters (see https://github.com/ceph/ceph/pull/38832)

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1183600

### DIFF
--- a/srv/salt/ceph/dashboard/default.sls
+++ b/srv/salt/ceph/dashboard/default.sls
@@ -70,7 +70,7 @@ dashboard user exists:
 set username and password:
   cmd.run:
     # This command is printed although the 'onchange' statement evaluates as true. This might cause confusion.
-    - name: ceph dashboard ac-user-create {{ dashboard_user }} {{ dashboard_pw }} administrator
+    - name: "echo -n {{ dashboard_pw }} | ceph dashboard ac-user-create {{ dashboard_user }} -i - administrator"
     - onchanges:
         - cmd: dashboard user exists
 
--- a/srv/salt/ceph/rgw/dashboard.sls
+++ b/srv/salt/ceph/rgw/dashboard.sls
@@ -1,11 +1,11 @@
 configure dashboard rgw access key:
   cmd.run:
-    - name: "ceph dashboard set-rgw-api-access-key $(radosgw-admin user info --uid=admin | jq -r .keys[0].access_key)"
+    - name: "echo -n $(radosgw-admin user info --uid=admin | jq -r .keys[0].access_key) | ceph dashboard set-rgw-api-access-key -i -"
     - fire_event: True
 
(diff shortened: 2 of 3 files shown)


tserong:
The one possible kink here is that as written this commit means DeepSea won't do the right thing with Ceph < 14.2.17, which means you need to be updating Ceph and DeepSea all at the same time. I could probably make it check the Ceph version first, but I'm not sure it's worth the added complexity.

smithfarm:
Testing with `sesdev create nautilus --single-node --deepsea-branch wip-bsc1183600 --non-interactive`

srv/salt/ceph/dashboard/default.sls
```diff
@@ -70,7 +70,7 @@ dashboard user exists:
 set username and password:
   cmd.run:
     # This command is printed although the 'onchange' statement evaluates as true. This might cause confusion.
-    - name: ceph dashboard ac-user-create {{ dashboard_user }} {{ dashboard_pw }} administrator
+    - name: "echo {{ dashboard_pw }} | ceph dashboard ac-user-create {{ dashboard_user }} -i - administrator"
```
sebastian-philipp, line 73: we're sure the password doesn't end with a newline? relates to https://github.com/ceph/ceph/pull/39438 @smithfarm this was part of the latest nautilus release, right?

smithfarm: @sebastian-philipp That backport PR points to the backport tracker issue which has that information.

tserong: In that context, the password DeepSea generates is a random string of crap which won't include newlines.

sebastian-philipp: (it's just `echo` without `-n` adds a newline)

smithfarm: > (it's just `echo` without `-n` adds a newline)

Good point!

```
$ echo "foo" | hexdump -c
0000000   f   o   o  \n                                                
0000004
```

smithfarm:
> Testing with `sesdev create nautilus --single-node --deepsea-branch wip-bsc1183600`

Trying that again with https://github.com/SUSE/sesdev/pull/584 applied...

smithfarm (approved):
Probably will be good to add `-n` to all the `echo` statements as per @sebastian-philipp 's suggestion.

Once that is done, :+1: to getting a build into filesystems:ceph:nautilus and Devel:Storage:6.0 for further testing.

PR #1876, opened 2021-05-24
mds/restart: only check MDS processes (not all processes)

tserong:
When restarting the MDS during stage 4, we were invoking ceph.processes to check if all processes were running.  The problem with doing this is that if you have custom global ceph configuration set up prior to initial deployment *and* are also deploying other stage 4 services which are normally deployed *after* MDS (i.e. RGW or Ganesha), the check for custom configuration changes flags all services for restart, so when ceph.processes is invoked it waits for all daemons to be running, even those which aren't yet deployed.  That was never going to work, so instead we're now only checking for MDS processes when restarting MDSes, which makes more sense than checking all processes anyway.

-----------------

**Checklist:**
- [ ] Adapted documentation

### DIFF
--- a/srv/salt/ceph/functests/1node/restart/common.sls
+++ b/srv/salt/ceph/functests/1node/restart/common.sls
@@ -52,7 +52,7 @@ distribute ceph.conf {{ service }}:
     - tgt_type: compound
     - sls: ceph.configuration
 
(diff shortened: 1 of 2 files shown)


tserong:
@susebot run teuthology

PR #1880, opened 2021-07-21
osd: handle `ceph osd ok-to-stop` output from ceph v14.2.22

tserong:
Prior to https://github.com/ceph/ceph/commit/a45bfebb, the `ceph osd ok-to-stop` command output a long, human readable
string.  DeepSea was looking at this string to see if OSDs really were OK to stop.  Since that commit, the output of this command is now a chunk of JSON, which obviously can't work with that string comparison, so let's try to parse the JSON first, in case we're using the latest Nautilus release, and if that fails, fall back to the previous check in case we're still running against an older Ceph.

### DIFF
--- a/srv/modules/runners/osd.py
+++ b/srv/modules/runners/osd.py
@@ -459,13 +459,19 @@ def ok_to_stop_osds(osd_list):
         tgt_type="glob",
     )
     message = list(ret.values())[0]
-    if 'are ok to stop without reducing availability' in message:
-        return True
-    else:
-        print(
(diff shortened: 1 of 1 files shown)


tserong:
This made it through my manual testing against ceph 14.2.21 and 14.2.22 (`salt-run state.orch ceph.functests.1node.replace.delay` succeeded and `salt-run state.orch ceph.functests.1node.replace.multiple` failed, when run against both those versions of ceph, which is the outcome I expected).

ml8mr (approved):
LGTM

sallen-suse (approved):
Looks good to me.

PR #1882, opened 2021-11-12
Bump OSD available timeout to 120, make it configurable (bsc#1192520)

tserong:
When restarting OSDs, we wait for them to become available before proceeding to the next OSD.  Previosly, this wait was in a loop with a timeout of 60 seconds.  This commit increases the default timeout to 120 seconds, and also makes it configurable in case that's still too short in some circumstances.  If you need a different timeout, specify it in `/srv/pillar/ceph/stack/global.yml`, e.g.:

```
osd_available_timeout: 240
```

Note that the way the loop is implemented, the sleep in the loop doubles on each iteration.  The runtime of the loop will thus be pretty close to a power of 2, e.g.: 1, 2, 4, 8, 16, 32, 64, ... seconds.  It won't stop when the exact timeout is reached, rather
it will stop at approximately the next power of 2 seconds greater than the specified timeout.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1192520

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -978,17 +978,36 @@ def takeover():
 
 
 def wait_until_available(osd_id):
-    """Wait slightly over a minute for an OSD to become available (use e.g. after restarting)"""
-    timeout=0.5
-    while timeout < 60:
-        _, stdout, _ = __salt__['helper.run']("ceph daemon osd.{} status 2>/dev/null | jq -r .state".format(osd_id))
+    """Wait for an OSD to become available (use e.g. after restarting)"""
+    # This will return immediately if the OSD is already active.  If it's
+    # not active, it will sleep in a loop and retry the check periodically
+    # until a timeout is exceeded.  In order to return quickly for OSDs
+    # that start quickly, the first sleep is 1 second, then 2 seconds,
+    # then 4, 8, 16, 32, but after that it reverts to sleeping 60 seconds
+    # per iteration.  The default timeout is 600 seconds (10 minutes).
+    timeout = __pillar__.get('osd_available_timeout', 600)
+    delay = 0.5
+    start = time.monotonic()
+    while True:
+        duration = int(time.monotonic() - start)
+
+        _, stdout, _ = __salt__['helper.run'](
+            "ceph daemon osd.{} status 2>/dev/null | jq -r .state".format(osd_id)
+        )
         if stdout == "active":
-            log.info("OSD {} is now active (timeout: {})".format(osd_id, timeout))
+            log.info("OSD {} is now active (waited {} seconds)".format(osd_id, duration))
             return True
-        time.sleep(timeout)
-        timeout *= 2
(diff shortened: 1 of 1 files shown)


tserong:
Please ignore the lint error `E:  1, 0: Module file has the wrong file permissions(expected 0o644): 0o664 (file-perms)` -- it's been like that forever, along with all the other modules files, and I'm not changing that now.

trociny (commented):
So it might be a little confusing for a user if she does not know how this `osd_available_timeout` is used, when e.g. she sets it to 240 as in your example but it will actually be 127.5 sec only.

I don't have objection against the current implementation though it seems it could be improved. E.g. if you double `timeout` before sleeping, than for the actual timeout would be always bigger than what the user set (255 for the 240 example above), which seems less confusing.

srv/salt/_modules/osd.py
```diff
@@ -978,10 +978,13 @@ def takeover():
 
 
 def wait_until_available(osd_id):
-    """Wait slightly over a minute for an OSD to become available (use e.g. after restarting)"""
-    timeout=0.5
-    while timeout < 60:
-        _, stdout, _ = __salt__['helper.run']("ceph daemon osd.{} status 2>/dev/null | jq -r .state".format(osd_id))
+    """Wait for an OSD to become available (use e.g. after restarting)"""
+    timeout_limit = __pillar__.get('osd_available_timeout', 120)
```
trociny, line 982: and I would prefer the default something power of two so it is close to the "effective".

tserong:
Thanks, I'll try to rework that slightly so it's less confusing. With the current implementation, if you set 240, it _will_ actually run for slightly over 255 seconds, not 127.5, but I had to instrument it with a couple of calls to time.monotonic() at the start and end, then run it myself to verify that, which definitely means it's too confusingly implemented ;-)

tserong:
I've reworked it somewhat, but TBH I'm not sure if it's less confusing or not :-/

The diff is a bit ugly - it's easier to view side-by-side/split than unified.

tserong:
(lint failure is file perms again and should be ignored for this PR)

srv/salt/_modules/osd.py
```diff
@@ -978,17 +978,39 @@ def takeover():
...
+                "OSD {} not active after {} seconds (timeout of {} exceeded)".format(
+                    osd_id, duration, timeout
+                )
+            )
+            return False
+
+        if delay < 32:
+            delay *= 2
+        else:
+            delay = 60
```
trociny, line 1012: nit: you could use something like `delay = min(delay * 2, 60)`, but please ignore if you like your version more.

trociny (approved):
Sorry for delay -- the github notifications for this project seem to be lost in my gmail filters.

It looks good to me, and I like the idea to sleep max 60 sec. Thanks!

tserong:
Thanks @trociny! I've used `delay = min(delay * 2, 60)` because I think that's neater, and I've also squashed the commits and fixed up the commit message.

tserong:
(That last force push was just a rebase)

PR #1883, opened 2021-11-18
Use python3.6 for tox and travis

tserong:
Commit f8a6269 switched to using generic python3, allegedly because Tumbleweed had python 3.7, not 3.6.  I assume that was true at the time, because I wrote it, but given that python 3.6 *is* currently available on Tumbleweed, perhaps I was mistaken.  Anyway, it makes the most sense to force testing with python3.6, because that's what we deploy against with SES6 on SLE 15 SP1.

This commit also:

- Forces the use of isort < 5 when running lint, because version 5 produces a pile of "module 'isort' has no attribute 'SortImports'" errors, and no actual useful lint output.
- Gets rid of the file permission checks when running in Travis CI, because Travis sets umask 002, which messes up the file permissions during `git clone`, and causes all the tests to fail.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -41,7 +41,7 @@ unsafe-load-any-extension=no
 extension-pkg-whitelist=
 
 # Fileperms Lint Plugin Settings
-fileperms-default=0644
+fileperms-default=0o644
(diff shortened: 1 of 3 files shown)


tserong:
Note that it doesn't matter here that the Travis CI builds failed in lint - they're failing _less_ now after this commit than they were before, so things are improved.

PR #1887, opened 2022-02-04
Zap disks in rebuild.node when there are no partitions

0xavi0:
Doing this to zap the partition table, as there are issues deploying
OSDs with GPT partition tables after rebuilding the nodes.

Signed-off-by: Xavi Garcia <xavi.garcia@suse.com>

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1194807

There is a problem when zapping partitions that are part of a disk with GPT partition table.
Deepsea is zapping the partition but not the whole disk, so the partition table remains and doesn't allow the later deployment of the OSD in SES6.

This is mainly happening for clusters upgraded from SES5, with non-lvm disks.

### DIFF
--- a/srv/modules/runners/osd.py
+++ b/srv/modules/runners/osd.py
@@ -344,6 +344,7 @@ class OSDUtil(Util):
         """
         Remote call to minion to zap devices listed by c-v simple scan output
         """
+        parent_devices = self._get_osd_parent_devices()
         simple_scan_json = 'cat /etc/ceph/osd/{}-*.json'.format(self.osd_id)
         cmd = simple_scan_json + (' | jq \'.block.path,.data.path,'
                                   '.["block.db"].path,.["block.wal"].path\''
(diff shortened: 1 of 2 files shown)


tserong (approved):
LGTM!

I've also tested it a couple times on my SES5->6 upgraded VMs, using both `salt-run rebuild.node` to do all OSDs, and `salt-run osd.replace $ID`, to destroy an individual OSD (but keep it in the CRUSH Map for future reuse).

tserong:
Minor point of style - can you please put "Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1194807" in the commit message, just above the Signed-off-by: line? See e.g. https://github.com/SUSE/DeepSea/commit/1168b856 (`git commit --amend ; git push --force` should take care of it)

0xavi0:
Commit message updated, thanks Tim

PR #1888, opened 2022-02-10
Extra information when PGs take too long to be in active+clean state

0xavi0:
When running rebuild.node it waits for PGs to be in the active+clean
state.
If PGs are, for example scrubbing, the user only gets the message: "Waiting for PGs to recover..."

This branch adds extra information, showing the actual state of PGs and,
in the case that they're scrubbing, it sugests to disable scrubbing and
trying.

The return value of quiescent method has been changed to return (in addition to the return result) a message that the rebuild runner can display to the user as a warning.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1194807
Signed-off-by: Xavi Garcia <xavi.garcia@suse.com>

### DIFF
--- a/srv/salt/_modules/osd.py
+++ b/srv/salt/_modules/osd.py
@@ -485,10 +485,16 @@ class CephPGs(object):
             current = self.pg_states()
             if not current:
                 log.warning("PGs are not present")
-                return False
+                return {"result": False, "message": "PGs are not present"}
             if len(current) == 1 and current[0]['name'] == 'active+clean':
                 log.warning("PGs are active+clean")
-                return True
+                return {"result": True, "message": "PGs are active+clean"}
+            if any('scrub' in item['name'] for item in current):
+                message = ("PGs are scrubbing, "
+                           "disable scrubbing and retry.")
+                log.warning(message)
+            else:
+                message = "PGs states: " + str(current)
             log.warning("Waiting on active+clean {}".format(pprint.pformat(current)))
             if self._pg_value(last) != self._pg_value(current):
                 # Making progress - reset counter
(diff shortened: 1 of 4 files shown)


srv/salt/_modules/osd.py
```diff
@@ -481,14 +481,23 @@ def quiescent(self):
...
-                log.warning("PGs are active+clean")
-                return True
+                return {"result": False, "message": "PGs are not present"}
+            if len(current) == 1:
+                if current[0]['name'] == 'active+clean':
+                    log.warning("PGs are active+clean")
+                    return {"result": True, "message": "PGs are active+clean"}
+                else:
+                    if 'scrub' in current[0]['name']:
+                        message = ("PGs are scrubbing, "
```
tserong, line 496: I think this check needs to be moved out of the `if len(current) == 1` block. Instead, if we want to notify if PGs are scrubbing, we probably have to check every element of the `current` array to see if it contains the string "scrub" (some pgs might be active+clean, some might be scrubbing, some might be undersized, some might be in some other state, and for each one of those states, there'll be another element in the `current` array)

srv/salt/_modules/osd.py
```diff
@@ -481,14 +481,23 @@ def quiescent(self):
...
+                                   "disable scrubbing and retry.")
+                    else:
+                        message = "PGs state: " + current[0]['name']
+                    log.warning(message)
             log.warning("Waiting on active+clean {}".format(pprint.pformat(current)))
```
tserong, line 501: This should possibly set `message` to the above string that's logged as a warning (otherwise this function may return "Unknown error")

tserong (changes requested):

tserong:
Alright, I finally managed to get a test of this with an actual scrub running, and I reckon we're good :-) Can you please squash this back to a single commit?

PR #1889, opened 2022-02-21
Update ceph.upgrade.ses7.adopt for SES 7.1

tserong:
This updates the OS and Ceph version checks used when upgrading from SES6. We don't want to merge this immediately - we need to wait and synchronize this with main the SES 7.1 update.


0xavi0 (approved):
Looks good to me

tserong:
That force push was just me tweaking the commit message to add a bug number

